PR #4, opened 2020-03-04
🚧 Introduce base fragment helping with basic state listening

semanticer:
I've put some common functionality to a new `BaseFragment`. The main goal was to force people to handle state inside of the correct scope. There might be more functionality coming into `BaseFragment` as well in the future. One thing we could do would be to hide the storeViewModel entirely and expose only `dispatch` method and this `handleState` override to minimize opportunities to screw something up but I don't know if we want to make things so rigid.
Let me know if this makes any sense.
I'm also not sure about: 
- `BaseFragment` placement
- `BaseFragment` name
- `handleState` name


heytherewill:
Me and @semanticer agreed that this is not needed atm

PR #10, opened 2020-03-05
Time entry database

semanticer:
Added databases, daos, repositories, entities, and other goodies


database-android/build.gradle
```diff
@@ -0,0 +1,46 @@
+apply plugin: 'com.android.library'
```
heytherewill: Why is it named `database-android` and not just database?

models/src/main/java/com/toggl/models/domain/TimeEntry.kt
```diff
@@ -1,9 +1,12 @@
 package com.toggl.models.domain
 
+import androidx.room.Entity
+import androidx.room.PrimaryKey
 import java.util.*
 
+@Entity
 data class TimeEntry(
-    val id: Long,
+    @PrimaryKey(autoGenerate = true) val id: Long = 0,
```
heytherewill: Do we want to mix the models exposed to the state with models that are persisted to the database?

repository/src/main/java/com/toggl/repository/timeentry/TimeEntryRepositoryImpl.kt
```diff
@@ -0,0 +1,42 @@
+package com.toggl.repository.timeentry
+
+import com.toggl.database.dao.TimeEntryDao
+import com.toggl.models.domain.TimeEntry
+import java.util.*
+import javax.inject.Inject
+
+class TimeEntryRepositoryImpl @Inject constructor(private val timeEntryDao: TimeEntryDao): TimeEntryRepository {
```
heytherewill: `RoomTimeEntryRepository` is a better name

database-android/build.gradle
heytherewill: I see that there are two modules, one of which only has two interfaces. I'd just ditch the non-android module (I have no idea what purpose it serves) and just make the `database` module have all the relevant database code

models/src/main/java/com/toggl/models/domain/TimeEntry.kt
heytherewill: As in: Would it not be better to map from the publicly exposed `TimeEntry` to the database relevant `TimeEntry` in the `repository` layer?

app/src/main/java/com/toggl/architecture/AppAction.kt
```diff
@@ -1,9 +1,13 @@
 package com.toggl.architecture
 
+import com.toggl.models.domain.TimeEntry
 import com.toggl.onboarding.domain.actions.OnboardingAction
 import com.toggl.timer.common.domain.TimerAction
 
 sealed class AppAction {
+    object InitTimeEntries : AppAction()
```
heytherewill: I'd just make a single `Load` action that loads all entities

app/src/main/java/com/toggl/architecture/AppAction.kt
```diff
@@ -1,9 +1,13 @@
...
+    data class TimeEntriesLoaded(val timeEntries: List<TimeEntry>) : AppAction()
```
heytherewill: This would be `EntitiesLoaded`

database-android/build.gradle
semanticer: I agree it looks a bit weird right now when both modules are almost empty but...

I was inspired here again by Tivi app. Take o look at his 
https://github.com/chrisbanes/tivi/tree/master/data-android/src/main/java/app/tivi/data
and android less
https://github.com/chrisbanes/tivi/tree/master/data/src/main/java/app/tivi/data

As you can see, the idea is that you can actually keep most of your database code android less which has two benefits. 
- Faster and easier testing
- Faster builds

For comparison here is the Gradle build graph of pure kotlin/java module:
1_0ynyCK44jw2f-1EbCJTRGA
 
And here is Android module:

1_wuGWpml9oNmzvq4is3cYeA

Let me know if we care about this. If we do, we should probably think about some other opportunities for android-less modules.

models/src/main/java/com/toggl/models/domain/TimeEntry.kt
semanticer: Good question. And there is no good answer. I agree that what you are suggesting is definitely a "cleaner" solution that you would find in the true clean architecture where you want to separate the layers as much as possible. However, after a few apps build like that we finally decided to go with a single model/entity approach and never looked back. Room doesn't have any super special requirements for your entities so you can keep all your `**data** class` and `vals` and make everything immutable etc. You just need to apply a couple of annotations on top of it to make Room happy. The benefits are clear, no boilerplate classes and no mappings (and you are usually mapping basically the same class).

But ultimately this is your decision, I understand the motivation for both, especially if you want the domain classes somehow very different.

repository/src/main/java/com/toggl/repository/timeentry/TimeEntryRepositoryImpl.kt
semanticer: Yes `Impl` is terrible :-D but `Room` doesn't seem exactly right either if we use this repository to combine and query different sources? If for example, this repository can trigger BE sync or do the some in-memory cashing etc? Usually, the idea is that repository is somehow handling various data sources and provides a unified entry point for its consumers right? But if our Repositories will be DB only (maybe with some cashing) and we will handle the networking/sync stuff somewhere else, then it makes perfect sense to use `Room` I guess. Just please confirm that this is the case and I will 🔥that `Impl` 💩immediately.

PR #11, opened 2020-03-06
Slightly improved loadable class

semanticer:
I took some inspiration from MvRx's [Async class](https://github.com/airbnb/MvRx/blob/master/mvrx/src/main/kotlin/com/airbnb/mvrx/Async.kt)

- Renaming to `Uninitialized` makes sense because `Nothing` is an important Kotlin type that we actually need to use here.
- `Uninitialized` type should have been object rather than the class from the beginning
-  covariant `out` gets us more flexibility I guess
- `Invoke` is a small and handy operator method.


semanticer:
@heytherewill They also have some custom equals and hashCode implementations if you are interested and it makes sense to you.

Issue #5, opened 2020-03-04
Set up automatic linting


heytherewill:
## Description
We need to setup ktlint (https://github.com/pinterest/ktlint). Ideally, it should run whenever one saves/builds the project locally!

## Definition of done
A pull request that follows these guidelines can close this issue:
- [ ] One commit adds the gradle changes needed to run the linter on save/build
- [ ] One commit commits the changes to the existing files

heytherewill:
Closed by #17

PR #17, opened 2020-03-09
Automatic klint linting

semanticer:
I used [spotless](https://github.com/diffplug/spotless/tree/master/plugin-gradle#applying-ktlint-to-kotlin-files) plugin as it appears to be the most popular solution for Klint integration. 
It provides integration with other useful plugins that we might consider in the future.

Simply run `spotlessKotlinCheck` Gradle task to check your code
Run `spotlessKotlinApply` to fix your code
Also, IntelliJ idea code style has been changed to reflect this so when you hit option+cmd+l you should format the code in a way to satisfy Klint. (also control+option+o for organizing imports) 

Closes: #5


semanticer:
I still need to do more research on how much we can "automate" this.

build.gradle
```diff
@@ -10,28 +11,53 @@ buildscript {
...
+    }
+
+    tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
+        kotlinOptions {
+            // Treat all Kotlin warnings as errors
+            allWarningsAsErrors = true
+
+            // Enable experimental coroutines APIs, including Flow
+            // TODO disable coroutines checks
+//            freeCompilerArgs += "-Xuse-experimental=kotlinx.coroutines.ExperimentalCoroutinesApi"
```
heytherewill: Are these needed?

PR #18, opened 2020-03-09
🚧 Get rid of MainFragment in order to make our UI hierarchy flatter

semanticer:
MainFragment has been removed and MainActivity is the new root of our navigation. This simplifies our UI hierarchy by removing unnecessary fragment nesting.


app/src/main/res/navigation/main_nav_graph.xml
```diff
@@ -12,7 +12,7 @@
 
     <fragment
         android:id="@+id/main"
-        android:name="com.toggl.ui.MainFragment"
+        android:name="com.toggl.ui.MainActivity"
```
heytherewill: Is this legal?

semanticer: no

PR #25, opened 2020-03-09
Add TimeService

### LINKED ISSUE #19: Add a TimeService
## Description
We need to add some interface capable of providing time when we need it. Unlike the current apps, which need something that ticks every second, this time service will have a single responsibility for starters: It will have a `now` method that returns the current time to allow for mocking.

## Definition of done

-  [x] JSR-310 Android Backport (https://github.com/JakeWharton/ThreeTenABP) is added to the project so we can use the new date and time apis
- [ ] The existing date and time types are changed accordingly
- [ ] A `TimeService` interface is added
- [ ] An implementation of the above is added
- [ ] The service is injected in the adequate places
- [ ] The service is used in the adequate places

heytherewill:
### Summary
This adds the JSR-310 backport and uses it in the TimeService, a much needed abstraction for our codebase

We agreed internally on not using joda-time. I inverted the order of dependencies so that others depend on environment because things are simpler this way

### Relationships
Closes #19 

---
### Pull Request checklist

### Pointers
Ensure that this works on older android versions (e.g.: The backport works)

### Review template
- [x] The code considers older platforms well and will work in older devices just like it will in new ones

- [ ] The code considers i18n

## Merge permissions
Only myself


models/build.gradle
```diff
@@ -3,4 +3,5 @@ apply plugin: 'kotlin'
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
     api "androidx.room:room-common:2.2.4"
+    api 'com.jakewharton.threetenabp:threetenabp:1.2.2'
```
daividssilverio: I suspect this dependency should be placed in the `common` module, not in the `models` and it has to be imported in every other module.

models/src/main/java/com/toggl/models/domain/TimeEntry.kt
```diff
@@ -2,14 +2,15 @@ package com.toggl.models.domain
 
 import androidx.room.Entity
 import androidx.room.PrimaryKey
-import java.util.Date
+import java.time.Duration
```
daividssilverio:
```suggestion
import org.threeten.bp.Duration
```

models/src/main/java/com/toggl/models/domain/TimeEntry.kt
```diff
@@ -2,14 +2,15 @@ package com.toggl.models.domain
...
+import java.time.OffsetDateTime
```
daividssilverio:
```suggestion
import org.threeten.bp.OffsetDateTime
```

models/build.gradle
daividssilverio: Well, at least it should be in the `environment` module.

PR #24, opened 2020-03-09
Configure bitrise

heytherewill:
### Summary
This adds our `bitrise.yml` to ensure that any change to bitrise needs to be audited by other developers and will be versioned

### Relationships
Closes #6

---
### Pull Request checklist
- [X] The code does not hinder the app's performance
- [X] The code does not introduce regressions
- [X] The code in this PR is tested thoroughly enough that changing the behavior to an incorrect one breaks tests

### Pointers
Ensure that I'm not forgetting anything and make sure to check the app's dashboard on Bitrise.

## Merge permissions
Only myself

PR #26, opened 2020-03-10
Add crashlytics

### LINKED ISSUE #7: Setup crashlytics
## Description
We agreed on not using app center for crash reporting since we can only use one tool and the reason we didn't go for crashlytics in the old app was Xamarin. Being so, we should setup crashlytics in our app to collect crash data.

## Definition of done
- [ ] The crashlytics packages are added to the project
- [ ] Crashlytics is setup and hooked to our apps on startup
- [ ] ❓ Data is sent to crashlytics? pinging @amulware here for clarification, since we might want to only configure that later, when we are using `com.toggl.giskard` rather than just `com.toggl`

daividssilverio:
## Info
This adds crashlytics

### Summary
This adds the `google-services.json` file and base firebase configuration.
Mainly so that we can add crashlytics.

The `google-services.json` is using the `$BITRISE_KEY` format. The idea is that those keys will be replaced by bitrise.

### Technical considerations
It's textbook firebase + crashlytics setup.

This is the way, right?

### Relationships
Closes #7 

---
### Pull Request checklist

👇  not really yet, gotta make sure which `gservices` configuration we'll be using.

### Pointers
- You can replace the `google-services.json` file manually with the secrets from `Firebase` for the debug projects. Crashlytics is enabled automatically.

### Review template

## Merge permissions
🦑 it after @heytherewill takes a look and @amulware confirms which firebase id we'll be using.


app/google-services.json
```diff
@@ -0,0 +1,94 @@
...
+  "project_info": {
+    "project_number": "$TOGGL_GCM_SENDER_ID",
+    "firebase_url": "$TOGGL_DATABASE_URL",
+    "project_id": "$TOGGL_PROJECT_ID",
+    "storage_bucket": "$TOGGL_STORAGE_BUCKET"
+  },
+  "client": [
+    {
+      "client_info": {
+        "mobilesdk_app_id": "$TOGGL_DROID_GOOGLE_SERVICES_MOBILE_SDK_APP_ID",
```
heytherewill: How will the replacement of these secrets work? Will we have to run some script? If so, this needs to wait for #24 so that you can edit our `bitrise.yml`

build.gradle
```diff
@@ -6,11 +6,16 @@ buildscript {
...
+        maven {
+            url 'https://maven.fabric.io/public'
+        }
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:3.6.1'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
         classpath "de.mannodermaus.gradle.plugins:android-junit5:1.5.0.0" // provides plugin for junit5 support
+        classpath 'com.google.gms:google-services:4.3.3'
+        classpath 'io.fabric.tools:gradle:1.31.2'
```
heytherewill: For reals that the plugin is still called Fabric? 😂

daividssilverio: for reals

app/google-services.json
daividssilverio: Oh... I was checking the plugins... and it seems there won't be any magic...
I guess I'll just create a gradle script to replace the strings, much like how we did for the old app.

But yeah, it depends on #24.

daividssilverio: Now the replacements should work fine!

daividssilverio:
So @heytherewill  @toggl/androids 

I've made it so that the `google-services.json` isn't versioned. We have a `google-services.json.template` that serves as a base to generate the actual `google-services.json` file.

The gradle task `prepareGooglePlayServices` does that. The `Compiler Check` and the `Unit Tests` are failing because of the lack of the `google-services.json`. That can be fixed by running the `prepareGooglePlayServices` in the `base` step (or before any other step that would compile the app).

It's either that, or we can just have the `google-services.json` file regularly and I'll update the gradle task to replace them in place.

What are your thoughts on that?

build.gradle
daividssilverio: Actually... there are multiple docs... The one from the firebase console tells us to do something different, I'll update to reflect that.

heytherewill:
Please merge develop and add the dependencies to the new kotlin file that centralizes dependencies

daividssilverio:
> Please merge master and add the dependencies to the new kotlin file that centralizes dependencies

Will do @heytherewill .

Could you leave a comment about https://github.com/toggl/android/pull/26#issuecomment-597377726 ?

amulware:
I was pinged above as a dependency, but this was merged without me - what happened?

PR #33, opened 2020-03-10
Build types

### LINKED ISSUE #16: Set up app flavors
## Description
Let's configure the build system so that installing a Debug, AdHoc and release builds yield different app ids so that we can have different installed apps on our phones. *NOTE: For now, let's stick to using `com.toggl` as the base bundle ID*

## Definition of done
- [x] The app name is changed to Aurora (we will revert that once we are ready to ship)
- [x] The app icons is set to be the same as the one from the Xamarin app
- [x] The Debug AdHoc and Release builds of the app have different icons 
- [x] The Debug AdHoc and Release builds can co-exist in our phones
- [x] The debug bundle id should be `{base bundle Id}.debug`
- [x] The AdHoc bundle id should be `{base bundle Id}.adhoc`

semanticer:
### Summary
This PR introduces three build types: release, debug, adhoc

flavors

### Technical considerations
1. I defined these build types in all android modules with `applicationIdSuffix` and `versionNameSuffix` for all of them
2. I added build type specific colors for launcher icon tinting
3. I added buid type specific app names

**Signing is now shared across all build types** - that is something we definitely want to change but maybe not in this PR?

We could do this without using the resource system but that would be ugly imo. Feel free to suggest anything else. 

### Relationships

Closes #16 

---
### Pull Request checklist

### Pointers
Make sure you can compile all three variants and that the icon, name, and id is correct.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


heytherewill:
Getting the review from Piotr since he's on the Xamarin rotation (press F to pay respects)

calendar/build.gradle
```diff
@@ -20,8 +20,9 @@ android {
             minifyEnabled false
             proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
         }
+        adhoc {}
```
heytherewill: Are these mandatory even if empty? 🤔 
In  the end, most modules will end up on being the same

app/build.gradle
```diff
@@ -20,9 +20,21 @@ android {
     }
     buildTypes {
         release {
+            initWith(buildTypes.debug)
+
             minifyEnabled false
             proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
         }
+        adhoc {
+            initWith(buildTypes.debug)
```
heytherewill: AdHoc should be a build without debug symbols; what's with this line?

semanticer: That relates to this line from my PR:  **Signing is now shared across all build types** - that is something we definitely want to change but maybe not in this PR?
I needed to build these all build variants to see if they work properly and to do that without specifying signing config for them I reused debug setup. 
Now when I think about it, I should probably remove it since it's already tested.

calendar/build.gradle
semanticer: These are mandatory yes, otherwise the module doesn't "support" that build type. But when I went back to double-check, I discovered that `debug` and `release` are provided automagically so I removed them.

PR #35, opened 2020-03-10
Handle time entry taps

### LINKED ISSUE #28: Add action to handle time entry taps
## Description
We need to add the `TimeEntryTapped` action. For more info, refer to [docs](https://www.notion.so/toggl/Time-Entries-Log-c5c56f70bbad474487d461018adb66fd). No UI is to be handled in this issue

## Definition of done
- [ ] The action is added

heytherewill:
### Summary
This adds the action that handles individual time entry taps

### Technical considerations
When a time entry is tapped, we change the `timeEntryEdited` property and that will trigger navigation to the edit time entry screen (once that screen exists). The edit time entry screen can then handle things as it prefers

### Relationships
Closes #28 

### Pull Request checklist

### Pointers
Not much to see here, just make sure everything is properly tested 

### Review template

- [ ] The code considers older platforms well and will work in older devices just like it will in new 

## Merge permissions
Anyone, anytime


timer/src/main/java/com/toggl/timer/common/domain/TimerState.kt
```diff
@@ -9,9 +9,10 @@ data class TimerState(
     val localState: LocalState
 ) {
     data class LocalState internal constructor(
-        internal val editedDescription: String
+        internal val editedDescription: String,
+        internal val editedTimeEntry: TimeEntry?
     ) {
-        constructor() : this("")
+        constructor() : this("", null)
```
semanticer: I feel like you already explained to someone somewhere bud why again you prefer this over default parameters?

timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogReducer.kt
```diff
@@ -11,6 +11,14 @@ typealias TimeEntriesLogReducer = Reducer<TimeEntriesLogState, TimeEntriesLogAct
 internal fun createTimeEntriesLogReducer(repository: TimeEntryRepository) =
     TimeEntriesLogReducer { state, action ->
         when (action) {
+            is TimeEntriesLogAction.TimeEntryTapped -> {
+                val editedTimeEntry = state.value.timeEntries
+                    .firstOrNull { it.id == action.id }
+                    ?: return@TimeEntriesLogReducer noEffect()
```
semanticer: Maybe we should actually crash here with Illegal argument exception or something like that or at least log this as a non-fatal issue?

timer/src/test/java/com/toggl/timer/log/domain/TimeEntriesLogReducerTests.kt
```diff
@@ -48,4 +50,55 @@ class WhenReceivingContinueButtonTapped : StringSpec() {
             })
         }
     }
-}
+})
+
+class TimeEntryTappedActionTests : WordSpec({
```
semanticer: ContinueButtonTappedActionTests and TimeEntryTappedActionTests seem to be extremely similar. Maybe some light abstraction would make sense here?

heytherewill: We thought it this way in the other apps and now the tests are all bogus. I say we repeat things 😂

timer/src/main/java/com/toggl/timer/common/domain/TimerState.kt
heytherewill: Properties are meant to be internal for these LocalStates, so we need 2 constructors

timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogReducer.kt
heytherewill: That's not a bad idea

timer/src/main/java/com/toggl/timer/common/domain/TimerState.kt
semanticer: Ok, only other solution I can see here is to make the primary constructor public and keep properties internal, that would make solve the access issue but it would still allow anyone from the outside to pass the values. Got it.

timer/src/test/java/com/toggl/timer/log/domain/TimeEntriesLogReducerTests.kt
semanticer: Sound believable 👍

timer/src/test/java/com/toggl/timer/log/domain/TimeEntriesLogReducerTests.kt
```diff
@@ -6,37 +6,39 @@ import com.toggl.models.domain.TimeEntry
 import com.toggl.repository.timeentry.TimeEntryRepository
 import io.kotlintest.properties.assertAll
 import io.kotlintest.shouldBe
-import io.kotlintest.specs.StringSpec
+import io.kotlintest.specs.WordSpec
 import io.mockk.mockk
 import java.util.Date
 
 fun TimeEntriesLogState.toSettableValue(setFunction: (TimeEntriesLogState) -> Unit) =
```
daividssilverio: For the sake of IDE integrations, please move these classes into a outer class so that the IDE can find them and run all of the tests from the file at once.

Otherwise, move each class to its own file.

timer/src/test/java/com/toggl/timer/log/domain/TimeEntriesLogReducerTests.kt
```diff
@@ -6,37 +6,39 @@ import com.toggl.models.domain.TimeEntry
...
+        1,
+        "test",
+        Date(),
+        null,
+        false,
+        null,
+        null
+    )
+
+    "The ContinueButtonTapped action" should {
```
daividssilverio: where is the test for when the continue button tapped is handling an actual valid time entry?

heytherewill: I will create an issue to reify the tests of the action added during the prototype phase 👍

PR #39, opened 2020-03-11
Handle Time Entry Swipes

### LINKED ISSUE #30: Add action to handle time entry swipes
## Description
We need to add the `TimeEntrySwiped` action. For more info, refer to [docs](https://www.notion.so/toggl/Time-Entries-Log-c5c56f70bbad474487d461018adb66fd). No UI is to be handled in this issue

## Definition of done
- [ ] The action is added

heytherewill:
### Summary
This adds time entry swipe handling to our reducers

### Relationships
Closes #30 

### Pull Request checklist

### Pointers
Easy peasy review. Go commit by commit for an easier view. I had to add the ability to delete entries. They need to be soft deleted because of the way sync works. Proper purging happens in our database differently and will only be needed later in the project.

### Review template
- [x] The code is thoroughly tested in well written and resilient manner

## Merge permissions
Anyone, anytime


models/src/main/java/com/toggl/models/common/SwipeDirection.kt
```diff
@@ -0,0 +1,12 @@
+package com.toggl.models.common
+
+sealed class SwipeDirection {
```
daividssilverio: This could just be an enum. Is there a reason to have it like a sealed class?

timer/src/test/java/com/toggl/timer/log/domain/ContinueButtonTappedActionTests.kt
```diff
@@ -0,0 +1,42 @@
...
+
+import com.toggl.architecture.core.noEffect
+import com.toggl.repository.timeentry.TimeEntryRepository
+import com.toggl.timer.common.createTimeEntry
+import io.kotlintest.properties.assertAll
+import io.kotlintest.shouldBe
+import io.kotlintest.specs.WordSpec
+import io.mockk.mockk
+
+class ContinueButtonTappedActionTests : WordSpec({
```
daividssilverio: Didn't I just see this file in another PR?

timer/src/test/java/com/toggl/timer/log/domain/TimeEntrySwipedActionTests.kt
```diff
@@ -0,0 +1,70 @@
...
+                var state = initialState
+                val settableValue = state.toSettableValue { state = it }
+                val effect =
+                    reducer.reduce(settableValue, TimeEntriesLogAction.TimeEntrySwiped(id, SwipeDirection.Left))
+                effect shouldBe noEffect()
+                state shouldBe initialState
+            })
+        }
+
+        "when swiping right" - {
```
daividssilverio: either `when swiping right should`
or `continues the swiped time entry`

daividssilverio: I guess you fixed it later...

models/src/main/java/com/toggl/models/common/SwipeDirection.kt
heytherewill: I come from a land where enums suck balls and aren't properly pattern matched. They are fine in Kotlin, though, so I'll change this :)

daividssilverio:
> Let's investigate them and add them in batch in another issue?

Do you mind creating the issue for that?

PR #42, opened 2020-03-12
Add basic telemetry service and reducer

### LINKED ISSUE #9: Setup telemetry services
## Description
We need to setup our telemetry services to emit events. This should be a reducer of its own.
We must use both AppCenter and Firebase. The approach is simple: we should have some injectable analytics service with a `track(event: Event)` method. This service will be a composite of two implementations (one for AppCenter and one for Firebase). Every app action must implement an `asEvent` extension method. Then we will have an `AnalyticsReducer` that takes the `analyticsService` as a dependency and handles all app actions to send events over the internet 🛰 

## Definition of done
- [x] The firebase and crashlytics telemetry packages are added to the project
- [ ] All existing actions implement an `asEvent` extension method
- [ ] **All implementations of `asEvent` make sure the events are anonymous and that no user info is being sent to neither telemetry service**
- [ ] The `AnalyticsReducer` is setup and working

daividssilverio:
### Summary
This PR adds the Analytics service and reducer.

### Technical considerations
1. The analytics service is using both Firebase & AppCenter.
2. The analytics reducer consider that the the `AppAction.toEvent` extension will handle mapping app actions into events (we could move it somewhere else for sure).

I think the `AppAction.toEvent` idea can be implement in many different ways. I'm all ears.

And I think we could add an issue just to map those events.

### Relationships
Closes #9 

---
### Pull Request checklist

### Pointers
You can add a log to the analytics reducer in order to see it working.

### Review template

## Merge permissions
Let's discuss it a bit before merging; But changing this later shouldn't be much of a problem.
But yeah, 🦑 it after the reviews.


environment/src/main/java/com/toggl/environment/services/analytics/AnalyticsServiceImpl.kt
```diff
@@ -0,0 +1,26 @@
+package com.toggl.environment.services.analytics
+
+import android.content.Context
+import android.os.Bundle
+import com.google.firebase.analytics.FirebaseAnalytics
+import com.microsoft.appcenter.analytics.Analytics
+import javax.inject.Inject
+
+class AnalyticsServiceImpl @Inject constructor(context: Context) : AnalyticsService {
+    private val firebaseAnalytics: FirebaseAnalytics = FirebaseAnalytics.getInstance(context)
```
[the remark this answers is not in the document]
semanticer: This would help us with extensibility and especially testability since right now this class is not testable at all.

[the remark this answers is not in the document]
heytherewill: I came here to say this :)

environment/src/main/java/com/toggl/environment/services/analytics/AnalyticsServiceImpl.kt
```diff
@@ -0,0 +1,26 @@
+package com.toggl.environment.services.analytics
+
+import android.content.Context
+import android.os.Bundle
+import com.google.firebase.analytics.FirebaseAnalytics
+import com.microsoft.appcenter.analytics.Analytics
+import javax.inject.Inject
+
+class AnalyticsServiceImpl @Inject constructor(context: Context) : AnalyticsService {
```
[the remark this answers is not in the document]
heytherewill: `CompositeAnalyticsService` is the perfect definition of it

app/src/main/java/com/toggl/TogglApplication.kt
```diff
@@ -16,6 +18,10 @@ class TogglApplication : Application(), OnboardingComponentProvider, TimerCompon
     override fun onCreate() {
         super.onCreate()
         AndroidThreeTen.init(this)
+
+        if (!BuildConfig.DEBUG) {
+            AppCenter.start(this, "TheAppSecret", Analytics::class.java)
```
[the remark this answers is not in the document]
daividssilverio: It can be cool, do you think we should start do it now?
What else will we have to initialize 🤔 ?

environment/src/main/java/com/toggl/environment/services/analytics/AnalyticsServiceImpl.kt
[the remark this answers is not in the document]
daividssilverio: Will do.

app/src/main/java/com/toggl/domain/reducers/AnalyticsReducer.kt
```diff
@@ -0,0 +1,16 @@
...
+import com.toggl.environment.services.analytics.AnalyticsService
+import com.toggl.environment.services.analytics.Event
+
+fun createAnalyticsReducer(analyticsService: AnalyticsService) =
+    Reducer<AppState, AppAction> { _, action ->
+        analyticsService.track(action.toEvent())
+        noEffect()
+    }
+
+fun AppAction.toEvent(): Event? = null
```
daividssilverio: No comments on this?

environment/src/main/java/com/toggl/environment/di/EnvironmentModule.kt
```diff
@@ -18,6 +21,19 @@ class EnvironmentModule {
...
-        AnalyticsServiceImpl(context)
+    fun analyticsService(
+        @Named("FirebaseAnalyticsService") firebaseAnalyticsService: FirebaseAnalyticsService,
+        @Named("AppCenterAnalyticsService") appCenterAnalyticsService: AppCenterAnalyticsService
+    ): AnalyticsService =
+        CompositeAnalyticsService(firebaseAnalyticsService, appCenterAnalyticsService)
+
+    @Provides
+    @Singleton
+    @Named("FirebaseAnalyticsService")
```
heytherewill: Why the name? I think the type should've sufficed, no?

environment/src/main/java/com/toggl/environment/services/analytics/AppCenterAnalyticsService.kt
```diff
@@ -0,0 +1,10 @@
+package com.toggl.environment.services.analytics
+
+import com.microsoft.appcenter.analytics.Analytics
+import javax.inject.Inject
+
+class AppCenterAnalyticsService @Inject constructor() : AnalyticsService {
```
heytherewill: What are you injecting here?

app/src/main/java/com/toggl/di/AppModuleBinds.kt
```diff
@@ -2,11 +2,23 @@ package com.toggl.di
...
     @Binds
     abstract fun provideApplication(bind: TogglApplication): Application
+
+    @Binds
+    @IntoSet
+    abstract fun provideAppCenterInitializer(bind: AppCenterInitializer): AppInitializer
+
+    @Binds
+    @IntoSet
+    abstract fun provide310Initializer(bind: ThreeTenInitializer): AppInitializer
```
heytherewill:
```suggestion
    abstract fun provideThreeTenInitializer(bind: ThreeTenInitializer): AppInitializer
```

app/src/main/java/com/toggl/domain/reducers/AnalyticsReducer.kt
heytherewill: The `toEvent` methods should be on each action's file, not elsewhere, IMO

environment/src/main/java/com/toggl/environment/di/EnvironmentModule.kt
daividssilverio: there are two `AnalyticsServices` being provided, they gotta be named (or I could use `Binds`)

heytherewill: If you:

```kotlin
fun firebaseAnalyticsService(context: Context): FirebaseAnalyticsService = FirebaseAnalyticsService(context)
```

That should remove the need for a name, shouldn't it? I don't like raw strings being used like that; proper types are clearer

daividssilverio: Oh no, I didn't notice I've used the concrete classes. Will update it.

environment/src/main/java/com/toggl/environment/services/analytics/AppCenterAnalyticsService.kt
daividssilverio: Oh, not needed (unless we use `Binds`).

app/src/main/java/com/toggl/domain/reducers/AppReducer.kt
```diff
@@ -25,11 +26,13 @@ typealias AppReducer = Reducer<AppState, AppAction>
...
     timeEntryListReducer: TimeEntryListReducer,
     onboardingReducer: OnboardingReducer,
-    timerReducer: TimerReducer
+    timerReducer: TimerReducer,
+    analyticsService: AnalyticsService
 ): AppReducer =
     combine(
         timeEntryListReducer,
         createLoggingReducer(),
+        createAnalyticsReducer(analyticsService),
```
heytherewill: I think a clearer approach would be to make every reducer be provided by dagger and only list them here, what do you think?

environment/src/main/java/com/toggl/environment/services/analytics/AnalyticsService.kt
```diff
@@ -0,0 +1,10 @@
+package com.toggl.environment.services.analytics
+
+interface AnalyticsService {
+    fun track(event: Event?)
+}
+
+interface Event {
+    val name: String
+    fun toMap(): Map<String, String>
```
heytherewill: Why `toMap` and not just a map?

heytherewill: Also, why make this an interface and not a data class?

heytherewill: Also, this should be in its own file

environment/src/main/java/com/toggl/environment/services/analytics/AnalyticsService.kt
```diff
@@ -0,0 +1,10 @@
+package com.toggl.environment.services.analytics
+
+interface AnalyticsService {
+    fun track(event: Event?)
```
heytherewill: Why should it receive a nullable? I think we should filter null events elsewhere and not in the service

app/src/main/java/com/toggl/domain/reducers/AppReducer.kt
daividssilverio: Sounds good to me o/

environment/src/main/java/com/toggl/environment/services/analytics/AnalyticsService.kt
daividssilverio: > Why toMap and not just a map?

Because the events are tracked like that 🤔

daividssilverio: > Also, this should be in its own file

I fell into my own complaints.

environment/src/main/java/com/toggl/environment/services/analytics/AnalyticsService.kt
daividssilverio: I agree, I can do that check in the reducer instead.

environment/src/main/java/com/toggl/environment/services/analytics/AnalyticsService.kt
daividssilverio: > Also, why make this an interface and not a data class?

The initial idea was to have the `AppAction`s implement `Event`, but a data class should work just fine.

app/src/main/java/com/toggl/domain/reducers/AppReducer.kt
daividssilverio: To avoid using the `@Named` injectables, I had to open the `Reducer` class and make the log and analytics reducers inherit from it.

What do you think? Is there a reason for the `Reducer` to be final?

app/src/main/java/com/toggl/domain/reducers/AnalyticsReducer.kt
daividssilverio: That would pollute each action and distribute of the events across a lot of files.
Why do you rather have it that way? What are the pros & cons?

If we leave the events mapping here, it's easier to find them and the reducer and its tests are kept simple.
If we move the `toEvent` to each action, we'll also need to `case-match` each one of them in this reducer.
If we keep them here, we'll do the same in the `toEvent` extension, but nothing would need to change in the reducer itself or its tests.

app/src/main/java/com/toggl/TogglApplication.kt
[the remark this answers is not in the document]
semanticer: Right now only the AndroidThreeTen above and I think we will need Timber for logging. Anyway, let's not deal with this now in this PR, I will discuss this with @heytherewill separately.

app/src/main/java/com/toggl/domain/reducers/AnalyticsReducer.kt
semanticer: @daividssilverio Not 100% sure If I understand correctly what you proposing but I see an issue in enforcing cases for all Actions. Ideally, we want compile-time errors when you add a new action and forget to add a tracking event, can you enforce that?

app/src/main/java/com/toggl/TogglApplication.kt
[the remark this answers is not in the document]
semanticer: Oh, you already implemented the whole thing... cool 👍

app/src/main/java/com/toggl/domain/reducers/AnalyticsReducer.kt
daividssilverio: > Ideally, we want compile-time errors whenever you add a new action and forget to add a tracking event, can you enforce that?

We want that? We'd have to make the `AppActions` implement an interface or inherit from an abstract class to enforce `AppActions` to always have a tracking event. @heytherewill was against that.

> Not 100% sure If I understand correctly what you are proposing but I see an issue in enforcing cases for all Actions.
The implementation would look like this:

```
return when (this) {
        AppAction.Load -> {
            Event("someEvent", mapOf())
        }
        is AppAction.EntitiesLoaded -> {
            Event("someEvent", mapOf())
        }
        is AppAction.Onboarding -> return when (this.onboarding) {
            OnboardingAction.LoginTapped -> Event("someEvent", mapOf())
            is OnboardingAction.SetUser -> Event("someEvent", mapOf())
            is OnboardingAction.SetUserError -> Event("someEvent", mapOf())
            is OnboardingAction.EmailEntered -> Event("someEvent", mapOf())
            is OnboardingAction.PasswordEntered -> Event("someEvent", mapOf())
        }
        is AppAction.Timer -> return when (this.timer) {
            is TimerAction.StartTimeEntry -> Event("someEvent", mapOf())
            is TimerAction.TimeEntriesLog -> Event("someEvent", mapOf())
        }
    }
```

What is the issue?

If we would make the actions implement an `Event` interface, we wouldn't need that.
if use an extension in each AppAction instead, we would move that `when` inside the reducer basically.

heytherewill: Imagine that the `AppAction` file has the following:

```kotlin
fun AppAction.toEvent() : Event? = when (this) {
    AppAction.Load -> Event("someEvent", mapOf())
    is AppAction.EntitiesLoaded -> Event("someEvent", mapOf())
    is AppAction.Onboarding ->  action.onboarding.toEvent()
    is AppAction.Timer -> action.timer.toEvent()
}
```

Then, each action file will have its own overload of `toEvent`, which would delegate creation further down the chain. This still ensures compiler errors when adding action without having a god file, but, as @daividssilverio says, scatters the events. We could alternatively keep this all centralized in the `app` module under a couple of extension files inside an `analytics` package, but I'm not sure which approach I like best

heytherewill:
Assigning myself to get this merged

PR #43, opened 2020-03-12
Handle group actions

### LINKED ISSUE #29: Add action to handle time entry group taps
## Description
We need to add the `TimeEntryGroupTapped` action. For more info, refer to [docs](https://www.notion.so/toggl/Time-Entries-Log-c5c56f70bbad474487d461018adb66fd). No UI is to be handled in this issue

## Definition of done
- [ ] The action is added

### LINKED ISSUE #31: Add action to handle time entry group swipes
## Description
We need to add the `TimeEntryGroupSwiped` action. For more info, refer to [docs](https://www.notion.so/toggl/Time-Entries-Log-c5c56f70bbad474487d461018adb66fd). No UI is to be handled in this issue

## Definition of done
- [ ] The action is added

heytherewill:
### Summary
This adds and handles the two actions related to groups

### Technical considerations
The first two commits are a bit different, so I'll comment on them. The first one moves away from using a List to using a Map. This is because we will be doing a lot of direct access, so the O(1) performance is needed. The second commit changes the soft delete mechanism to accept a list of entries. This is done because, when you only need to delete one it's simpler to send a list of one item than to have two methods. In the future, deleting can be done smarter (e.g.: If you delete an unsynced time entry, you can remove it from the database permanently) but for now that's good enough!

The action handling itself is fairly trivial

Feel free to comment on my solution to improve it.

### Relationships
Closes #29 
Closes #31 

---
### Pull Request checklist

### Pointers
Check that the tests are correct and look at the PR commit by commit

### Review templateuch UI -->

## Merge permissions
Only myself or @semanticer


app/src/main/java/com/toggl/domain/AppState.kt
```diff
@@ -9,7 +9,7 @@ import com.toggl.timer.common.domain.TimerState
 
 data class AppState(
     val user: Loadable<User> = Loadable.Uninitialized,
-    val timeEntries: List<TimeEntry> = listOf(),
+    val timeEntries: Map<Long, TimeEntry> = mapOf(),
```
daividssilverio: With the maps in place, don't you think that the mapping done in the `TimeEntriesLogFragment` should be run in a background thread?

timer/src/test/java/com/toggl/timer/log/domain/TimeEntryGroupTappedActionTests.kt
```diff
@@ -0,0 +1,49 @@
...
+    val testTe = createTimeEntry(1, "test")
+
+    "The TimeEntryTapped action" - {
+        "should thrown when there are no time entries" - {
+            "with the matching id" {
+                val initialState = createEmptyState(listOf(testTe))
+                var state = initialState
+                val settableValue = state.toSettableValue { state = it }
+                shouldThrow<IllegalStateException> {
+                    reducer.reduce(settableValue, TimeEntriesLogAction.TimeEntryTapped(2))
```
daividssilverio: You're still testing `TimeEntryTapped` instead of `TimeEntryGroupTapped` here.

timer/src/test/java/com/toggl/timer/log/domain/TimeEntryGroupTappedActionTests.kt
```diff
@@ -0,0 +1,49 @@
...
+import io.kotlintest.specs.FreeSpec
+import io.mockk.mockk
+
+class TimeEntryGroupTappedActionTests : FreeSpec({
+
+    val repository = mockk<TimeEntryRepository>()
+    val reducer = createTimeEntriesLogReducer(repository)
+    val testTe = createTimeEntry(1, "test")
+
+    "The TimeEntryTapped action" - {
```
daividssilverio: ©️ 🍝  Did you forget to replace the tests with their `Grouped` variations?

Issue #45, opened 2020-03-12
Add cell UI for the header


heytherewill:
## Description
We need to add the cell for the time entries header and handle it in the adapter

## Definition of done
- [ ] The cell UI is added
- [ ] The cell UI is considered by the adapter when choosing what to render

heytherewill:
Closed by #47

PR #47, opened 2020-03-13
Timer headers

### LINKED ISSUE #44: Add time entry headers when displaying time entry ViewModels
## Description
Right now we are only mapping time entries into `FlatTimeEntryViewModel`. We need to add the day headers, along with the totals.

## Dependencies
Depends on #45

## Definition of done
- [ ] A day header is added before all time entries of a day
- [ ] Custom dates (today and yesterday) are shown as strings. The rest of dates should be formatted
- [ ] The header should have the duration of all time entries for a particular day
- [ ] Tests need to ensure the grouping is correct

heytherewill:
### Summary
This adds the header UI and starts grouping the headers as well

### Technical considerations
Not much to consider here other than where the mapping takes place. 

### Relationships
Closes #44 and #45

### Pull Request checklist

### Pointers
Use the app and make sure things are working fine

### Review template

- [ ] The code considers i18n


common-android/src/main/res/values-ja/strings.xml
```diff
@@ -3,9 +3,11 @@
     <string name="login">ログイン</string>
     <string name="add_description">説明を追加</string>
     <string name="calendar">カレンダー</string>
-    <string name="reports">レポート</string>
+    <string name="reports">レポート</string>C:\Users\William\Documents\Projects\mobileapp\Toggl.Shared\Resources.ja.resx
```
daividssilverio: 👀

heytherewill: Errou feio, errou rude

timer/src/main/java/com/toggl/timer/extensions/DurationExtensions.kt
```diff
@@ -0,0 +1,13 @@
+package com.toggl.timer.extensions
+
+import org.threeten.bp.Duration
+import java.util.concurrent.TimeUnit
+
+fun Duration.formatForDisplaying(): String {
```
daividssilverio: Oh, please add a test for this extension.

timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogSelector.kt
```diff
@@ -0,0 +1,63 @@
...
+import com.toggl.environment.services.TimeService
+import com.toggl.models.domain.Project
+import com.toggl.models.domain.TimeEntry
+import com.toggl.timer.R
+import org.threeten.bp.Duration
+import org.threeten.bp.LocalDate
+import org.threeten.bp.format.DateTimeFormatter
+
+fun timeEntriesLogSelector(
+    context: Context,
```
daividssilverio: isn't it just safer to request the strings `R.string.today, R.string.yesterday` or the `formatHeaderDate` enclosing the context or just the strings instead of using the `requireContext` where this function is called from?

something like: 
```
formatHeaderDate(today: String, yesterday: String, date: LocalDate, today: LocalDate, yesterday: LocalDate)`
```

heytherewill: That works too. The whole activity gets destroyed on configuration changes anyways, so it should work fine

timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogAdapter.kt
```diff
@@ -1,41 +1,84 @@
...
             override fun areItemsTheSame(
                 oldItem: TimeEntryViewModel,
                 newItem: TimeEntryViewModel
             ): Boolean =
-                oldItem is FlatTimeEntryItem &&
-                    newItem is FlatTimeEntryItem &&
-                    oldItem.id == newItem.id
+                when (oldItem) {
+                    is FlatTimeEntryViewModel ->
+                        if (newItem is FlatTimeEntryViewModel) oldItem.id == newItem.id else false
```
daividssilverio: All of these `if - else` clauses can be reduced into statements like:
```
                    is FlatTimeEntryViewModel ->
                        newItem is FlatTimeEntryViewModel && oldItem.id == newItem.id
                    is DayHeaderViewModel ->
                        newItem is DayHeaderViewModel && oldItem.dayTitle == newItem.dayTitle
                    is TimeEntryGroupViewModel ->
                        newItem is TimeEntryGroupViewModel && oldItem.timeEntryIds == newItem.timeEntryIds
```

timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogAdapter.kt
```diff
@@ -1,41 +1,84 @@
...
             override fun areContentsTheSame(
                 oldItem: TimeEntryViewModel,
                 newItem: TimeEntryViewModel
             ): Boolean =
-                oldItem is FlatTimeEntryItem &&
-                    newItem is FlatTimeEntryItem &&
-                    oldItem.description == newItem.description
+                when (oldItem) {
+                    is FlatTimeEntryViewModel ->
+                        if (newItem is FlatTimeEntryViewModel) oldItem == newItem else false
```
daividssilverio: Same goes for these `if  - else`s

timer/src/test/java/com/toggl/timer/extensions/DurationFormatExtensionTests.kt
```diff
@@ -0,0 +1,19 @@
...
+import com.toggl.timer.generators.threeTenDuration
+import io.kotlintest.matchers.numerics.shouldBeGreaterThanOrEqual
+import io.kotlintest.properties.Gen
+import io.kotlintest.properties.assertAll
+import io.kotlintest.specs.FreeSpec
+import org.threeten.bp.Duration
+
+class DurationFormatExtensionTests : FreeSpec({
+    "The formatForDisplaying method" - {
+        "Uses the hh:mm:ss format" - {
```
daividssilverio: Please add at least three tests, one for a duration that would result in seconds only, one that would include minutes, and one that include hours.

timer/src/test/java/com/toggl/timer/log/domain/TimeEntriesLogSelectorTests.kt
```diff
@@ -0,0 +1,102 @@
...
+                                    DateTimeFormatter.ofPattern(timeEntriesLogHeaderTimeFormat)
+                                )
+                            }
+                        }.distinct().toList()
+
+                headers shouldContainExactlyInAnyOrder flatTimeEntryHeaders
+            }
+        }
+
+        "correctly formats header dates for today and yesterday" - {
```
daividssilverio: Please add at least one test for the headers of the other types of days (not today, not yesterday).

timer/src/test/java/com/toggl/timer/extensions/DurationFormatExtensionTests.kt
heytherewill: I'm using property based testing here, so the test always generates a bunch of random durations for formatting. Since the format is hh:mm:ss, I'm ensuring that we're not getting weird stuff like `1:1:1`, but rather `01:01:01` instead. This does include cases with very short and very long durations. Do you think the assertions can be improved somehow?

daividssilverio: Here's my point: 
You not testing that hours are parsed as hours, minutes as minutes and seconds as seconds. I think it's important for the tests to fail when the format changes. What if someone decides to use "HH:mm:ss" or "kk:mm:ss" for some reason.

If you disagree, ignore this comment and let's move on.

heytherewill: There's no way to test this without copying the code from the extension, which is not a good idea.

timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogSelector.kt
```diff
@@ -0,0 +1,60 @@
...
+
+    val today = timeService.now().toLocalDate()
+    val yesterday = today.minusDays(1)
+
+    return timeEntries.values
+        .filter { it.duration != null }
+        .sortedByDescending { it.startTime }
+        .groupBy { timeEntry -> timeEntry.startTime.toLocalDate() }
+        .flatMap { (groupDate, timeEntries) ->
+            sequence {
```
semanticer: I have nothing against imperative code in general but when we already have a nice chain like that I would prefer to avoid that sequence generation. What about: 
```kotlin
...
.groupBy { timeEntry -> timeEntry.startTime.toLocalDate() }`
.flatMap { (groupDate, timeEntries) -> listOf(toHeader(groupDate, timeEntries)) + toFlatItems(timeEntries) }
```
And define `toHeader` and `toFlatItems` separately below. 
What do you think?

timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogFragment.kt
```diff
@@ -41,11 +47,18 @@ class TimeEntriesLogFragment : Fragment(R.layout.time_entries_log_fragment) {
 
         recycler_view.adapter = adapter
 
+        val context = requireContext()
+        val todayString = context.getString(R.string.today)
+        val yesterdayString = context.getString(R.string.yesterday)
+
+        val curriedTimeEntriesSelector: suspend (TimeEntriesLogState) -> List<TimeEntryViewModel> = {
```
semanticer: What is the reasoning behind doing this here on the UI level? Not sure what is correct so I would like to clarify that.

timer/src/main/java/com/toggl/timer/log/ui/TimeEntryHeaderViewHolder.kt
```diff
@@ -0,0 +1,17 @@
+package com.toggl.timer.log.ui
+
+import android.view.View
+import android.widget.TextView
+import com.toggl.timer.R
+import com.toggl.timer.extensions.formatForDisplaying
+import com.toggl.timer.log.domain.DayHeaderViewModel
+
+class TimeEntryHeaderViewHolder(itemView: View) : TimeEntryLogViewHolder(itemView) {
+    private val title = itemView.findViewById<TextView>(R.id.title)
```
semanticer: There is a way to use android extensions even for View holders. See https://kotlinlang.org/docs/tutorials/android-plugin.html under `Experimental features`
It is experimental for a year now and it works well.
But anyway - wait when I rewrite everything with epoxy :-D

timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogSelector.kt
heytherewill: I find sequence yielding much clearer and it also involves less list creations

timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogFragment.kt
heytherewill: How do you mean here and not on UI level? The fragment is the UI 🤔 Where else would you put it?

timer/src/main/java/com/toggl/timer/log/ui/TimeEntryHeaderViewHolder.kt
heytherewill: A man can dream 😂 
Once we have all proper grouping you can experiment with epoxy :)

timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogFragment.kt
semanticer: Sorry, my question was if this even should be a part of UI layer.

timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogSelector.kt
semanticer: I don't but I can live with that :-D

timer/src/test/java/com/toggl/timer/extensions/DurationFormatExtensionTests.kt
daividssilverio:
```
"Displays seconds" - {
            val duration = Duration.ofSeconds(10)

            val formatted = duration.formatForDisplaying()

            formatted shouldBe "00:00:10"
        }

        "Displays minutes" - {
            val duration = Duration.ofMinutes(1)

            val formatted = duration.formatForDisplaying()

            formatted shouldBe "00:01:00"
        }

        "Displays hours" - {
            val duration = Duration.ofHours(1)

            val formatted = duration.formatForDisplaying()

            formatted shouldBe "01:00:00"
        }

        "Displays hours, minutes and seconds" - {
            val duration = Duration.ofHours(1) + Duration.ofMinutes(1) + Duration.ofSeconds(1)

            val formatted = duration.formatForDisplaying()

            formatted shouldBe "01:01:01"
        }
```

I was talking about something like that ☝️

PR #56, opened 2020-03-13
Enforce our effect contract

### LINKED ISSUE #54: Improve the contract of our effects
## Description
Our effects can only ever emit a single action, but right now their interfaces allow for more. We need to fix this. While at it, we should make the effects a proper interface to make them easier to mock/inject

## Definition of done
- [ ] The effect type is remodeled to return 0 or 1 action instead of 0 or many as it is right now

heytherewill:
### Summary
As discussed with @semanticer, even though effects should only return a single action, atm they allow for many because of Flow, We should change that

I used concrete interfaces/classes because that's what we are moving towards (ideally we will even change the reducer type to be the same to make boilerplate easier on us), but I'm all ears.

### Relationships
Closes #54 

---
### Pull Request checklist

### Pointers
Please make sure nothing broke during this minor refactor.

### Review template

- [X] The code is thoroughly tested in well written and resilient manner
- [X] The code is performant and leads to an optimal user experience
- [X] The code is testable and does not cut corners for the sake of delivering
- [X] The code considers older platforms well and will work in older devices just like it will in new ones

## Merge permissions
Only myself


architecture/src/main/java/com/toggl/architecture/extensions/EffectExtensions.kt
```diff
@@ -0,0 +1,31 @@
+package com.toggl.architecture.extensions
+
+import com.toggl.architecture.core.Effect
+
+fun <T, R> Effect<T>.map(mapFn: (T?) -> R?): Effect<R> =
```
semanticer: I was just thinking if we want to introduce these custom extensions for our Effect interface or if we should just embrace the raw power of suspending functions rather than introducing new features to our "wrapper" around them. But after some experimentation, I probably agree that your approach is cleaner.

PR #57, opened 2020-03-13
Add swipes to the UI

### LINKED ISSUE #46: Add swipes to the UI
## Description
We need to enable swiping right/left and make the cells emit the actions accordingly. 

## Definition of done
- [ ] Swipes are added to the cell that represents a time entry
- [ ] Swipe actions are emitted when you swipe left or right

daividssilverio:
### Summary
This adds the basic swipe handler for the main log adapter

### Technical considerations
The implementation is pretty straightforward, using a simple drawable for the background when a time entry is swiped.
We can add some text or icon later as necessary.
The swiping directions are defaulting to `swipe right to continue` and `swipe left to delete`.

Because drawing using `onChildDraw` is easy enough and doesn't add anything to actual layout of each view holder we'll eventually have.

### Relationships
Closes #46

---
### Pull Request checklist

### Pointers
Just swipe some time entries to see that it works.

### Review template

## Merge permissions
🦑 it


timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogFragment.kt
```diff
@@ -50,4 +58,19 @@ class TimeEntriesLogFragment : Fragment(R.layout.time_entries_log_fragment) {
...
+        if (adapterPosition >= currentItems.size) {
+            adapter.notifyDataSetChanged()
+            return
+        }
+
+        when (val item = currentItems[adapterPosition]) {
+            else -> {}
+        }
+        if (swipeDirection == SwipeDirection.Right) {
+            adapter.notifyItemChanged(adapterPosition)
```
heytherewill: Why is this needed? To update the color of the background?

timer/src/main/java/com/toggl/timer/log/ui/SwipeActionCallback.kt
```diff
@@ -0,0 +1,75 @@
...
+        dX: Float,
+        dY: Float,
+        actionState: Int,
+        isCurrentlyActive: Boolean
+    ) {
+        super.onChildDraw(c, recyclerView, viewHolder, dX, dY, actionState, isCurrentlyActive)
+        val itemView = viewHolder.itemView
+        when {
+            dX > 0 -> {
+                swipeBackground.color = Color.GREEN
```
heytherewill: We should reuse the colors of the start and stop button here

timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogFragment.kt
daividssilverio: To reset the swiped view. Otherwise it stays all green.

daividssilverio:
> I remember that we initially had a solution like this one in the Xamarin app, but then changed to updating the layout (as you can see here, the layout for the time entry log item contains swipe views, which you should remove if we are using the canvas approach).

I don't remember exactly, but I guess I was just afraid of the canvas back then, tbh.

semanticer:
@daividssilverio You afraid of the canvas? Impossible...

timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogFragment.kt
```diff
@@ -50,4 +58,19 @@ class TimeEntriesLogFragment : Fragment(R.layout.time_entries_log_fragment) {
                 .launchIn(this)
         }
     }
+
+    private fun onLogItemSwiped(adapterPosition: Int, swipeDirection: SwipeDirection) {
+        val currentItems: List<TimeEntryViewModel> = adapter.currentList
+        if (adapterPosition >= currentItems.size) {
+            adapter.notifyDataSetChanged()
```
semanticer: Calling `notifyDataSetChanged` is not verry efficient although this is not called that often.  Would `adapter.notifyItemChanged(x)` or `adapter.notifyItemRangeChanged(x, y)` work?

timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogFragment.kt
```diff
@@ -67,6 +67,8 @@ class TimeEntriesLogFragment : Fragment(R.layout.time_entries_log_fragment) {
         }
 
         when (val item = currentItems[adapterPosition]) {
+            is FlatTimeEntryItem -> store.dispatch(TimeEntriesLogAction.TimeEntrySwiped(item.id, swipeDirection))
```
semanticer: I would maybe just return the event from the when statement and dispatch on the next line

timer/src/main/java/com/toggl/timer/log/ui/SwipeActionCallback.kt
```diff
@@ -0,0 +1,75 @@
...
+                swipeBackground.setBounds(
```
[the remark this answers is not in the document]
heytherewill: Turns out we can't use `updateBounds` because it uses `bounds.side` as default parameters and here we are using `item.side`. When I tried using the extension method, I got a transparent background rather than the colored one :/

timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogFragment.kt
heytherewill: I wonder if this is at all needed. @daividssilverio when will the swiped adapter's position be bigger than the amount of items? When the user swipes and there's an update in the recyclerview? 🤔

heytherewill:
Updated to cover our reviews, @semanticer. Still missing:

    The name of the actions should be rendered as well
    Daivid needs to clarify one of the comments in your review

timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogFragment.kt
daividssilverio: It is needed as far as I tested, maybe just notifying the changes in the range could work better, just gotta figure which part of the range changed.

This is needed basically because when you continue a time entry, that time entry doesn't change, so the swiped view remains swiped (green background).
We could add a listener to the swipe action, wait for a little while and revert the swiping if necessary.

daividssilverio: If there is other way without notifying changes to the adapter, please go for it.

heytherewill: > This is needed basically because when you continue a time entry, that time entry doesn't change, so the swiped view remains swiped 

Isn't that why we have `adapter.notifyItemChanged(adapterPosition)` a little further down the line? Why is this global reset needed, and only needed when the adapter position is greater than the total amount of items? When will this ever happen?

daividssilverio: It's because you can't be sure about the position of the changed position.
When continuing a time entry, it might stop the running time entry.

semanticer: Let's leave it there for now.

timer/src/main/java/com/toggl/timer/log/ui/SwipeActionCallback.kt
```diff
@@ -1,19 +1,46 @@
...
+
+    private val swipeActionPadding = resources.getDimension(R.dimen.swipe_action_text_padding)
+    private val swipeLeftTextWidth = textPaint.measureText(swipeToLeftString)
+    private val swipeRightTextWidth = textPaint.measureText(swipeToRightString)
+
+    @RequiresApi(Build.VERSION_CODES.M)
+    val swipeLeftLabelLayout = buildLabelLayout(swipeToLeftString, swipeLeftTextWidth.toInt())
+
+    @RequiresApi(Build.VERSION_CODES.M)
+    val swipeRightLabelLayout = buildLabelLayout(swipeToRightString, swipeRightTextWidth.toInt())
```
heytherewill: What about lollipop? 🤔

timer/src/main/java/com/toggl/timer/log/ui/SwipeActionCallback.kt
```diff
@@ -1,19 +1,46 @@
...
 import androidx.recyclerview.widget.ItemTouchHelper
 import androidx.recyclerview.widget.RecyclerView
 import com.toggl.models.common.SwipeDirection
+import com.toggl.timer.R
 
 class SwipeActionCallback(
     @ColorInt private val swipeToLeftBackground: Int,
     @ColorInt private val swipeToRightBackground: Int,
+    @ColorInt private val actionTextColor: Int,
+    swipeToLeftString: String,
```
heytherewill: I think it'd be good to have an entity like:

```kotlin
data class SwipeAction(
    val text: String,
    @ColorInt val backgroundColor: Int,
    @ColorInt val textColor: Int
)
```

This way we can just pass two of  those to the `SwipeActionCallback` instead of having so many parameters. This is particularly good once we start letting users have custom swipe actions

timer/src/main/java/com/toggl/timer/log/ui/SwipeActionCallback.kt
semanticer: Oh shit I forgot I took the easy way out :-)

PR #62, opened 2020-03-16
🔧 Run prepareGooglePlayServices as a dependency of tasks that need it

### LINKED ISSUE #61: Run prepareGooglePlayServices automatically
## Description
We should run prepareGooglePlayServices task automatically, so that the project builds fine out of the box, without running any custom steps by hand.

## Definition of done
- [ ] Cloning the repo and hitting run in Android Studio should build the app, not complain about missing google play services json file and run the app on an emulator/device.
- [ ] The Bitrise.yml step is updated to prevent the task from running twice

pdrj:
### Summary
This makes `prepareGooglePlayServices` task run as a dependency of tasks that need it.

It has to be done with `whenTaskAdded`. When I just tried to filter `tasks`, the ones dealing with google play services were not available yet.

### Relationships
Closes #61 

---
### Pull Request checklist

### Pointers
Clone this branch locally, import in AS, run in emulator. Command line `./gradlew assembleDebug` doesn't work initially (because of a different error); after a build in AS, command line works OK.

### Review template

## Merge permissions
Me


daividssilverio:
It would be nicer if the task only run if the `google-services.json` file wasn't there... it's running unnecessarily for pretty much each build, isn't it?

PR #72, opened 2020-03-18
Add a higher order reducer to prevent premium actions from being forwarded

### LINKED ISSUE #58: Add a higher order reducer that prevents paid feature actions from being forwarded
## Description
For the sake of simplicity and cleanliness, instead of checking for premium on every single reducer we should simply add a higher order reducer that will block all premium actions from being further emitted when the user does not have the correct requirements.

We will take a similar approach for preventing teapots/deprecated apis from going further into the app.

## Definition of done
- [ ] We have a higher order reducer that checks the paid feature requirements of each action and forwards them only if proper requirements are met
- [ ] The logging reducer is turned into a higher order reducer to prevent it from not logging the actions filtered out

heytherewill:
### Summary
This adds a higher order reducer to check for premium actions and not allow them to go through if the user doesn't have a premium plan

### Technical considerations
I'm using Dagger to solve the dependencies. iOS is using a language, but I find having Dagger solving it for us is much more trustworthy

We could have some DSL for creating the main reducer, but I find that not needed

### Relationships
Closes #58 

---
### Pull Request checklist
> No code in there yet, so no test yet

### Pointers
Check and see if my solution is clear enough and point out anything you'd do differently

### Review template

## Merge permissions
Only myself


app/src/main/java/com/toggl/domain/reducers/LoggingReducer.kt
```diff
@@ -13,7 +12,9 @@ import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
-class LoggingReducer @Inject constructor() : Reducer<AppState, AppAction> {
+class LoggingReducer @Inject constructor(
+    private val innerReducer: FeatureAvailabilityReducer
```
semanticer: I think we should generalize this concept of wrapping (decorating) reducers with other reducers to make it easier to change what we are wrapping in what. I don't see any reason for `LogginReducer` to be tied to`FeatureAvailabilityReducer`. 
To make this decorating pattern even more explicit, we could introduce a new interface, for example:
```kotlin
interface DecoratorReducer<State, Action>: Reducer<State, Action> {
    val decoratedReducer: Reducer<State, Action>
}
```
`LoggingReducer` would than look like: 

```kotlin
@Singleton
class LoggingReducer @Inject constructor(
    override val decoratedReducer: Reducer<AppState, AppAction>
) : DecoratorReducer<AppState, AppAction> {
    override fun reduce(
        state: SettableValue<AppState>,
        action: AppAction
    ): List<Effect<AppAction>> {
        Log.i(
            "LoggingReducer", when (action) {
                is AppAction.Onboarding -> action.onboarding.formatForDebug()
                is AppAction.Timer -> action.timer.formatForDebug()
                AppAction.Load -> "Initial load of entities"
                is AppAction.EntitiesLoaded -> "Entities loaded"
            }
        )

        return decoratedReducer.reduce(state, action)
    }
}
```
And the same thing for `FeatureAvailabilityReducer`.
With this, we can manipulate the hierarchy of reducers just by changing `@Named()` parameter.

heytherewill: While I agree that an explicit interface might be better and that it feel weird to have the Logging reducer explicitly as the main reducer, how would you propose we name the dependencies? The only issue I see here is, as usual, naming. The Logging reducer could easily be called `outermostReducer`, but what about the one that gets injected inside it?

`@Named("reducerThatGoesInsideLogging")`? `@Named("reducerAfterLogging")`?

semanticer: I think we should be specific with the Dagger naming so `@Named("FeatureAvailabilityReducer")` is completely fine. We only have to worry about classes being generic enough, but the Dagger meta-information can be concrete because it's easy to change down the road => just changing the strings and everything should work.

app/src/main/java/com/toggl/di/ReducerModule.kt
```diff
@@ -0,0 +1,55 @@
...
+        createAppReducer(
+            entityLoadReducer,
+            onboardingReducer,
+            timerReducer,
+            analyticsReducer
+        )
+
+    @Provides
+    @Singleton
+    @LoggingReducerDependency
```
semanticer: I'm not sure about the naming you chose for these dependency annotations. You have a provider named `featureCheckReducer` (btw why not`FeatureAvailabilityReducer` ?) but you forever tied this to Logging with the annotation name `@LoggingReducerDependency`. Providers should be universal, for example, what if some other component needs to inject `FeatureAvailabilityReducer` ? 
So I would just use annotation names that actually describe what they provide not the other way around.

heytherewill: If I name according to what they provide I then need to change the class itself whenever I need to change the order of reducers (as in, open the reducer file and change the parameter it depends on), which is one of the things we were trying to avoid. The way things are right now, I just need to change the `ReducerModule` to move the `LoggingReducerDependency`  annotation to other reducer et voilà.

I totally agree with you here with the philosophy of naming the dependencies the way you mention, BTW, I just find it very unlikely that we will be injecting reducers anywhere else. I don't have strong opinions either way, so I'll let you decide if you want me to change that or not under the light of my reasons and, if you so wish, I can change it

The `featureCheckReducer` thing is just a mess up, so I'll change it anyways 😂

semanticer: change it anyways - even though it's unlikely we will inject these reducers, we should use these annotations consistently across our codebase and I believe this is the correct way. 
And think that changing annotation for one parameter is a similar effort as making changes in `ReducerModule`

PR #76, opened 2020-03-19
Add workspace persisting

### LINKED ISSUE #74: Add and persist the Workspace Model
## Description
We need a workspace DAO and we need it now

## Dependencies
Needed in order to perform the checks for #59 

## Definition of done
- [ ] The Workspace Model is created in the models package
- [ ] The Workspace Model is persisted accordingly in the database
- [ ] The Relationships between workspace and time entry are set
- [ ] The Relationships between workspace and project are set
- [ ] A load workspaces effect should be created and it should create a fake workspace if none is found on the DB. This is to simplify things until sync comes along.

heytherewill:
### Summary
This adds the Workspace entity and workspace persisting

### Technical considerations
Right now we are auto-adding a workspace when none is found. We'll change that later

### Relationships]
Closes #74 

---
### Pull Request checklist

### Pointers
Review per commit. Also, the `Add workspace fetching to the Repository` commit has a little refactor.

### Review template

## Merge permissions
Myself


app/src/main/java/com/toggl/domain/AppAction.kt
```diff
@@ -1,12 +1,14 @@
 package com.toggl.domain
 
 import com.toggl.models.domain.TimeEntry
+import com.toggl.models.domain.Workspace
 import com.toggl.onboarding.domain.actions.OnboardingAction
 import com.toggl.timer.common.domain.TimerAction
 
 sealed class AppAction {
     object Load : AppAction()
-    data class EntitiesLoaded(val timeEntries: List<TimeEntry>) : AppAction()
```
semanticer: I thought the idea was to have one single action for the initial load or that was only because we didn't haw batching yet?

database/src/main/java/com/toggl/database/dao/WorkspaceDao.kt
```diff
@@ -0,0 +1,32 @@
...
+@Dao
+interface WorkspaceDao {
+    @Query("SELECT * FROM Workspace")
+    fun getAll(): List<Workspace>
+
+    @Query("SELECT * FROM Workspace WHERE id = :id")
+    fun getOne(id: Long): Workspace
+
+    @Insert
+    fun insertAll(vararg projects: Workspace): List<Long>
```
semanticer: Projects are workspaces now? Or you just copy-pasted `projectDao` :)

models/src/main/java/com/toggl/models/domain/TimeEntry.kt
```diff
@@ -9,16 +9,19 @@ import org.threeten.bp.OffsetDateTime
 
 @Entity(
     foreignKeys = [
-        ForeignKey(entity = Project::class, parentColumns = ["id"], childColumns = ["projectId"])
+        ForeignKey(entity = Project::class, parentColumns = ["id"], childColumns = ["projectId"]),
+        ForeignKey(entity = Workspace::class, parentColumns = ["id"], childColumns = ["workspaceId"])
```
semanticer: I think this ForeignKey to Workspace is redundant. Because of: Project -> Workspace

app/src/main/java/com/toggl/domain/AppAction.kt
heytherewill: Yeah, now that we have batching we can have the luxury of having multiple things running

database/src/main/java/com/toggl/database/dao/WorkspaceDao.kt
heytherewill: ©️ 🍝

models/src/main/java/com/toggl/models/domain/TimeEntry.kt
heytherewill: What about time entries with no project?

semanticer: hmm... true.
Be we should find some way to make sure we have no inconsistencies in our database. Like TimeEntries with a project from different workspace than the one specified directly in TimeEntry entity... 
But I guess we would have to use some SQL triggers or placeholder empty projects for every workspace 🤮

PR #77, opened 2020-03-19
Add grouping logic

### LINKED ISSUE #23: Add logic for grouping time entries
## Description
We need to add the logic that groups similar time entries, similar to what [we have in the current app](https://github.com/toggl/mobileapp/blob/83418effb77e5957a68679c0ab2c4fe49ec36512/Toggl.Core.Tests/UI/Transformations/TimeEntriesFlatenningTests.cs). Since there's no switch to enable/disable grouping, we should leave grouping on by default until we start implementing the settings view

## Dependencies
#20, #21 and #44

## Definition of done
- [ ] Similar time entries are turned into a group
- [ ] Tests ensure the grouping logic is correct

semanticer:
### Summary
I implemented group time entries functionality as a part of our `timeEntriesLogSelector` 

### Technical considerations
I implemented this inside of the existing sequence constructor in `timeEntriesLogSelector` because half of the data processing I needed to group time entries was already there. Otherwise, I would have to "open" and group the stream later down the chain again and repeat some of the data processing again. So this seemed like the most efficient way to do this. 

I also slightly refactored `timeEntriesLogSelector` as I needed to share some of the functionality. Moreover, I was able to separate the code into two levels of detail where the main chain processing is now describing higher-level processes, whereas helper inner functions are going into more detail.

I already explained some of the reasoning behind the decision to process groups in our existing `timeEntriesLogSelector` method but let me know if you have any suggestions or complaints.

### Relationships

Closes #23

---
### Pull Request checklist

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogSelector.kt
```diff
@@ -13,48 +14,107 @@ fun timeEntriesLogSelector(
...
+        var result = description.hashCode()
+        result = 31 * result + billable.hashCode()
+        result = 31 * result + (projectId?.hashCode() ?: 0)
+        result = 31 * result + (taskId?.hashCode() ?: 0)
+        return result
+    }
+
+    fun List<TimeEntry>.mapToSortedGroups(): List<List<TimeEntry>> =
+        this.groupBy(TimeEntry::similarityHashCode)
+            .map { (_, timeEntries) -> timeEntries.sortedByDescending { it.startTime } }
```
heytherewill: Wouldn't it be more efficient to just sort everything before?

timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogSelector.kt
```diff
@@ -13,48 +14,107 @@ fun timeEntriesLogSelector(
...
+    fun TimeEntry.similarityHashCode(): Int {
+        var result = description.hashCode()
+        result = 31 * result + billable.hashCode()
+        result = 31 * result + (projectId?.hashCode() ?: 0)
+        result = 31 * result + (taskId?.hashCode() ?: 0)
+        return result
+    }
+
+    fun List<TimeEntry>.mapToSortedGroups(): List<List<TimeEntry>> =
+        this.groupBy(TimeEntry::similarityHashCode)
```
heytherewill: Does the Kotlin grouping algorithm ensure order? 🤔

timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogSelector.kt
```diff
@@ -13,48 +14,107 @@ fun timeEntriesLogSelector(
...
+                duration = timeEntries.totalDuration(),
+                project = getProjectViewModelFor(firstTimeEntry),
+                billable = firstTimeEntry.billable
+            )
+        )
+    }
+
     return timeEntries.values
         .filter { it.duration != null }
+        .filter { !it.isDeleted }
```
heytherewill: Ideally we shouldn't even be getting these here in the first place, but it's better safe than sorry.

Could you add it to the filter above, though, instead of having two different filters?

timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogSelector.kt
```diff
@@ -13,48 +14,107 @@ fun timeEntriesLogSelector(
...
-                                ?: throw IllegalStateException("You can't display a running time entry in the log"),
-                            project = project,
-                            billable = timeEntry.billable
-                        )
-                    )
-                }
+            sequence<TimeEntryViewModel> {
+                yieldDayHeader(groupDate, timeEntries)
+
+                if (shouldGroup)
```
heytherewill: I really dislike multi-line statements inside ifs without braces. It adds to the cognitive load

heytherewill: I'm also not sure about these `forEachs`. They try to look function but are basically side effects (and the syntax is not the best since you are using lambdas). What's wrong with the for loop?

timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogFragment.kt
```diff
@@ -52,7 +52,7 @@ class TimeEntriesLogFragment : Fragment(R.layout.time_entries_log_fragment) {
         val yesterdayString = context.getString(R.string.yesterday)
 
         val curriedTimeEntriesSelector: suspend (TimeEntriesLogState) -> List<TimeEntryViewModel> = {
-            timeEntriesLogSelector(it.timeEntries, it.projects, timeService, todayString, yesterdayString)
+            timeEntriesLogSelector(it.timeEntries, it.projects, timeService, todayString, yesterdayString, false)
```
heytherewill: I'd rather have them always grouping, since it allows us to test everything related to the grouping logic

timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogSelector.kt
semanticer: I guess the idea was to be consistent since I use `forEach` right above as part of the chain. But this kind of consistency probably doest really help in any way.

timer/src/main/res/layout/time_entries_group_item.xml
```diff
@@ -0,0 +1,219 @@
...
+    <androidx.constraintlayout.widget.ConstraintLayout
+        android:id="@+id/content_view"
+        android:elevation="1dp"
+        android:background="@color/background"
+        android:layout_height="@dimen/time_entry_log_item_height"
+        android:layout_width="match_parent">
+        <androidx.constraintlayout.widget.Guideline
+            android:id="@+id/top_guideline"
+            android:orientation="horizontal"
+            app:layout_constraintGuide_begin="8dp"
```
heytherewill: Let's extract this and reuse it both here and on the other non-grouped cell 👀

timer/src/main/res/layout/time_entries_group_item.xml
```diff
@@ -0,0 +1,219 @@
...
+        android:visibility="invisible"
+        android:gravity="center_vertical|end"
+        android:fontFamily="sans-serif-medium"
+        android:textColor="@android:color/white"
+        android:background="@color/stop_time_entry_button_background"
+        android:layout_width="match_parent"
+        android:layout_height="match_parent" />
+    <androidx.constraintlayout.widget.ConstraintLayout
+        android:id="@+id/content_view"
+        android:elevation="1dp"
```
heytherewill: Let's use the constants here as well.

Though, does elevation even apply here?

timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogFragment.kt
semanticer: The issue was that grouping crashes the Adapter since we do not support group type views. I added a small dummy implementation for debugging and enabled grouping.

timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogSelector.kt
heytherewill: You didn't reply here

timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogSelector.kt
semanticer: 🤔 from groupBy docs: 

> The returned map preserves the entry iteration order of the keys produced from the original collection.

semanticer:
@heytherewill Do you have any suggestions where to put test data? TimeEntriesLogSelectorTests looks a bit crowded right now. 
As far as the layout is concerned, I added our classic counter and removed that "G..." debug string. I also implemented continue action and prepared expanding listener.
So the layout is usable for now but we should definitely do MUCH more work to share code between single and group layouts, therefore I would prefer to address this in separate issue/PR.

PR #89, opened 2020-03-25
Update dependencies

### LINKED ISSUE #91: Update dependencies
## Description
Run `dependencyUpdate` Gradle task and update everything to new stable versions. In case we already use some beta version of a library it's ok to update it to newer beta/RC.

## Definition of done
- [x] All dependencies are updated

semanticer:
### Summary
I updated a few dependencies only to stable versions or newer beta versions if we were using beta before. I especially needed this because the last version of Kotlin was buggy in our tests.

### Relationships
Closes #91 

## Merge permissions
Anyone


heytherewill:
Updated to cover our reviews, @semanticer. Still missing:

- The name of the actions should be rendered as well
- Daivid needs to clarify one of the comments in your review

PR #90, opened 2020-03-25
Timer start domain tests

### LINKED ISSUE #40: Ensure that all actions have tests
## Description
Some actions (namely the ones added during the prototype phase) don't have proper tests associated to them. Let's fix this.

## Definition of done
- [x] All actions related to the `TimeEntriesLog` feature should have tests
- [ ] All actions related to the `StartTimeEntry` feature should have tests

semanticer:
### Summary
This PR adds missing tests for StartTimeEntry Actions 

### Technical considerations
I added `Reducer.testReduce` extension to simplify and increase the readability of our test code. If we decide to use this method, it should be probably moved to some common test package. 

I decided not to test Effects here until we decide how to proceed.

### Relationships
Closes #40

---
### Pull Request checklist

### Pointers
Make sure all the tests make sense.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


timer/src/test/java/com/toggl/timer/start/domain/TimeEntryDescriptionChangedActionTests.kt
```diff
@@ -0,0 +1,36 @@
...
+    val reducer = StartTimeEntryReducer(repository)
+    coEvery { workspace.id } returns 1
+
+    "The TimeEntryDescriptionChangedAction action" - {
+        reducer.testReduce(
+            initialState = state,
+            action = StartTimeEntryAction.TimeEntryDescriptionChanged("new description")
+        ) { state, effect ->
+            "should change EditableTimeEntry's description" {
+                state.editableTimeEntry shouldBe editableTimeEntry.copy(description = "new description")
```
heytherewill: Why not:
```suggestion
                state.editableTimeEntry.description shouldBe "new description"
```
?

semanticer: I guess the idea was also to check that nothing other than description changed in editableTimeEntry. But I guess that is outside of this test's responsibility. I'm gonna change it as you suggested.

PR #104, opened 2020-03-30
✨ Display the billable icon as necessary

### LINKED ISSUE #82: Reflect the billable flag in the time entries log
## Description
We need to ensure that time entries (both groups and flat ones) display the little billable icon when the time entry is marked as billable

## Definition of done
- [ ] When a time entry is billable,  the view that indicates it in the log is visible
- [ ] When a time entry is not billable, the view that indicates it is hidden
- [ ] *STRETCH GOAL* For performance reasons, it might make sense to implement the billable view as a ViewStub, to prevent inflating it all the time. Let's not do it if it's not worth the hassle, though

pdrj:
### Summary
This hides/shows the billable icon as necessary.

### Technical considerations
Using a ViewStub, because performance.

Because performance, but I've not measured this and think that it might be just a marginal gain :). 

### Relationships

Closes #82 

---
### Pull Request checklist

### Pointers
Create some TEs with billable flag set. I did: `.map { it.copy(duration = Duration.between(it.startTime, now), billable = Random.nextBoolean()) }` in `Repository` when stopping TEs.

Then, validate the ones that should have the billable icon displayed, indeed have it visible _and_ the ones that don't don't.

### Review template

- [ ] The code considers i18n

## Merge permissions
Me.


timer/src/main/res/layout/time_entries_log_item_billable_icon.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<ImageView xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:id="@+id/billable_icon"
+    android:layout_width="18dp"
+    android:layout_height="18dp"
+    android:layout_marginEnd="4dp"
+    android:src="@drawable/ic_billable"
+    android:tint="@color/icon"
+    app:layout_constraintBottom_toBottomOf="@+id/duration"
```
daividssilverio: Are the constraint attributes necessary here?

daividssilverio: removing them from the file and keeping the constraint attrs in the view that actually uses the stub seems to be enough.

pdrj: I thought I tried that and it didn't work - I'll recheck and delete these if it works.

pdrj:
> Why not do the same in the `TimeEntryGroupViewHolder`?

Because I somehow thought we don't display the icon for groups 😄

PR #105, opened 2020-03-31
Add and load tags

### LINKED ISSUE #85: Add and persist the Tag model
## Description
We need a tag DAO

## Dependencies
Depends on #84 

## Definition of done
- [x] The Tag Model is created in the models package
- [x] The Tag Model is persisted accordingly in the database
- [x] The Relationships between tag and workspace are set
- [ ] The Relationships between tag and time entry are set
- [x] A load tags effect should be created 
- [x] The load tags effect should be added to the initial load of the app

daividssilverio:
### Summary
This PR:
- Adds the entity for the tags used in time entries 
- Adds the state, action and effect needed to load tags
- Defines the relationship between time entries and tags

### Technical considerations
Please consider the limitations of Room when reviewing this PR.
Refer to https://developer.android.com/training/data-storage/room/relationships.

The tags themselves (without the relationship with the time entries) are pretty much boilerplate.
The relationship, we still need to discuss how we are actually implementing it.

### Relationships
Closes #85

---
### Pull Request checklist

### Pointers
You can check that everything works by manually adding tags using the repository during debug.

PS*: I think we should define some actual database tests to run on devices that actually create sqlite tables, just to make sure everything is working as we would expect.

#### Definition of done:
- [ ] The Tag Model is created in the models package
- [ ] The Tag Model is persisted accordingly in the database
- [ ] The Relationships between tag and workspace are set
- [ ] The Relationships between tag and time entry are set
- [ ] A load tags effect should be created
- [ ] The load tags effect should be added to the initial load of the app

### Review template

## Merge permissions
🦑 it


daividssilverio:
@semanticer @heytherewill @Odrakir 

Didn't work exactly the way I thought it would, but I think we managed to get into a point where it might work alright.

I was talking with @heytherewill and what I'm currently doing here doesn't sound right... It might be best to create database models separate from the actual models used throughout the app.

We could have it only for the TimeEntry and the tags since they have an m:n relationship.
Or we could go all on with every model.

We would have `DatabaseTimeEntry`, `DatabaseTag`, `DatabaseTimeEntryTag`, `DatabaseProject` ... and that would be used by `Room` to do whatever queries it needs to do.
On the actual model's module, we would have `TimeEntry` (with array of tag ids, etc), `Tag`, `Project`...

This way we would need to expose how the database works during the mapping to the `FlatTimeEntryViewModel`. We would receive time entry with tags from the Dao.

What do you think?

daividssilverio:
Please check https://github.com/toggl/android/tree/doivid/tags-on-steroids to see how it would like (at least on android)

Odrakir:
That's more or less what I was planning for iOS as well. Keep relationships in the DB but map to/from repository (or domain) models when moving through that boundary. I don't think we'll need `DatabaseTimeEntryTag` though, because our DAOs are NSManagedObjects (CoreData objects) and will have the concrete relationships that just need to be explicitly fetched.

TL;DR: Looks good.

semanticer:
We discussed this separation with @heytherewill during the prototyping phase and I was slightly against it. I still think it's a ton of boilerplate that at least on android has currently only one practical benefit: 
Instead of this nested model
```kotlin
data class DBTimeEntryWithTags(
    val timeEntry: DBTimeEntry,
    var tags: List<Long>
)
```
we will have this nicer flatter version in layers above the repository:
```kotlin
data class TimeEntry(
    val id: Long = 0,
    val description: String,
    val startTime: OffsetDateTime,
    val workspaceId: Long,
    val projectId: Long?,
    val taskId: Long?,
    val isDeleted: Boolean
    val tagIds: List<Long>
)
```

It definitely looks cleaner to have this extra separation between our layers but the practical benefit is still small as far as I can tell at this moment.

Having said that, my perspective is coming from much more dynamic development where models are changing quite often and the cost of changing all the boilerplate is much higher than in our app where entries can survive decades with very little change, so I think we can definitely afford this, and we probably should.

pdrj:
I'm not a huge fan of having this split TBH. There's just so much duplication and mapping (copying) that it makes me feel uneasy. 

If that doesn't have too much of a performance impact, I guess it's fine. As said by Tomas, our models are quite stable so the duplicated part won't have to be touched often.

daividssilverio:
So, are we fine with what we have on the https://github.com/toggl/android/tree/doivid/tags-on-steroids ?
If so, I'll clean up that branch and make the changes here as necessary.

@pdrj @semanticer @heytherewill 

If the mapping is too much of a problem, we could simplify it a bit by having only the problematic models   have a `DB_Model`. That would reduce the necessary mappings only to the TimeEntry, Tags and their relationship table. The domain would use direct Room Entities for mostly everything, except by the TimeEntry which would look exactly like we have my experimental branch.

The problem I see with this idea is that we'd have mixed classes spread around multiple modules and the domain models wouldn't be "pure".

semanticer:
I would go one way or the other. Mixing both is definitely the worst option.

daividssilverio:
Alright, with the latest changes, our database schema looks like this:

```
CREATE TABLE `time_entries` (`id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, `description` TEXT NOT NULL, `startTime` INTEGER NOT NULL, `duration` INTEGER, `billable` INTEGER NOT NULL, `workspaceId` INTEGER NOT NULL, `projectId` INTEGER, `taskId` INTEGER, `isDeleted` INTEGER NOT NULL, FOREIGN KEY(`projectId`) REFERENCES `projects`(`id`) ON UPDATE NO ACTION ON DELETE NO ACTION , FOREIGN KEY(`workspaceId`) REFERENCES `workspaces`(`id`) ON UPDATE NO ACTION ON DELETE NO ACTION );
CREATE INDEX `index_time_entries_projectId` ON `time_entries` (`projectId`);
CREATE TABLE `projects` (`id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, `name` TEXT NOT NULL, `color` TEXT NOT NULL, `active` INTEGER NOT NULL, `isPrivate` INTEGER NOT NULL, `billable` INTEGER, `workspaceId` INTEGER NOT NULL, `clientId` INTEGER, FOREIGN KEY(`workspaceId`) REFERENCES `workspaces`(`id`) ON UPDATE NO ACTION ON DELETE NO ACTION , FOREIGN KEY(`clientId`) REFERENCES `clients`(`id`) ON UPDATE NO ACTION ON DELETE NO ACTION );
CREATE TABLE `workspaces` (`id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, `name` TEXT NOT NULL, `features` TEXT NOT NULL);
CREATE TABLE `clients` (`id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, `name` TEXT NOT NULL, `workspaceId` INTEGER NOT NULL, FOREIGN KEY(`workspaceId`) REFERENCES `workspaces`(`id`) ON UPDATE NO ACTION ON DELETE NO ACTION );
CREATE TABLE `tags` (`id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, `name` TEXT NOT NULL, `workspaceId` INTEGER NOT NULL, FOREIGN KEY(`workspaceId`) REFERENCES `workspaces`(`id`) ON UPDATE NO ACTION ON DELETE NO ACTION );
CREATE TABLE `time_entries_tags` (`timeEntryId` INTEGER NOT NULL, `tagId` INTEGER NOT NULL, PRIMARY KEY(`timeEntryId`, `tagId`));
CREATE TABLE room_master_table (id INTEGER PRIMARY KEY,identity_hash TEXT);
```

Independent of how we setup/organize the models using Room or some other persistency library, we can make further changes having that in mind.

database/src/main/java/com/toggl/database/dao/TimeEntryDao.kt
```diff
@@ -3,33 +3,70 @@ package com.toggl.database.dao
...
+    fun insertTimeEntryWithTags(timeEntryWithTags: DatabaseTimeEntryWithTags) {
+        val te = insertTimeEntry(timeEntryWithTags.timeEntry)
+        clearDeletedRelationships(te, timeEntryWithTags.tags)
+        insertAllTimeEntryTagsPairs(timeEntryWithTags.tags.map {
+            DatabaseTimeEntryTag(te, it)
+        })
+    }
+
+    @Query("DELETE FROM time_entries_tags WHERE timeEntryId = :timeEntryId AND tagId NOT IN (:tagIds)")
+    fun clearDeletedRelationships(timeEntryId: Long, tagIds: List<Long>)
```
pdrj, line 68: Hm, I'm not super knowledgable about this - but are we sure there's no way for room to do this automagically for us?

database/src/main/java/com/toggl/database/dao/TimeEntryDao.kt
```diff
@@ -3,33 +3,70 @@ package com.toggl.database.dao
...
+        insertAllTimeEntryTagsPairs(databaseTimeEntryWithTags.tags.map {
+            DatabaseTimeEntryTag(databaseTimeEntryWithTags.timeEntry.id, it)
+        })
+    }
+
+    @Transaction
+    @Insert
+    fun insertTimeEntryWithTags(timeEntryWithTags: DatabaseTimeEntryWithTags) {
+        val te = insertTimeEntry(timeEntryWithTags.timeEntry)
+        clearDeletedRelationships(te, timeEntryWithTags.tags)
```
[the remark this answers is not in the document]
daividssilverio: Yeah, nobody is using it yet. I can 🔥 it and someone can add it again when we actually have a `createTimeEntry` action/effect, probably for when we have the fully functional start/edit views.

database/src/main/java/com/toggl/database/models/DatabaseUser.kt
```diff
@@ -0,0 +1,5 @@
+package com.toggl.database.models
+
+import com.toggl.models.validation.ApiToken
+
+data class DatabaseUser(val apiToken: ApiToken)
```
heytherewill, line 5: I'm not even sure the user will be in the DB. We can always only have one and we need it to load :sanic:, so maybe there's a case to store it in a faster encrypted storage. Either way, since this is not an entity yet, maybe we can 🔥 it until we care about it?

database/src/main/java/com/toggl/database/dao/TimeEntryDao.kt
daividssilverio: Not from the research I did. It would be great we had it that way though...
I think we can figure out something with the foreign keys and trigger deletion in the relationship table when a time entry is actually deleted. 🤔 I'll give it a try.

database/src/main/java/com/toggl/database/dao/TimeEntryTagDao.kt
```diff
@@ -0,0 +1,16 @@
...
+import androidx.room.OnConflictStrategy
+import androidx.room.Query
+import com.toggl.database.models.DatabaseTimeEntryTag
+
+@Dao
+interface TimeEntryTagDao {
+    @Query("SELECT * FROM time_entries_tags")
+    fun getAll(): List<DatabaseTimeEntryTag>
+
+    @Insert(onConflict = OnConflictStrategy.IGNORE)
```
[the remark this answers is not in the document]
daividssilverio: In case we try to add a duplicated entry in the `TimeEntryTag` relationship table it wouldn't fail.

database/src/main/java/com/toggl/database/dao/TimeEntryTagDao.kt
```diff
@@ -0,0 +1,16 @@
+package com.toggl.database.dao
+
+import androidx.room.Dao
+import androidx.room.Insert
+import androidx.room.OnConflictStrategy
+import androidx.room.Query
+import com.toggl.database.models.DatabaseTimeEntryTag
+
+@Dao
+interface TimeEntryTagDao {
```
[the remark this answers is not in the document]
daividssilverio: I don't have a particular opinion on this as well.
I'm currently leaning towards not having this DAO, maybe it makes more sense for everything to stay on `TimeEntryDao` only since it we'll be only using this entity when handling time entries as well.

database/src/main/java/com/toggl/database/models/DatabaseUser.kt
daividssilverio: I agree, lets 🔥 it.

database/src/main/java/com/toggl/database/dao/TimeEntryTagDao.kt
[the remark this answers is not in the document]
heytherewill: Agreed

database/src/main/java/com/toggl/database/dao/TimeEntryDao.kt
```diff
@@ -3,33 +3,70 @@ package com.toggl.database.dao
...
 @Dao
 interface TimeEntryDao {
-    @Query("SELECT * FROM TimeEntry WHERE NOT isDeleted")
-    fun getAll(): List<TimeEntry>
+    // TimeEntries only
 
-    @Query("SELECT * FROM TimeEntry WHERE NOT isDeleted AND duration is null")
-    fun getAllRunning(): List<TimeEntry>
+    @Query("SELECT * FROM time_entries WHERE NOT isDeleted")
+    fun getAllTimeEntries(): List<DatabaseTimeEntry>
```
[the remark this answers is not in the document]
daividssilverio: This method in specific doesn't need to exist anymore, but the `getAllRunningTimeEntries` right bellow is used in the repository when starting/stopping time entries (although I think we could do the work there with a single query instead).

If we change how the `Repository:stopRunningTimeEntry` and `Repository:startTimeEntry` to use simple queries, we don't need this method at all.

`getAllRunningTimeEntries` is faster because we don't need to query tags to stop/start a time entry.

database/src/main/java/com/toggl/database/dao/TimeEntryDao.kt
daividssilverio: Yep, that's not needed if we have the right constraints set. Done it.

daividssilverio:
Btw, that's how our schema looks after all the changes:
```
CREATE TABLE IF NOT EXISTS `time_entries` (`id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, `description` TEXT NOT NULL, `startTime` INTEGER NOT NULL, `duration` INTEGER, `billable` INTEGER NOT NULL, `workspaceId` INTEGER NOT NULL, `projectId` INTEGER, `taskId` INTEGER, `isDeleted` INTEGER NOT NULL, FOREIGN KEY(`projectId`) REFERENCES `projects`(`id`) ON UPDATE NO ACTION ON DELETE NO ACTION , FOREIGN KEY(`workspaceId`) REFERENCES `workspaces`(`id`) ON UPDATE NO ACTION ON DELETE NO ACTION )"
CREATE INDEX IF NOT EXISTS `index_time_entries_projectId` ON `time_entries` (`projectId`)"
CREATE TABLE IF NOT EXISTS `projects` (`id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, `name` TEXT NOT NULL, `color` TEXT NOT NULL, `active` INTEGER NOT NULL, `isPrivate` INTEGER NOT NULL, `billable` INTEGER, `workspaceId` INTEGER NOT NULL, `clientId` INTEGER, FOREIGN KEY(`workspaceId`) REFERENCES `workspaces`(`id`) ON UPDATE NO ACTION ON DELETE NO ACTION , FOREIGN KEY(`clientId`) REFERENCES `clients`(`id`) ON UPDATE NO ACTION ON DELETE NO ACTION )"
CREATE TABLE IF NOT EXISTS `workspaces` (`id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, `name` TEXT NOT NULL, `features` TEXT NOT NULL)"
CREATE TABLE IF NOT EXISTS `clients` (`id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, `name` TEXT NOT NULL, `workspaceId` INTEGER NOT NULL, FOREIGN KEY(`workspaceId`) REFERENCES `workspaces`(`id`) ON UPDATE NO ACTION ON DELETE NO ACTION )"
CREATE TABLE IF NOT EXISTS `tags` (`id` INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, `name` TEXT NOT NULL, `workspaceId` INTEGER NOT NULL, FOREIGN KEY(`workspaceId`) REFERENCES `workspaces`(`id`) ON UPDATE NO ACTION ON DELETE NO ACTION )"
CREATE TABLE IF NOT EXISTS `time_entries_tags` (`timeEntryId` INTEGER NOT NULL, `tagId` INTEGER NOT NULL, PRIMARY KEY(`timeEntryId`, `tagId`), FOREIGN KEY(`timeEntryId`) REFERENCES `time_entries`(`id`) ON UPDATE NO ACTION ON DELETE CASCADE , FOREIGN KEY(`tagId`) REFERENCES `tags`(`id`) ON UPDATE NO ACTION ON DELETE CASCADE )"
```

If you need to test the table locally, don't forget to set `PRAGMA foreign_keys = 1` otherwise sqlite will ignore foreign keys.

database/src/main/java/com/toggl/database/dao/TimeEntryDao.kt
```diff
@@ -3,33 +3,53 @@ package com.toggl.database.dao
...
 @Dao
 interface TimeEntryDao {
-    @Query("SELECT * FROM TimeEntry WHERE NOT isDeleted")
-    fun getAll(): List<TimeEntry>
+    // TimeEntries only
 
-    @Query("SELECT * FROM TimeEntry WHERE NOT isDeleted AND duration is null")
-    fun getAllRunning(): List<TimeEntry>
+    @Query("SELECT * FROM time_entries WHERE NOT isDeleted AND duration is null")
+    fun getAllRunningTimeEntries(): List<DatabaseTimeEntry>
```
semanticer, line 19: Why did you rename all these methods?

repository/src/test/java/com/toggl/repository/RepositoryTest.kt
```diff
@@ -66,8 +72,23 @@ class RepositoryTest : StringSpec() {
...
+            val loaded = repository.loadTags()
+
+            verify(exactly = 1) { tagDao.getAll() }
+            verify {
+                workspaceDao wasNot called
+                timeService wasNot called
+                timeEntryDao wasNot called
+                projectDao wasNot called
+            }
+            loaded shouldBe listOf()
```
semanticer, line 87: There is `.shouldBeEmpty()`

repository/src/test/java/com/toggl/repository/RepositoryTest.kt
```diff
@@ -66,8 +72,23 @@ class RepositoryTest : StringSpec() {
             loaded shouldBe listOf()
```
semanticer, line 72: There is `.shouldBeEmpty()`

database/src/main/java/com/toggl/database/dao/TimeEntryDao.kt
daividssilverio: To enforce the difference between the "time entries with tags" and "just time entries".

semanticer: I would probably still choose the "neutral" naming and simply use `getAllWithTags()` for "time entries with tags" but this is up to you.

PR #111, opened 2020-04-02
✨ Display client names in time entry list

### LINKED ISSUE #87: Display client names in the main log
## Description
The name of the client should appear next to the project name on both simple and grouped time entries

## Dependencies
Depends on #86 

## Definition of done
- [ ] The client name, when present, is displayed next to the project name in the main log

pdrj:
### Summary
This displays client names for TEs that have clients associated (via projects).

### Technical considerations
You'll see that the project name display length is now capped at 60% of available width, even when there's no client name to display. We could dynamically remove that constraint in case there's no client, but I don't think it's worth it.

We could maybe embed clients in project model, but I like the current approach more (no duplication, simple, flat, etc.).

### Relationships

Closes #87 

---
### Pull Request checklist

### Review template

- [ ] The code considers i18n

## Merge permissions
Me.


pdrj:
Screenshot 2020-04-02 at 10 01 08

timer/src/main/java/com/toggl/timer/log/domain/TimeEntryViewModel.kt
```diff
@@ -11,6 +11,7 @@ data class FlatTimeEntryViewModel(
     val startTime: OffsetDateTime,
     val duration: Duration,
     val project: ProjectViewModel?,
+    val client: ClientViewModel?,
```
heytherewill, line 14: While I agree with you that Flat Is Justice, this current model allows us to model literally invalid time entries. A client should never exist if a project is not there and we should model our projects accordingly.

As a matter of fact, I think the `ClientViewModel` is not even needed; we can simply add a `clientName: String?` property to the `ProjectViewModel` instead. If we need the Id, though, then please include a `client: ClientViewModel?` property inside `ProjectViewModel` to better reflect the reality of the model

pdrj:
> Why did you choose to add a whole new TextView if that makes the layout behave a bit worse? Wouldn't using a SpannableString with a couple of Spans for Project and client color be a better solution?
- I don't like playing with substrings, indexes etc. - there's always a way to mess it up 😄 In this case, the spans should be pretty easy though and it would indeed simplify the layout 🤔
- if I make the proposed change and the project name is very long, the client name will not be visible at all, whereas in my solution it will be at least partially visible.

Do you think that's better @heytherewill? I'm mainly referring to the second point.

heytherewill:
> if I make the proposed change and the project name is very long, the client name will not be visible at all, whereas in my solution it will be at least partially visible.

That's how we are doing it in the current app. We can always refer back to designer and maybe make this a `MarqueeView`, but for now I think the project name takes precedence :)

pdrj:
OK, I'll make the changes then. Thanks Will.

newbein123:
Thanks for letting the app to apear

pdrj:
@heytherewill it should have been two commits, but I messed something up when doing `--amend`...

timer/src/main/java/com/toggl/timer/log/ui/TimeEntryItemViewHolder.kt
```diff
@@ -25,16 +27,19 @@ class TimeEntryItemViewHolder(itemView: View, private val onContinueTappedListen
...
+            projectDescription.append(item.project.name)
+            projectDescription.setSpan(ForegroundColorSpan(item.project.color.toColorInt()), 0, item.project.name.length, Spannable.SPAN_EXCLUSIVE_EXCLUSIVE)
         }
-        project.isVisible = item.project != null
 
-        if (item.client != null) {
-            client.text = item.client.name
+        if (item.project?.clientName != null) {
+            projectDescription.append(" ")
+            projectDescription.append(item.project.clientName)
```
heytherewill, line 38: This check could be inside the `if (item.project != null)`, couldn't it?

timer/src/main/java/com/toggl/timer/log/ui/TimeEntryItemViewHolder.kt
```diff
@@ -25,16 +27,19 @@ class TimeEntryItemViewHolder(itemView: View, private val onContinueTappedListen
...
         }
-        client.isVisible = item.client != null
+
+        project.text = projectDescription
```
heytherewill, line 41: I'd rather have an extension method like:

```kotlin
fun ProjectViewModel.formatForDisplay() : Spannable
```

then you can simply:

```kotlin
project.text = item.project?.formatForDisplay() ?? ""
```

Or something like that. This is particularly beneficial because of the repetition. Also, it allows for proper testing

timer/src/main/java/com/toggl/timer/log/ui/TimeEntryItemViewHolder.kt
pdrj: Yes, but this would introduce one more nested layer, which I'm not a fan of. But yeah, I’m aware that I'm repeating the project check, so that's not good either.

timer/src/main/java/com/toggl/timer/log/ui/TimeEntryItemViewHolder.kt
pdrj: That's nice, thanks.

timer/src/main/java/com/toggl/timer/log/ui/TimeEntryItemViewHolder.kt
```diff
@@ -27,18 +24,7 @@ class TimeEntryItemViewHolder(itemView: View, private val onContinueTappedListen
...
-            projectDescription.setSpan(ForegroundColorSpan(item.project.color.toColorInt()), 0, item.project.name.length, Spannable.SPAN_EXCLUSIVE_EXCLUSIVE)
-        }
-
-        if (item.project?.clientName != null) {
-            projectDescription.append(" ")
-            projectDescription.append(item.project.clientName)
-        }
-
-        project.text = projectDescription
+        project.text = item.project?.formatForDisplay()
```
heytherewill, line 27: Won't setting the text to null be a problem? 🤔

pdrj: Nope [https://android.googlesource.com/platform/frameworks/base/+/jb-mr0-release/core/java/android/widget/TextView.java#3415](https://android.googlesource.com/platform/frameworks/base/+/jb-mr0-release/core/java/android/widget/TextView.java#3415)

But since that's relying on something that's not explicitly documented, I'll change that before merging.

pdrj: I simplified it further by having the extension on an optional.

PR #123, opened 2020-04-02
Show tags in the main log

### LINKED ISSUE #83: Reflect the existence of tags in the time entries log
## Description
We need to ensure that time entries (both groups and flat ones) display the little tag icon when they have any number of tags associated to them.

## Dependencies
Depends on #85

## Definition of done
- [x] When a time entry has any tag,  the view that indicates it in the log is visible
- [x] When a time entry has no tags, the view that indicates it is hidden
- [x] *STRETCH GOAL* For performance reasons, it might make sense to implement the billable view as a ViewStub, to prevent inflating it all the time. Let's not do it if it's not worth the hassle, though

daividssilverio:
### Summary
This PR makes the "has tags" icon appear in the time entry items in the main log when necessary.

### Technical considerations
None, this is way too simple.

Using a `ViewStub` seems like the way to go. We don't need to inflate what we aren't gonna use.

### Relationships
Closes #83

---
### Pull Request checklist

### Pointers
Manually add tags to your database.
Manually add the relationship between tags and time entries and see the magic.
Or you can just change the boolean in the view holders to see the icon appear.

### Review template

## Merge permissions
🦑 it


timer/src/main/java/com/toggl/timer/log/domain/TimeEntryViewModelExtensions.kt
```diff
@@ -22,7 +23,8 @@ fun List<TimeEntry>.toTimeEntryGroupViewModel(groupId: Long, isExpanded: Boolean
         description = first().description,
         duration = totalDuration(),
         project = projects.getProjectViewModelFor(this.first()),
-        billable = first().billable
+        billable = first().billable,
+        hasTags = first().tagIds.isNotEmpty()
```
heytherewill, line 27: Are we considering tags in the hash function for the groups?  🤔

pdrj: Good point.

I was actually thinking about this and clients too - but in that case since a project can have only one client, I think we should be good.

heytherewill: For the project we're fine because the `projectId` is being considered, yeah.
But we should add the hash of tags in this PR to ensure unity :)

daividssilverio: I've added the hash of the sum of the tags ids, do you think that's enough?

pdrj: There's `contentHashCode`, but we'd have to convert to array, which in this case is very yuck. Maybe we could use `fold` to combine the values' hashes? Or we leave it as is, since this function is imperfect anyway.

pdrj: Actually, I think that for lists, `hashCode` behaves like `contentHashCode`: [https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/src/kotlin/collections/AbstractList.kt#L135](https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/src/kotlin/collections/AbstractList.kt#L135). So I think the best way is just to use the hash code if the whole list here.

PR #126, opened 2020-04-03
Modal bottom sheet

### LINKED ISSUE #122: Refactor star/edit UI to use modal version of bottom sheet
## Description
Instead of using a bottom sheet that is part of the view hierarchy we should use more independent modal dialog bottom sheet
## Dependencies
#112 
## Definition of done
- [ ] Bottom sheet is refactored to use a modal dialog
- [ ] Bottom sheet should be dismissable by a back button tap
- [ ] Bottom sheet should be dismissable by a tap on log content
- [ ] Keyboard should align correctly with the edit text but with controls below still visible

semanticer:
### Summary
This PR moves the implementation of StartTimeEntry fragment from bottom sheet card to modal bottom sheet. This should give us all the behavior we need like proper dismiss actions, back button handling, proper keyboard behavior etc.

### Technical considerations
I'm using the navigation library to control this new dialog,  instead of managing it "by hand".
Leak Canary is reporting some memory leaks - I fixed some of them but there is still something wrong. Let me know if you find out something, otherwise, I will take a closer look on Monday again.

Not sure if there is any other way, but feel free to suggest anything

### Relationships

Closes #122 

---
### Pull Request checklist

### Pointers
Play with it, it's fun.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


app/src/main/res/navigation/tabs_nav_graph.xml
```diff
@@ -10,6 +10,13 @@
         <deepLink app:uri="toggl://timeentrieslog" />
     </fragment>
 
+    <dialog
+        android:id="@+id/my_dialog_fragment"
```
heytherewill, line 14: Please use a better ID

common-android/src/main/java/com/toggl/common/sheet/BottomSheetCallback.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    fun addOnSlideAction(action: OnSlideAction): Boolean {
+        return onSlideActions.add(action)
+    }
+
+    fun removeOnSlideAction(action: OnSlideAction): Boolean {
+        return onSlideActions.remove(action)
+    }
+
+    fun addOnStateChangedAction(action: OnStateChangedAction): Boolean {
```
heytherewill, line 27: Don't we need `setOnStateChangedAction` and `setOnSlideAction`? 🤔

common-android/src/main/java/com/toggl/common/sheet/OnSlideAction.kt
```diff
@@ -0,0 +1,26 @@
...
+    fun onSlide(
+        sheet: View,
+        @FloatRange(
+            from = -1.0,
+            fromInclusive = true,
+            to = 1.0,
+            toInclusive = true
+        ) slideOffset: Float
+    )
+}
```
heytherewill, line 16: Wouldn't it be better if this was a simple typealias?

`typealias OnSlideAction = (View, Float) -> Void`

common-android/src/main/java/com/toggl/common/sheet/OnSlideAction.kt
```diff
@@ -0,0 +1,26 @@
...
+
+class AlphaSlideAction(
+    private val view: View,
+    private val reverse: Boolean = false
+) : OnSlideAction {
+    override fun onSlide(sheet: View, slideOffset: Float) {
+        view.alpha = if (!reverse) slideOffset else 1F - slideOffset
+    }
+}
```
heytherewill, line 25: Shouldn't this be in its own file?

timer/src/main/java/com/toggl/timer/start/ui/StartTimeEntryDialogFragment.kt
```diff
@@ -0,0 +1,187 @@
...
+        savedInstanceState: Bundle?
+    ): View? {
+        return inflater.inflate(
+            R.layout.fragment_dialog_start_time_entry,
+            container,
+            false
+        )
+    }
+
+    @ExperimentalCoroutinesApi
```
heytherewill, line 79: Is this annotation needed? 🤔

timer/src/main/java/com/toggl/timer/start/ui/StartTimeEntryDialogFragment.kt
```diff
@@ -0,0 +1,187 @@
...
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View? {
+        return inflater.inflate(
+            R.layout.fragment_dialog_start_time_entry,
+            container,
+            false
+        )
+    }
```
heytherewill, line 77:
```suggestion
    ): View? = inflater.inflate(R.layout.fragment_dialog_start_time_entry, container, false)
```

timer/src/main/java/com/toggl/timer/start/ui/StartTimeEntryDialogFragment.kt
```diff
@@ -0,0 +1,187 @@
...
+                    showSoftInput(time_entry_description, InputMethodManager.SHOW_IMPLICIT)
+                }
+            }
+        }
+        time_entry_description.requestFocus()
+
+        time_entry_description.doOnTextChanged { text, _, _, _ ->
+            val action = StartTimeEntryAction.DescriptionEntered(text.toString())
+            store.dispatch(action)
+        }
```
heytherewill, line 113: A `with` block would serve the same purpose as the one above, creating initialization contexts for those reading the code

timer/src/main/java/com/toggl/timer/start/ui/StartTimeEntryDialogFragment.kt
```diff
@@ -0,0 +1,187 @@
...
+            store.state
+                .filter { it.editableTimeEntry == null }
+                .distinctUntilChanged()
+                .onEach { dismiss() }
+                .launchIn(this)
+
+            store.state
+                .map { it.editableTimeEntry?.description }
+                .first { !it.isNullOrEmpty() }
+                .also(time_entry_description::setText)
```
heytherewill, line 134: If `first` is a terminal operator like `collect`, won't this code run synchronously and set the text before running the code below? Wouldn't the correct behavior be limiting the flow to a single element (`.take(1)`) and `launchIn`? 🤔

timer/src/main/java/com/toggl/timer/start/ui/StartTimeEntryDialogFragment.kt
```diff
@@ -0,0 +1,187 @@
...
+
+            store.state
+                .map { it.timeEntries.runningTimeEntryOrNull() != null }
```
heytherewill, line 137: I don't think this is correct. If I'm editing a time entry and the sync kicks in bringing a new running time entry from the server, I don't want my edited time entry to change. This should check whether the `editableTimeEntry` has a duration or not

timer/src/main/res/layout/bottom_control_panel_layout.xml
```diff
@@ -0,0 +1,57 @@
...
+<com.google.android.material.card.MaterialCardView xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    android:layout_width="match_parent"
+    android:layout_height="wrap_content"
+    app:cardCornerRadius="0dp">
+
+    <LinearLayout
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:elevation="64dp"
```
heytherewill, line 11: No constants 🙅‍♂

timer/src/main/res/layout/bottom_control_panel_layout.xml
```diff
@@ -0,0 +1,57 @@
...
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_margin="@dimen/grid_1"
+            app:srcCompat="@drawable/ic_project" />
+
+        <ImageView
+            android:id="@+id/tag_action"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_margin="@dimen/grid_1"
```
heytherewill, line 25: We should probably consider changing the log to use grids as well!

timer/src/main/res/layout/fragment_dialog_start_time_entry.xml
```diff
@@ -96,62 +98,16 @@
...
-        app:srcCompat="@drawable/ic_play" />
-
     <TextView
         android:id="@+id/extended_options"
         android:layout_width="wrap_content"
         android:layout_height="400dp"
-        android:textSize="23dp"
         android:gravity="center"
         android:text="Extended Start/Edit features"
+        android:textSize="23dp"
```
heytherewill, line 107: Using dps in textSize + odd prime number + non extracted constant. Also, is this 23 taken from our designs?

timer/src/main/res/layout/running_time_entry_blank_layout.xml
```diff
@@ -3,14 +3,20 @@
     xmlns:app="http://schemas.android.com/apk/res-auto"
     xmlns:tools="http://schemas.android.com/tools"
     android:id="@+id/empty_running_time_entry_layout"
+    android:foreground="?selectableItemBackgroundBorderless"
     android:layout_width="match_parent"
-    android:layout_height="match_parent">
+    android:layout_height="match_parent"
+    android:clickable="true"
+    android:focusable="true"
+    >
```
heytherewill, line 11: Are these needed?

heytherewill, line 11:
```suggestion
    android:clickable="true"
    android:focusable="true">
```

common-android/src/main/java/com/toggl/common/sheet/OnSlideAction.kt
semanticer: Could be but the main argument for the interface is that you can have that `@FloatRange` annotation attached to `slideOffset` parameter - this is unfortunately impossible with the type alias currently :(

common-android/src/main/java/com/toggl/common/sheet/OnSlideAction.kt
semanticer: Since this and possible future implementations will be probably very simple one-liners I thought we could keep them in one file but I don't have a strong opinion here so decide what you prefer 🤷‍♂

timer/src/main/java/com/toggl/timer/start/ui/StartTimeEntryDialogFragment.kt
semanticer: Good catch! It looks like you're completely right.

timer/src/main/java/com/toggl/timer/start/ui/StartTimeEntryDialogFragment.kt
pdrj, line 137: How about `mapNotNull`?

timer/src/main/res/layout/bottom_control_panel_layout.xml
```diff
@@ -0,0 +1,57 @@
...
+            android:layout_weight="1" />
+
+        <ImageView
+            android:id="@+id/done_action"
+            android:layout_width="wrap_content"
+            android:layout_height="wrap_content"
+            android:layout_margin="@dimen/grid_1"
+            android:tint="@color/start_time_entry_button_background"
+            android:background="?selectableItemBackgroundBorderless"
+            app:srcCompat="@drawable/ic_play" />
```
pdrj, line 54: We should remember to change it to a tick later, if we don't have it ATM.

timer/src/main/java/com/toggl/timer/start/ui/StartTimeEntryDialogFragment.kt
```diff
@@ -0,0 +1,187 @@
...
+        (dialog as BottomSheetDialog).setOnShowListener(null)
+        super.onCancel(dialog)
+    }
+
+    private fun DialogInterface.attachBottomView(
+        bottomSheetDialog: BottomSheetDialog,
+        @LayoutRes layoutToAttach: Int
+    ) {
+        val coordinator =
+            (this as BottomSheetDialog).findViewById<CoordinatorLayout>(com.google.android.material.R.id.coordinator)
```
pdrj, line 167: A lot is going on here, attaching stuff dynamically etc. Also we rely on internal android ids (probably that is fine, but who knows). Is there no way to make the control panel live in the main layout and then just set its visibility? What breaks in such approach?

timer/src/main/java/com/toggl/timer/start/ui/StartTimeEntryDialogFragment.kt
pdrj: I'm wrong, you need a bool.

timer/src/main/java/com/toggl/timer/start/ui/StartTimeEntryDialogFragment.kt
semanticer: I entered a coroutine rabbit hole here... I think we are holding them wrong everywhere.

timer/src/main/java/com/toggl/timer/start/ui/StartTimeEntryDialogFragment.kt
semanticer: This is now useful again, after refactor that moved one of the coroutines to this method.

timer/src/main/java/com/toggl/timer/start/ui/StartTimeEntryDialogFragment.kt
semanticer: I've changed it to check if the editableTE is new or "in edit" mode via `editableTimeEntry.ids.isNullOrEmpty()`

timer/src/main/res/layout/bottom_control_panel_layout.xml
semanticer: There are no constraints needed since the parent is `MaterialCardView` = `FrameLayout`
But the elevation was completely useless here.

timer/src/main/res/layout/bottom_control_panel_layout.xml
semanticer: Yeah, it should be tick and the whole button should be circular. This is just preliminary like the rest of this layout.

timer/src/main/res/layout/fragment_dialog_start_time_entry.xml
semanticer: this whole TextView is completely preliminary just to demonstrate hiding/showing functionality based on the current bottom sheet state (half vs fully expanded).
This will probably be replaced with a wheel view and other cool stuff very soon.

common-android/src/main/java/com/toggl/common/sheet/OnSlideAction.kt
heytherewill: Oh. I see. Let's keep it this way then

timer/src/main/res/layout/running_time_entry_blank_layout.xml
semanticer: Ups, not anymore.

timer/src/main/res/layout/bottom_control_panel_layout.xml
heytherewill: I said constants, not constraints 😂

timer/src/main/res/layout/bottom_control_panel_layout.xml
```diff
@@ -0,0 +1,56 @@
+<?xml version="1.0" encoding="utf-8"?>
```
daividssilverio, line 1: This looks great when the keyboard is open, but it's hard to handle when it's closed, maybe we could add some padding to the icons, or some bottom margin, Idk 🤔

common-android/src/main/java/com/toggl/common/sheet/BottomSheetCallback.kt
semanticer: I forgot to use this to properly set visibility for `extended_options` - that is done now. However, the point is that we will probably need this very soon when we start to implement all these controls that should be present in a fully extended state only.

timer/src/main/res/layout/bottom_control_panel_layout.xml
semanticer: I agree but, since the UI is still more or less preliminary I wouldn't bother with these details right now.

daividssilverio:
I tried but couldn't pinpoint the cause of the leaks.

It seems to be coming from one of the listeners of the `StartTimeEntryDialogFragment`, but I don't see any hard references being hold... maybe it's something with the navigation? (Seems to be related to `onCancel/onDismiss` based on leak canary).

The other leak seems to be coming from the `also` usage on `onCreateDialog`, but that also doesn't make much sense.

There seem to be another leaks going on, but not related to your changes (it looks like our recyclerview from the main log is leaking...).

timer/src/main/java/com/toggl/timer/start/ui/StartTimeEntryDialogFragment.kt
semanticer: I couldn't find a better way to do this floating behaviour. The problem is that `BottomSheetDialogFragment` includes our layout into some scrollview automatically but we need to put this `bottom_control_panel_layout` outside of this scrollview so it can "float" above the rest of our content.

semanticer:
I believe I fixed all the leak except this one:
Screen Shot 2020-04-06 at 18 46 28

which seem to be out of our control and it's probably not a big deal

common-android/src/main/java/com/toggl/common/DeepLinkUrls.kt
```diff
@@ -1,13 +1,20 @@
...
+    val timeEntriesStartDialog = resources.getString(R.string.deep_link_start_time_entry_dialog).toUri()
+    val calendar = resources.getString(R.string.deep_link_calendar).toUri()
+    val reports = resources.getString(R.string.deep_link_reports).toUri()
 }
+
+val Context.deepLinks: DeepLinkUrls
+    get() = DeepLinkUrls(resources)
+
+val Fragment.deepLinks: DeepLinkUrls
+    get() = DeepLinkUrls(resources)
```
heytherewill, line 20: :sogood:

timer/src/main/java/com/toggl/timer/start/ui/StartTimeEntryDialogFragment.kt
```diff
@@ -147,6 +157,9 @@ class StartTimeEntryDialogFragment : BottomSheetDialogFragment() {
     override fun onDestroyView() {
         bottomSheetCallback.clear()
         store.dispatch(StartTimeEntryAction.DialogDismissed)
+        with(time_entry_description) {
+            removeTextChangedListener(descriptionChangeListener)
+        }
```
heytherewill, line 161: This `with` is kinda overkill since it only adds two extra lines

timer/src/main/java/com/toggl/timer/start/ui/StartTimeEntryDialogFragment.kt
```diff
@@ -0,0 +1,206 @@
...
+    private val descriptionChangeListener: TextWatcher = object : TextWatcher {
+        override fun afterTextChanged(text: Editable) {
+            val action = StartTimeEntryAction.DescriptionEntered(text.toString())
+            store.dispatch(action)
+        }
+
+        override fun beforeTextChanged(s: CharSequence, start: Int, count: Int, after: Int) {}
+
+        override fun onTextChanged(s: CharSequence, start: Int, before: Int, count: Int) {}
+    }
```
heytherewill, line 67: I'd rather see an implementation of our own `TextWatcher` in `common-android` that takes optional callbacks so we can have control over our listeners without needing this boilerplate in the fragment

timer/src/main/java/com/toggl/timer/start/ui/StartTimeEntryDialogFragment.kt
```diff
@@ -0,0 +1,206 @@
...
+            addBottomSheetCallback(bottomSheetCallback)
+            skipCollapsed = false
+            peekHeight = resources.getDimension(R.dimen.time_entry_edit_half_expanded_height).toInt()
+            state = BottomSheetBehavior.STATE_COLLAPSED
+        }
+
+        with(time_entry_description) {
+            setOnFocusChangeListener { _, _ ->
+                post {
+                    (activity?.getSystemService(Context.INPUT_METHOD_SERVICE) as InputMethodManager).apply {
```
heytherewill, line 124: Would be nice to have an extension like we do in Xamarin to avoid such casts:

```kotlin
fun <T> Activity.getSystemService(service: String) : T =
    getSystemService(service) as T
```

So this would become:
```suggestion
                    activity?.getSystemService<InputMethodManager>(Context.INPUT_METHOD_SERVICE).apply {
```

heytherewill: Also, why aren't you using `requireActivity`?

timer/src/main/java/com/toggl/timer/start/ui/StartTimeEntryDialogFragment.kt
daividssilverio: Thought the same thing.

daividssilverio:
@semanticer I don't think that there's too much to do about those leaks marked as "Library leak" by the leak canary.

But I still managed to get some leaks that seem related to the usage of `also` in the `onCreateDialog`, which are similar to the one you've reported and also have the same "Library leak" tag.

Maybe we don't have to do anything for now.

Some stuff I've tried:
- By inlining the following field, the leaks took longer to happen
```
private val bottomSheetBehavior: BottomSheetBehavior<FrameLayout> by lazy(LazyThreadSafetyMode.NONE) {
        (dialog as BottomSheetDialog).behavior
    }
```

- Inlining the usage of `also` in the `onCreateDialog`, removed the leak entry that mentioned the `also` anonymous lambda.
 
I've also took at look at this reply on StackOverflow linked on the Leak canary website: https://stackoverflow.com/questions/59503689/could-navigation-arch-component-create-a-false-positive-memory-leak/59504797#59504797

It could help to not use the `kotlinx.android.synthetic.*` for the dialogs, as the references to the views might not be actually getting nulled out for some reason.

app/src/main/res/navigation/tabs_nav_graph.xml
```diff
@@ -7,27 +7,27 @@
...
-        <deepLink app:uri="toggl://starttimeentrydialog" />
+        <deepLink app:uri="@string/deep_link_start_time_entry_dialog" />
     </dialog>
 
     <fragment
         android:id="@+id/reports"
         android:name="com.toggl.reports.ui.ReportsFragment"
         android:label="ReportsFragment">
-        <deepLink app:uri="toggl://reports" />
+        <deepLink app:uri="@string/reports" />
```
pdrj, line 24: This is wrong and causes a crash on navigating to reports tab ;)

timer/src/main/res/layout/bottom_control_panel_layout.xml
semanticer: 😂dyslexia.... funny that I fixed it anyway

timer/src/main/java/com/toggl/timer/start/ui/StartTimeEntryDialogFragment.kt
semanticer: I must have been in a very strange state of mind when I did this

semanticer: Or in some ©️ 🍝 frenzy

app/src/main/res/navigation/tabs_nav_graph.xml
semanticer: 🤦‍♂️

timer/src/main/java/com/toggl/timer/start/ui/StartTimeEntryDialogFragment.kt
semanticer: Since this is happening inside of a listener and a post I would rather be safe and use safe access rather than to risk the possibility of NPE in the future just because of some non-essential keyboard logic.

timer/src/main/java/com/toggl/timer/start/ui/StartTimeEntryDialogFragment.kt
semanticer: This is a sad story you probably don't want to hear. There is actually an optional TextWatcher and related extension method provided by ktx and I used it in previous interactions of this PR but then you cannot unsubscribe listener from `TextEdit` since there is no `clearAllListeners()` method available. Only way to remove a listener is `removeTextChangedListener(concreteInstance)` where you have to pass the instance that you want to remove 🤦‍♂ 
Removing this listener seems to solve one of our memory leaks. 

**EDIT: Ok, addTextChangedListener from ktx is actually returning text watcher instance so you can easily unsubscribe later. As always Jake knows what he's doing and I'm wrong.**

semanticer:
I've tried @pdrj's approach and I also tried to switch to findViewById and view binding and nothing helped with the last leak we are seeing. I still think it's stated pretty clearly here https://github.com/square/leakcanary/issues/1668 that this is the library issue and shouldn't be a huge deal. Also here is the quote from StackOverflow answer @daividssilverio [mentioned](https://stackoverflow.com/questions/59503689/could-navigation-arch-component-create-a-false-positive-memory-leak/59504797#59504797) 

> As per the Fragments: Past, Present, and Future talk, one of the future changes coming to Fragments is an opt in option to destroy Fragments on the back stack, rather than having two separate lifecycles. This isn't available as of yet.

So this could be fixable by the time we release this app.

PR #128, opened 2020-04-06
Add CommitDeletion action

### LINKED ISSUE #101: Introduce the CommitDeletion action
## Description
As discussed [here](https://github.com/toggl/mobile-discussions/issues/58) we need an action that "flushes" the pending changes to the repo. We need such action because we want to check that we are deleting the correct set of time entries.

*NOTE: We are adding this action before it's actually emitted by any effect. This is needed to keep issues as small as possible and doesn't affect the testability of how the reducer will interact with the action, so this is fine*

## Dependencies
Depends on #100

## Definition of done
- [ ] The `CommitDeletion` action is added
- [ ] The action is handled by the `TimeEntriesLogReducer` as shown in the linked mobile-discussions post
  - [ ] A check is performed to ensure that the entries about to be committed are still the ones in the list of entries pending deletion
  - [ ] If they are, empty the list of entries pending deletion and proceed to delete
  - [ ] If they aren't, do nothing
- [ ] Tests are added to ensure the reducer behavior is correct

pdrj:
### Summary
This adds CommitDeletion action which actually creates the effects deleting time entries pending deletion.
### Technical considerations
When creating delete effects, I get the TEs to delete from state, based on the IDs stored in `entriesPendingDeletion`. I think this is fine, but if you can think of a scenario where this can break, please enlighten me 😃   

This was the most straightforward.

### Relationships

Closes #101 

---
### Pull Request checklist

### Review template

- [ ] The code considers i18n

### DIFF
--- a/timer/src/main/java/com/toggl/timer/extensions/SetExtensions.kt
+++ b/timer/src/main/java/com/toggl/timer/extensions/SetExtensions.kt
@@ -1,0 +1,3 @@
+package com.toggl.timer.extensions
+
+fun <T> Set<T>.containsExactly(elements: Collection<T>) = this.containsAll(elements) && this.size == elements.size
--- a/timer/src/main/java/com/toggl/timer/extensions/SetExtensions.kt
+++ b/timer/src/main/java/com/toggl/timer/extensions/SetExtensions.kt
@@ -1,3 +1,3 @@
 package com.toggl.timer.extensions
 
-fun <T> Set<T>.containsExactly(elements: Collection<T>) = this.containsAll(elements) && this.size == elements.size
+fun <T> Set<T>.containsExactly(elements: Collection<T>) = this.size == elements.size && this.containsAll(elements)
--- a/timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogAction.kt
+++ b/timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogAction.kt
@@ -16,6 +16,7 @@ sealed class TimeEntriesLogAction {
     data class TimeEntryStarted(val startedTimeEntry: TimeEntry, val stoppedTimeEntry: TimeEntry?) :
         TimeEntriesLogAction()
     data class ToggleTimeEntryGroupTapped(val groupId: Long) : TimeEntriesLogAction()
+    data class CommitDeletion(val ids: List<Long>) : TimeEntriesLogAction()
 
     companion object {
         fun fromTimerAction(timerAction: TimerAction): TimeEntriesLogAction? =
@@ -39,4 +40,5 @@ fun TimeEntriesLogAction.formatForDebug() =
         is TimeEntriesLogAction.TimeEntryStarted -> "Time entry started $startedTimeEntry"
         is TimeEntriesLogAction.TimeEntryDeleted -> "Deleted time entry with id ${deletedTimeEntry.id}"
         is TimeEntriesLogAction.ToggleTimeEntryGroupTapped -> "Time entry group with id $groupId toggled"
+        is TimeEntriesLogAction.CommitDeletion -> "Commit deletion of $ids"
     }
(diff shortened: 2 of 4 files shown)


timer/src/main/java/com/toggl/timer/extensions/SetExtensions.kt
```diff
@@ -0,0 +1,3 @@
+package com.toggl.timer.extensions
+
+fun <T> Set<T>.containsExactly(elements: Collection<T>) = this.containsAll(elements) && this.size == elements.size
```
heytherewill, line 3: Checking size first is faster

timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogAction.kt
```diff
@@ -39,4 +40,5 @@ fun TimeEntriesLogAction.formatForDebug() =
         is TimeEntriesLogAction.TimeEntryStarted -> "Time entry started $startedTimeEntry"
         is TimeEntriesLogAction.TimeEntryDeleted -> "Deleted time entry with id ${deletedTimeEntry.id}"
         is TimeEntriesLogAction.ToggleTimeEntryGroupTapped -> "Time entry group with id $groupId toggled"
+        is TimeEntriesLogAction.CommitDeletion -> "Commit deletion of $ids"
```
heytherewill, line 43: Use the past tense here
```suggestion
        is TimeEntriesLogAction.CommitDeletion -> "Commited deletion of time entries with ids $ids"
```

timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogReducer.kt
```diff
@@ -102,6 +103,19 @@ class TimeEntriesLogReducer @Inject constructor(private val repository: TimeEntr
...
+                    val timeEntryIdsToDelete =
+                        if (state.value.entriesPendingDeletion.containsExactly(action.ids)) action.ids
+                        else listOf()
+
+                    if (timeEntryIdsToDelete.none()) {
+                        noEffect()
+                    } else {
+                        val timeEntriesToDelete = timeEntryIdsToDelete.mapNotNull { state.value.timeEntries[it] }
+                        state.value = state.value.copy(entriesPendingDeletion = setOf())
+                        timeEntriesToDelete.map { delete(it, repository) }
```
daividssilverio, line 116: 🤔 Can't we have a batch delete at this point?

timer/src/test/java/com/toggl/timer/log/domain/CommitDeletionActionTests.kt
```diff
@@ -0,0 +1,95 @@
...
+                    settableValue,
+                    TimeEntriesLogAction.CommitDeletion(listOf())
+                )
+
+                state shouldBe initialState
+                effect shouldBe noEffect()
+            }
+
+            "the ids pending deletion don't match those in state" {
+                val initialState = createInitialState(entriesPendingDeletion = setOf(1, 2, 3))
```
daividssilverio, line 37: it would be nice to include a semi-matching list of time entries in this test to enforce the idea that this would only work when you try to commit the deletion and the ids are the same.
Ex: what would happen if the state was `entriesPendingDeletion = [1, 2]`, but the action received is `[1, 2, 3]`?

timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogReducer.kt
pdrj: I don't think so 🤔 https://github.com/toggl/android/pull/97 @heytherewill would you please confirm that we are no longer supposed to be doing batch delete?

timer/src/test/java/com/toggl/timer/log/domain/CommitDeletionActionTests.kt
pdrj: In your example, this won't delete anything. I'll add a test like this. It only deletes when the ID sets match exactly.

timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogReducer.kt
heytherewill: This gets deleted in batch because of the way the store works; all of these get applied sequentially and only then emitted to the state

daividssilverio: What I wanted was an actual database batch delete, using a single transaction.
But we can do that later as well (if ever).

timer/src/test/java/com/toggl/timer/log/domain/CommitDeletionActionTests.kt
pdrj: https://toggl.slack.com/archives/CD40QCV25/p1586265976193400

timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogReducer.kt
pdrj: https://toggl.slack.com/archives/CD40QCV25/p1586265976193400

PR #129, opened 2020-04-06
Make it crash

### LINKED ISSUE #108: Force app to crash when an error is thrown
## Description
At the moment our app is silently swallowing the exception and then misbehaving. We don't want that. I suspect this is because of the call to [`GlobalScope.launch`](https://github.com/toggl/android/blob/master/architecture/src/main/java/com/toggl/architecture/core/Store.kt#L65) in our store. 

## Definition of done
- [ ] When an exception is thrown in a reducer, the app crashes

daividssilverio:
### Summary
This PR basically makes our `TogglApplication` be the provider of the coroutine scope used by our centralized `Store`, instead of using the `GlobalScope`.

### Technical considerations
Note that the store is processing the app actions in the main thread.
I've tried making it work on the `IO/Default` threads out of curiosity but they just wouldn't crash for some reason (still trying to figure out why...).

Using the `GlobalScope` is not recommended as we don't have control over the jobs being created by it and we are not handling their exceptions directly.
By creating our own scope we have more control of when/where our coroutines are running (at least the ones created off the store scope). Please note that whenever we use `lifecycleScope`, `viewModelScope`, they will use those scopes.

### Relationships
Closes #108 

---
### Pull Request checklist

### Pointers
At least run the app, take a look at the logging reducer, compare with it how it was before to make sure you are not surprised by anything.

One thing that I've noticed is that effects now are running sequentially, check the loading effects for example, compare it with how it was before the changes in this PR.

Try crashing the app from different reducers or parts of the app you think weren't crashing before.

### Review template

## Merge permissions
🦑 it after at least 2 @toggl/androids have reviewed.


heytherewill:
> Note that the store is processing the app actions in the main thread
This is fine. Any IO/CPU bound work needs to marshall, and using the main thread for dispatching ensures order :)

app/src/main/java/com/toggl/TogglApplication.kt
```diff
@@ -1,24 +1,43 @@
...
     @Inject
     lateinit var appInitializers: AppInitializers
 
+    @Inject
+    lateinit var dispatchersProviders: DispatcherProvider
+
+    lateinit var applicationJob: Job
+
+    override val coroutineContext: CoroutineContext
+        get() = dispatchersProviders.main + applicationJob
```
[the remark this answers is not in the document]
daividssilverio: Indeed, I don't see a reason why not.

app/src/main/java/com/toggl/TogglApplication.kt
```diff
@@ -28,4 +48,12 @@ class TogglApplication : Application(), OnboardingComponentProvider, TimerCompon
 
     override fun provideTimerComponent(): TimerComponent =
         appComponent.timerComponent().create()
+
+    override fun getStoreScope(): CoroutineScope =
+        this
+
+    override fun onTerminate() {
```
semanticer, line 55: From the `onTerminate` docs:
```
This method is for use in emulated process environments.  It will never be called on a production Android device, where processes are removed by simply killing them; no user code (including this callback) is executed when doing so.
```
I just want to check you are aware of this and that you are intentionally handling test environments 👆

app/src/test/java/com/toggl/common/CoroutineFreeSpec.kt
```diff
@@ -0,0 +1,31 @@
...
+import io.kotlintest.specs.AbstractFreeSpec
+import io.kotlintest.specs.FreeSpec
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.test.TestCoroutineDispatcher
+import kotlinx.coroutines.test.resetMain
+import kotlinx.coroutines.test.setMain
+
+@ExperimentalCoroutinesApi
+open class CoroutineFreeSpec(
```
heytherewill, line 15: `FreeCoroutineSpec` is a much better name. This sounds like a Spec without coroutines 😂

app/src/test/java/com/toggl/common/CoroutineFreeSpec.kt
```diff
@@ -0,0 +1,31 @@
...
+    var dispatcher: TestCoroutineDispatcher = TestCoroutineDispatcher(),
+    body: AbstractFreeSpec.() -> Unit = {}
+) : FreeSpec(body) {
+    var dispatcherProvider: DispatcherProvider = DispatcherProvider(dispatcher, dispatcher, Dispatchers.Main)
```
heytherewill, line 19: Why is main still used instead of a test dispatcher? 🤔

app/src/main/java/com/toggl/TogglApplication.kt
daividssilverio: Actually, I was not.
Do you have a suggestion for a better place to kill the job? Or maybe we shouldn't be doing this 🤔

app/src/test/java/com/toggl/common/CoroutineFreeSpec.kt
daividssilverio: The `Main` is replaced by the `TestCoroutineDispatcher` in the setup done in this class.

app/src/test/java/com/toggl/common/CoroutineFreeSpec.kt
daividssilverio: Fair enough.

app/src/main/java/com/toggl/TogglApplication.kt
semanticer: > no user code (including this callback) is executed when doing so.

This part is quite worrisome

However, we could maybe use [ProcessLifecycleOwner](https://developer.android.com/reference/androidx/lifecycle/ProcessLifecycleOwner) ?
Here is some guy implementing analytics with it for reference: https://proandroiddev.com/android-processlifecycleowner-by-example-2f965061b9da

daividssilverio: 🤔 Let's give it a try. Not exactly like the example though. That would make the store scope get cleared when going to background and that's not what we want.

daividssilverio: Yeah... I couldn't find a reliable way of detecting if the application has been killed/terminated by the OS.

Maybe we shouldn't do anything specific then 🤔 and that makes sense as well...
When the application is killed, everything is killed alongside with it anyway.

Removing the `SupervisorJob` completely doesn't change the behaviour we want to achieve.

daividssilverio: I don't want to add counters for non destroyed activities/services... that sounds like a bad idea.
What we could achieve.

semanticer: Oh please no counters 🙏
And yes you are probably right, we don't need to worry about this that much.

PR #134, opened 2020-04-07
Add UndoButtonPressed action

### LINKED ISSUE #103: Add UndoButtonPressed action
## Description
We need an action that allows the users to undo their deletion mishaps 

## Dependencies
Depends on the property added #100

## Definition of done
- [ ] The action is added
- [ ] The reducers clears the `entriesPendingDeletion` property when the action is received
- [ ] Tests are added

pdrj:
### Summary
This adds UndoButtonPressed action

This is pretty straightforward.

### Relationships

Closes #103 

---
### Pull Request checklist

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone.

### DIFF
--- a/timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogReducer.kt
+++ b/timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogReducer.kt
@@ -102,6 +102,11 @@ class TimeEntriesLogReducer @Inject constructor(private val repository: TimeEntr
                     state.value = state.value.copy(expandedGroupIds = newUngroupedTimeEntries)
                     noEffect()
                 }
+                is TimeEntriesLogAction.UndoButtonPressed -> {
+                    if (state.value.entriesPendingDeletion.isNotEmpty())
+                        state.value = state.value.copy(entriesPendingDeletion = setOf())
+                    noEffect()
+                }
             }
 
     private fun startTimeEntry(timeEntry: EditableTimeEntry, repository: TimeEntryRepository) =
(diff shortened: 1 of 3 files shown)


timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogReducer.kt
```diff
@@ -102,6 +102,11 @@ class TimeEntriesLogReducer @Inject constructor(private val repository: TimeEntr
                     state.value = state.value.copy(expandedGroupIds = newUngroupedTimeEntries)
                     noEffect()
                 }
+                is TimeEntriesLogAction.UndoButtonPressed -> {
+                    if (state.value.entriesPendingDeletion.isNotEmpty())
+                        state.value = state.value.copy(entriesPendingDeletion = setOf())
```
heytherewill, line 107: Why do we even need this check?
Just set it to `emptySet` and call it a day :)
```suggestion
                    state.value = state.value.copy(entriesPendingDeletion = emptySet())
```

pdrj: I wasn't sure how it worked here, coming from react world. I just wanted to make sure we don't needlessly update the state with updates that don't do anything, since that might trigger some subscribers to run computations and stuff. 

For instance (and yeah, I should maybe just test it) - if I reset it to `emptySet`, when it already is empty, will the time entries selector rerun? If it does, then I'd say we leave the check in place.

Issue #102, opened 2020-03-27
Introduce the WaitForUndoEffect


heytherewill:
## Description
We need to add the effect that will actually delete time entries that are pending because of the undo.

## Dependencies
This PR needs the `CommitDeletion` action, introduced on #101 

## Definition of done
- [ ] The effect that asynchronously waits for [5 seconds](https://github.com/toggl/mobileapp/blob/develop/Toggl.Core/Helper/Constants.cs#L26) and then emit a `CommitDeletion` action is added
- [ ] Make actions that delete (swipes) emit the following:
  - [ ] If there are no entries in the `entriesPendingDeletion` property, simply emit the `WaitForUndo` effect rather than emitting delete effects immediatelly
  - [ ] If there are entries in the  `entriesPendingDeletion` property, emit effects to delete each of those entries and also the `WaitForUndo` effect.
- [ ] Tests that cover all this new behavior are added

PR #135, opened 2020-04-07
Show undo deletion snackbar

### LINKED ISSUE #133: Show a snackbar to allow undoing time entry deletions
## Description
We need to show a snackbar to allow the user to undo deletions

## Dependencies
Requires #103

## Definition of done
- [ ] We display a Snackbar for 5 seconds whenever the list of pending deletions changes
- [ ] The Snackbar should show the same string from the Xamarin app + an undo button
- [ ] The undo button emits an `UndoButtonPressed` action
- [ ] Pressing the button makes the deleted entries reappear on the log immediatelly
- [ ] Pressing the button cancels the deletion of all entries pending deletion

daividssilverio:
### Summary
This PR is part of big picture of the undo deletion feature.

It does three things:
- Shows the snackbar when there are ids pending for deletion
- Dispatches `TimeEntriesLogAction.CommitDeletion` when the snackbar timeout happens
- Dispatches `TimeEntriesLogAction.UndoButtonPressed` when the user clicks on `Undo` in the snackbar

### Technical considerations
Nothing special here.

I've used the snackbar built in timeout to trigger `CommitDeletion`, but that could've been done using the delay that will be introduced by #102 .

If we decide to go that way, we could always map the state `entriesPendingDeletion.size` to tell if the snackbar is visible or not, leaving it visible indefinitely until we actually dismiss it when `entriesPendingDeletion.size == 0`.

### Relationships
Closes #133

---
### Pull Request checklist

### Pointers
You'll need manually add ids to the `entriesPendingDeletion` in the `TimeEntriesLogReducer` during swipes to see the snackbar (or simply set some ids).

The changes bellow made in the `TimeEntriesLogReducer` are enough to see/test the snackbar.
```
is TimeEntriesLogAction.TimeEntrySwiped -> {
                    val swipedEntry = state.value.timeEntries[action.id]
                        ?: throw IllegalStateException()

                    when (action.direction) {
                        SwipeDirection.Left -> {
                            state.value = state.value.copy(
                                entriesPendingDeletion = setOf(swipedEntry.id)
                            )
                            noEffect()
                        }
                        SwipeDirection.Right ->
                            startTimeEntry(EditableTimeEntry.fromSingle(swipedEntry), repository)
                    }
                }

                is TimeEntriesLogAction.TimeEntryGroupSwiped -> {

                    when (action.direction) {
                        SwipeDirection.Left -> {
                            state.value = state.value.copy(
                                entriesPendingDeletion = action.ids.toSet()
                            )
                            noEffect()
                        }
                        SwipeDirection.Right -> {
                            val timeEntryToStart = state.value.timeEntries[action.ids.first()]
                                ?.let { EditableTimeEntry.fromGroup(action.ids, it) }
                                ?: throw IllegalStateException()
                            startTimeEntry(timeEntryToStart, repository)
                        }
                    }
                }
```

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it after considering "Why is it done like this and not some other way?"


common-android/src/main/res/values-ja/strings.xml
```diff
@@ -12,4 +12,7 @@
     <string name="yesterday">昨日</string>
     <string name="continue_this">続ける</string>
     <string name="delete">削除</string>
+    <string name="entryDeleted">記録は削除されました</string>
+    <string name="multipleEntriesDeleted">%d個の記録が削除されました</string>
```
heytherewill, line 16: Please use the [plurals](https://developer.android.com/guide/topics/resources/string-resource#Plurals) API to support languages that have exceptions for a specific amount of items

common-android/src/main/res/values-pt/strings.xml
```diff
@@ -12,4 +12,7 @@
     <string name="yesterday">Ontem</string>
     <string name="continue_this">Continuar</string>
     <string name="delete">Excluir</string>
+    <string name="entryDeleted">Le temps d´entrée est supprimé</string>
+    <string name="multipleEntriesDeleted">%d entrées de temps supprimées</string>
+    <string name="undo">ANNULER</string>
```
heytherewill, line 17: Please don't add strings in all caps, since this is a display detail. We can capitalize the string when showing it to avoid having repeated resources

common-android/src/main/res/values-pt/strings.xml
```diff
@@ -12,4 +12,7 @@
     <string name="yesterday">Ontem</string>
     <string name="continue_this">Continuar</string>
     <string name="delete">Excluir</string>
+    <string name="entryDeleted">Le temps d´entrée est supprimé</string>
```
heytherewill, line 15:
```suggestion
    <string name="entryDeleted">L'entrée de temps est supprimé</string>
```

timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogFragment.kt
```diff
@@ -23,16 +22,17 @@ import com.toggl.timer.log.domain.TimeEntriesLogState
...
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.flow.distinctUntilChanged
 import kotlinx.coroutines.flow.launchIn
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.launch
+import javax.inject.Inject
 
 class TimeEntriesLogFragment : Fragment(R.layout.fragment_time_entries_log) {
+    private val undoTimeOutInMillis = 5000
```
heytherewill, line 35: This should be a constant in `common` so it can be used in other places as well

timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogFragment.kt
```diff
@@ -96,9 +99,45 @@ class TimeEntriesLogFragment : Fragment(R.layout.fragment_time_entries_log) {
...
+        if (idsToDelete.isEmpty())
+            return
+
+        val dismissCallback = object : Snackbar.Callback() {
+            override fun onDismissed(transientBottomBar: Snackbar?, event: Int) {
+                if (event == DISMISS_EVENT_TIMEOUT) {
+                    store.dispatch(TimeEntriesLogAction.CommitDeletion(idsToDelete.toList()))
+                }
+            }
+        }
```
heytherewill, line 131: Yeah, we are better off just relying on the effect, no need to dispatch CommitDeletion here

common-android/src/main/res/values-ja/strings.xml
daividssilverio: I've tried but the IDE complained about using the operation with the `count == 1` without the `%d` parameter. I didn't run to confirm, let's try that again.

common-android/src/main/res/values-pt/strings.xml
daividssilverio: I'm pretty sure capitalizing strings is wasteful from when I watched talks about fonts and text views. I think it's fine to make the changes, but let's keep that in mind.

common-android/src/main/res/values-pt/strings.xml
daividssilverio: That's how this was translated in the current app. Gotta fix it there too.

timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogFragment.kt
daividssilverio: Alright o/

common-android/src/main/res/values-pt/strings.xml
heytherewill: In plain text view we can set `android:textAllCaps` and in this case a simple `toUpperCase` won't be the bottleneck of performance, so it's okay

common-android/src/main/res/values-pt/strings.xml
daividssilverio: Oops, this is file for the Portuguese translations. 🙈

common-android/src/main/res/values-ja/strings.xml
```diff
@@ -12,11 +12,9 @@
     <string name="yesterday">昨日</string>
     <string name="continue_this">続ける</string>
     <string name="delete">削除</string>
-    <string name="entryDeleted">記録は削除されました</string>
-    <string name="multipleEntriesDeleted">%d個の記録が削除されました</string>
     <string name="undo">戻す</string>
-    <plurals name="entriesDeleted">
-        <item quantity="one">@string/entryDeleted</item>
-        <item quantity="other">@string/multipleEntriesDeleted</item>
+    <plurals name="entriesDeleted" tools:ignore="UnusedQuantity">
```
heytherewill, line 16: Should it be unused or implied? 🤔

common-android/src/main/res/values/strings.xml
```diff
@@ -11,11 +11,9 @@
     <string name="yesterday">Yesterday</string>
     <string name="continue_this">Continue</string>
     <string name="delete">Delete</string>
-    <string name="entryDeleted">Time entry was deleted</string>
-    <string name="multipleEntriesDeleted">%d time entries deleted</string>
     <string name="undo">Undo</string>
     <plurals name="entriesDeleted">
-        <item quantity="one">@string/entryDeleted</item>
-        <item quantity="other">@string/multipleEntriesDeleted</item>
+        <item quantity="one">Time entry was deleted</item>
```
heytherewill, line 16: Why is the annotation not needed here?

common-android/src/main/res/values-ja/strings.xml
daividssilverio: Actually, this is useless. I'll be removing it.

common-android/src/main/res/values/strings.xml
daividssilverio: In Japanese, the explanation was clear (the singular simply wasn't being used).
But I get a warning in Portuguese and French.
In English it doesn't say anything for some reason.

The docs say this about "one":
```
When the language requires special treatment of numbers like one (as with the number 1 in English and most other languages; in Russian, any number ending in 1 but not ending in 11 is in this class).
```

daividssilverio:
I thought that was going to be fixed somewhere else 🤔

PR #138, opened 2020-04-08
✨ Add WaitForUndoEffect and fix labels when swiping

### LINKED ISSUE #99: Swipes over TEs have wrong labels.
## Description
When swiping from left to right on a TE, the green overlay appears and the TE gets continued. The label on the green overlay says 'Delete' though. Same story for swiping right.
Screenshot 2020-03-27 at 15 27 39

## Definition of done
Label when swiping left says 'Continue', when swiping right 'Delete'.

### LINKED ISSUE #102: Introduce the WaitForUndoEffect

pdrj:
### Summary
This adds and uses WaitForUndoEffect. This also fixes labels when swiping TEs.

### Technical considerations
The effect is pretty straightforward. In tests, rather than having them wait for 5s, we could use a context to advance time (I think Daivid introduced it in one of his recent PRs).

I'm open to any suggestions!

### Relationships

Closes #102 
Closes #99 

---
### Pull Request checklist

### Review template

- [ ] The code considers i18n


timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogReducer.kt
```diff
@@ -55,7 +54,16 @@ class TimeEntriesLogReducer @Inject constructor(private val repository: TimeEntr
...
+                        SwipeDirection.Left -> {
+                            val entriesToCommitDeletion = state.value.entriesPendingDeletion
+                            val entriesToDelete = listOf(swipedEntry.id)
+                            state.value = state.value.copy(entriesPendingDeletion = entriesToDelete.toSet())
+
+                            val deleteEffects = prepareDeleteEffects(state, entriesToCommitDeletion)
+
+                            deleteEffects.add(WaitForUndoEffect(entriesToDelete))
+                            deleteEffects
+                        }
```
heytherewill, line 66: Couldn't all that be extracted into a method to prevent duplication of code?

timer/src/main/java/com/toggl/timer/log/domain/WaitForUndoEffect.kt
```diff
@@ -0,0 +1,16 @@
+package com.toggl.timer.log.domain
+
+import com.toggl.architecture.core.Effect
+import kotlinx.coroutines.delay
+
+class WaitForUndoEffect(
+    private val timeEntryIdsToDelete: List<Long>
+) : Effect<TimeEntriesLogAction> {
+    private val timeEntryDeletionDelayMs = 5000L
```
heytherewill, line 9: This needs to be moved to `:common` because it needs to be reused by the UI

timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogFragment.kt
```diff
@@ -123,14 +123,14 @@ class TimeEntriesLogFragment : Fragment(R.layout.fragment_time_entries_log) {
...
-            text = getString(R.string.continue_this),
+            text = getString(R.string.delete),
             backgroundColor = leftSwipeColor,
             textColor = actionTextColor,
             fontSize = actionFontSize,
             textPadding = actionPadding
         )
         val swipeRightParams = SwipeActionParams(
-            text = getString(R.string.delete),
+            text = getString(R.string.continue_this),
```
heytherewill, line 133: why `continue_this`? 🤔

timer/src/test/java/com/toggl/timer/log/domain/WaitForUndoEffectTests.kt
```diff
@@ -0,0 +1,23 @@
...
+import kotlin.system.measureTimeMillis
+
+class WaitForUndoEffectTests : FreeSpec({
+
+    "The WaitForUndoEffect" - {
+        "when executed, waits and then emits CommitDeletion" - {
+            val effect = WaitForUndoEffect(listOf(1, 8, 1337))
+
+            var action: TimeEntriesLogAction? = null
+            val elapsed = measureTimeMillis {
```
heytherewill, line 15: Can't you mock the time using a test scope so we don't actually need to wait for 5 seconds?

timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogFragment.kt
pdrj: I just swapped what was already in there, haven't added any resources ;)

timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogReducer.kt
pdrj: It could 👍

timer/src/main/java/com/toggl/timer/log/domain/WaitForUndoEffect.kt
pdrj: How so? Deletions from the edit view will have a 'popup' confirmation. Did you have something else in mind?

timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogFragment.kt
heytherewill: Let's maybe rename this resource?

timer/src/main/java/com/toggl/timer/log/domain/WaitForUndoEffect.kt
heytherewill: This == the 5000L constant. It will be reused by the UI to display the undo toast

pdrj: Haha 😂 demonstratives strike again

timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogFragment.kt
pdrj: Sure

PR #141, opened 2020-04-08
Implement new dependency scheme

### LINKED ISSUE #107: Implement our new dependency scheme
## Description
As discussed and agreed [here](https://github.com/toggl/mobile-discussions/issues/60), we want our new scheme to make sense!

## Definition of done
- [ ] We use `api` where we agreed to use `api` instead of `implementation` to make dependencies less convoluted
- [ ] The `:architecture` module is changed into a pure Kotlin module
- [ ] We add the intermediate `:common-feature` module is added and used as the only dependency for most modules

daividssilverio:
### Summary
This PR implements what was discussed on #107 

### Technical considerations
I've followed the dependency graph example on #107, adapting where necessary.
Since the `:database-android` project wasn't being added anywhere, I've just added it to the `app` project.

I found it weird at first, but for some reason only onboarding depends on `:api`, and Dagger uses it to build the dependency graph. There was a bug that was preventing the `DaggerAppComponent` from being generated and it was all because I removed the `:api` as direct dependency from the `:app` and `:onboarding` was importing  `:api`  using `implementation` instead of `api`. Unfortunately, the errors weren't helpful...

Feel free to suggest more changes. I'm guessing we can cut off more stuff.

### Relationships
Closes #107 

---
### Pull Request checklist

### Pointers
It would be helpful if you could compare compile/build times using master and this branch.
I didn't perceive much of a difference.

### Review template

## Merge permissions
🦑 Ideally, everybody from @toggl/androids should review this before merging


daividssilverio:
I think having `:database-android` only imported in the `:app` totally fine, it even helps hiding stuff from other modules.

We could give it a try having `:environment` exposing `:api`, if it works it makes things simpler when `:api` is used in other places 🤔

PR #143, opened 2020-04-09
Make DoneButtonTapped work as intended and make edits work too

### LINKED ISSUE #60: Add DoneButtonTapped action
## Description
As per what was discussed with @pdrj, we won't be saving the changes in the start time entry every time the user does something. Instead, we will keep the changes somewhere and commit them when the user emits a [`DoneButtonTapped`](https://www.notion.so/toggl/Start-time-entry-5783684cca064055b4512819087375bb?p=75aa87156f6e46859a5ce6a5711fea89) action. 

The action added in this issue should use gather all the intermediate changes and commit them to the database. 

## Definition of done
- [ ] The action is added and handled by the reducer
- [ ] Tests cover that the action changes the state correctly

pdrj:
### Summary
This makes DoneButtonTapped work as intended. As a side effect, edits work as well.

### Technical considerations
It's pretty straightforward I think.

I could have the `shouldStart` flag in state, but I like the current approach better. It's simpler.

### Relationships

Closes #60 

---
### Pull Request checklist

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
```diff
@@ -63,9 +64,23 @@ class StartEditReducer @Inject constructor(
                 noEffect()
             }
             StartEditAction.DoneButtonTapped -> {
-                val timeEntry = state.value.editableTimeEntry
+                val editableTimeEntry = state.value.editableTimeEntry!!
                 state.value = state.value.copy(editableTimeEntry = null)
-                startTimeEntry(timeEntry!!, repository)
+                if (editableTimeEntry.shouldStart) {
```
heytherewill, line 69: You don't need this `shouldStart` property. If the editable time entry doesn't have an id, it means that it should be started :)

pdrj: Right... 😅

Issue #144, opened 2020-04-09
Setup database testing infrastructure


semanticer:
## Description
We need to set up our testing infrastructure to support Dao and migration testing. We don't want to run these tests on the device since it's slow so we should consider using Roboelectric.
## Dependencies
If we want to write our test with Kotest + Roboeletric we need to migrate to Kotest 4 first to be able to use it's [Roboelectric extension](https://github.com/kotest/kotest/tree/master/kotest-extensions/kotest-extensions-robolectric)

Depends on #151

## Definition of done
- [ ] Setup working combo of Kotest + Robolectric
- [ ] Setup inMemory room database
- [ ] Create Dagger TestComponent that will provide Room, Daos etc
- [ ] Create a test for at least one Dao method
- [ ] Setup migration testing infrastructure

PR #147, opened 2020-04-09
Extract some of the Start/Stop functionality from Repository to Dao

### LINKED ISSUE #109: Make starting a time entry happen in a single transaction
## Description
For quick prototyping purposes, we made starting/stopping something that happens in the repository layer. It should instead be made in a single [`@Transaction`](https://developer.android.com/reference/android/arch/persistence/room/Transaction) in the DAO layer.

## Definition of done
- [ ] The TimeEntryDAO has a stop method that queries the DB and stops a running entry, if any
- [ ] The TimeEntryDAO has a start method that, in a single transaction, calls the method to stop the running TE and then create the new time entry.

semanticer:
### Summary
This PR moves some logic from the Repository's `stopRunningTimeEntry` and `startTimeEntry` to `TimeEntryDao`. Thanks to this, we are now able to put both methods into their own Transactions.

### Technical considerations
**Tests**
I had to change Repository tests because the repository is doing much less work now, basically just calling `Dao` methods. The problem is that we now really need to test our `Dao` and we should do so with a proper in-memory database. We either need to run our tests on an Android device or use Roboelectric. Roboelectric seems to be a much better option for many reasons, but there is probably no way we can make it work with Kotest 3 unfortunately. 
I created a separate Issue for database testing #144

**Transactions**
Here is the full discussion about them on slack https://toggl.slack.com/archives/CD40QCV25/p1586348824212300
Basically I'm still worried about some inconsistencies especially in our state and transactions can make this even worse (maybe?)

Feel free to suggest anything.

### Relationships

Closes #109
And if we want Dao tests we are currently blocked by #144

---
### Pull Request checklist

- [ ] **The code in this PR is tested thoroughly enough that changing the behavior to an incorrect one breaks tests** (we can't test Daos properly at the moment)

### Pointers
Make sure that the new Repository tests make sense and your opinion on the issue of transactions will be also valuable.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


repository/src/main/java/com/toggl/repository/Repository.kt
```diff
@@ -67,31 +66,18 @@ class Repository(
...
-        )
-        return StartTimeEntryResult(
-            timeEntryDao.getOneTimeEntry(id).let(DatabaseTimeEntry::toModel),
-            stoppedTimeEntry
-        )
+        return timeEntryDao.startTimeEntry(workspaceId, description, timeService.now())
+            .let { (started, stopped) ->
+                StartTimeEntryResult(
+                    started.let(DatabaseTimeEntry::toModel),
+                    stopped.firstOrNull()?.let(DatabaseTimeEntry::toModel)
```
heytherewill, line 73: This particular line might lead to complications 🤔 
Each new line of code makes me wanna have DB subscriptions harder. We really should re-evalutate this

PR #154, opened 2020-04-10
Migrate to Kotest4

### LINKED ISSUE #151: Investigate and upgrade our tests to use Kotest 4.0
## Description
Currently, our tests are slow to write/change because. IDE auto completion is very sluggish and not dev friendly. 
The idea is that maybe kotest 4.0 is faster.

## Dependencies
None.

## Definition of done
- [ ] Kotest is migrated to 4.0
- [ ] Tests are migrated to 4.0 (if needed)
- [ ] Auto completion is faster on test files

daividssilverio:
### Summary
~This PR makes a little bit of a mess.~
It does it so that we can use Kotest 4 in hope for speedier test writing experience.
I didn't perceive any improvements migrating to Kotest 4.
The one strong point of kotest 3/4 from imoh is the ability to create property tests. If we could replace that with something else, I don't see a reason to keep it.

From my point of view, we should pick one:
- Drop kotest and migrate tests to Junit4/5 only
- Keep kotest 3.x.x and write new tests in Junit4/5 onwards
- Have someone else give it a try on making this work without hurdles

### Technical considerations
- ~I had to migrate gradle from version 5.6.4 to 6.1.1 in order to make the tests discoverable.~
- ~Upgrading to gradle 6.1.1 required me to remove the manual inclusion of the `buildSrc` project on `settings.gradle`, which in turn broke our `com.toggl.buildsrc.dependencies.kt` usage.~
- ~To fix the build, I've moved `com.toggl.buildsrc.dependencies.kt` to `com.toggl.dependencies.kt` and inlined its references used in the root `build.gradle`. All the other projects are still using it.~
- We were actually just missing a dependency to runt the tests. I've reverted the commit that migrated our project to gradle 6.1.1.
- Besides the libraries and steps provided in https://github.com/kotest/kotest, I had to include another few libraries to get everything to work:
- Here's the list:
```kotlin
// required to get it working... I didn't get the spec classes without it
        const val kotestCore = "io.kotest:kotest-core-jvm:$kotestVersion"
// required for the property tests
        const val kotestProperty = "io.kotest:kotest-property-jvm:$kotestVersion"
// required to make the tests discoverable
        const val kotestJvmRunner = "io.kotest:kotest-runner-junit5-jvm:$kotestVersion"
// what was in the kotest repo
        const val kotestRunner = "io.kotest:kotest-runner-junit5:$kotestVersion"
        const val kotestAssertionsCore = "io.kotest:kotest-assertions-core-jvm:$kotestVersion"
```

I couldn't make it work otherwise.
I think we shouldn't keep using kotest to be honest. EDIT: But it works.
Junit4 isn't great as well, but it works with less hurdles and IDE support works just fine.
~Junit5 is mostly gonna have problems similars to the above related to the `dependencies.kt` file, but I guess we'll have to fix it anyway at some point.~
Actually, JUnit5 works fine, the tests are discoverable and nested classes for tests look great and all.

### Relationships
Closes #151 

---
### Pull Request checklist

### Pointers
Think about how you are are will be writing tests in the foreseeable future.
Consider my comments on #151 about the IDE support/responsiveness if we break the tests down into smaller tests.

### Review template

## Merge permissions
🦑 it or close this PR after everybody had a say on it.


pdrj:
Man I didn't know it'd turn out to be such a hassle. I've been advocating for kotest before, but maybe you're right and we should be using junit, if it plays better with roboelectric, db testing and stuff. I think it's OK to leave the current kotest tests there and not lose time migrating. Yes, that's a bit dirty, but we should focus on delivering features now and not refactors. Kotest deps don't pollute apks anyway.

> Junit5 is mostly gonna have problems similars to the above related to the dependencies.kt file, but I guess we'll have to fix it anyway at some point.

is it because of gradle as well?

daividssilverio:
> Man I didn't know it'd turn out to be such a hassle. I've been advocating for kotest before, but maybe you're right and we should be using junit, if it plays better with roboelectric, db testing and stuff. I think it's OK to leave the current kotest tests there and not lose time migrating. Yes, that's a bit dirty, but we should focus on delivering features now and not refactors. Kotest deps don't pollute apks anyway.

After the last commit and making kotest work again, I think it's ok to keep kotest for property tests only.
I wouldn't use it otherwise. My reason being poor IDE support and very "slow to write" cycles.
I think we can reach some balance and use whatever we need depending on the case. By that I mean that I think it's ok to have multiple test frameworks in our codebase, we would just need to define some ground rules for their usage to keep the consistency.

> > Junit5 is mostly gonna have problems similars to the above related to the dependencies.kt file, but I guess we'll have to fix it anyway at some point.
> 
> is it because of gradle as well?

Actually, after some trial and error I found that JUnit5 works fine with the current gradle version we are using, so JUnit5 shouldn't be a problem (maybe except by Android specific tests, roboeletric, espresso, ect).

daividssilverio:
I was doing some research on other options for our testing needs today and found this handy library for property-based testing for the JUnit platform: https://github.com/jlink/jqwik https://jqwik.net/

It's quite well documented and requires minimal setup to work. The usage is similar to regular `JUnit5` tests, but the style is a bit different (although combinable). 

Here's a dumb example:

```kotlin
package com.doivid.teststestbed

import io.kotest.matchers.nulls.shouldNotBeNull
import net.jqwik.api.*
import org.junit.jupiter.api.Nested
import org.junit.jupiter.api.Test
import java.util.*

internal class MyJqwikTest {

    // example from https://jqwik.net/
    @Property
    fun concatenatedStringWithInt(
        @ForAll("shortStrings") aShortString: String,
        @ForAll("10 to 99") aNumber: Int
    ): Boolean {
        val concatenated = aShortString + aNumber;
        return concatenated.length > 2 && concatenated.length < 11
    }

    @Provide
    fun shortStrings(): Arbitrary<String> = Arbitraries
        .strings()
        .withCharRange('a', 'z')
        .ofMinLength(1)
        .ofMaxLength(8)

    @Provide("10 to 99")
    fun numbers(): Arbitrary<Int> = Arbitraries.integers().between(10, 99)

    @Group // this is used to write nested tests for jqwik
    @Nested // this is used for junit's jupiter 
    inner class `We like nested tests` {

        @Test // for junit's jupiter
        fun `I work like this as well`() {
            assert(true)
        }

        @Property // prop test jqwik
        fun suchDataClasses(@ForAll("dataClasses") dataClass: MyCustomClass) {
            dataClass.shouldNotBeNull()
        }

        @Provide // generator for jqwik
        fun dataClasses(): Arbitrary<MyCustomClass> {
            val names = Arbitraries.strings()
            val dates = Arbitraries.longs()
            return Combinators.combine(names, dates)
                .`as`{ name, date -> MyCustomClass(name, Date(date)) }
        }
    }
}

data class MyCustomClass(val name: String, val date: Date)
```

semanticer:
After switching to full `init {}` blocks I don't experience any performance issues in either of Kotest versions. I have no idea why that's the case only for me. Overall I liked Kotest in many ways but if it doesn't work for most of the team, then I guess we should really fall back to a more conservative solution like JUnit 4/5

heytherewill:
We agreed internally on not proceeding with this. Let's just write new tests using jUnit and stop using Kotest for newer stuff, for simplicity/dev speed sake

pinging @semanticer so he knows it's real

PR #155, opened 2020-04-10
🔥Simplify view holders

### LINKED ISSUE #78: Consolidate log item layout files
## Description
`time_entries_log_item.xml` and `time_entries_group_item.xml` are extremely similar files used for similar purposes and their ViewHolders are almost identical. 
We are using a single layout file and ViewHolder class in the old Xamarin app. Both cases have then handled by some switching between different cases and hiding and showing different parts of layout - I believe we can do better than that.

ViewHolders should remain separate, maybe with some shared logic via delegation.

Layout files seem to be a better fit for complete unification, but I would first explore some other ways of sharing layout code without merging files completely. This should add some extra flexibility in case we need to deviate these two views even more in the future. This flexibility is more important than complete DRY especially with rebranding/redesign in front of us.  
Some of the things we can try:
- using `<include />` to share some layouts
- extracting more styles etc
- some clever play with Constraint Sets?
- ViewStubs??
- feel free to join layouts if nothing above feels right

## Definition of done
- [ ] `TimeEntryGroupViewHolder` and `TimeEntryItemViewHolder` share some functionality
- [ ] `time_entries_log_item.xml` and `time_entries_group_item.xml` are either unified to one file or they share some of the common code

pdrj:
### Summary
This simplifies the view models and view holders in main log. In the name of DRY.

### Technical considerations
This only has the code changes for now. I'll continue with layout on Tuesday - but if anyone wants to take a look at this part, they can :).

I'm open to suggestions.

### Relationships

Closes #78 

---
### Pull Request checklist

### Review template

- [ ] The code considers i18n


timer/src/main/java/com/toggl/timer/log/ui/TimeEntryContentViewHolder.kt
```diff
@@ -0,0 +1,33 @@
...
+
+abstract class TimeEntryContentViewHolder(itemView: View) : TimeEntryLogViewHolder(itemView) {
+    val addDescriptionLabel: View = itemView.findViewById(R.id.add_description_label)
+    val project: TextView = itemView.findViewById(R.id.project_label)
+    val description: TextView = itemView.findViewById(R.id.description)
+    val continueButton: View = itemView.findViewById(R.id.continue_button)
+    val billableIcon: View = itemView.findViewById(R.id.billable_icon)
+    val tagsIcon: View = itemView.findViewById(R.id.tags)
+
+    fun bind(item: TimeEntryContentViewModel) {
```
heytherewill, line 19: Shouldn't this be  an `open fun` to show that it should be overriden?

daividssilverio: Just having it open will require casts to work alright.

We could want to have instead:

```kotlin
abstract class TimeEntryContentViewHolder<in T: TimeEntryContentViewModel>(itemView: View)
...
    open fun bind(item: T)
```

heytherewill: +1 to that

pdrj:
I've tried some approaches which don't merge the layout files, but I've found them not really applicable here. 
- include introduces nesting
- merge requires some play with constraints in code, which I think is a bit messy
- stubs don't really make sense here, but they have the same issues as merge.

The layouts are identical, apart from the textview holding group count. I think that in this case, it's fine to have just one file and handle the group count visibility in code. If in the future we find that these two views should work/look more different, we can split and maybe extract some common stuff that doesn't require constraint set gimmicks, like icons etc.

timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogAdapter.kt
```diff
@@ -34,7 +34,7 @@ class TimeEntriesLogAdapter(
         val layoutId = when (viewType) {
             dayHeaderViewType -> R.layout.time_entries_log_header
             flatTimeEntryViewType -> R.layout.time_entries_log_item
-            timeEntryGroupViewType -> R.layout.time_entries_group_item
+            timeEntryGroupViewType -> R.layout.time_entries_log_item
```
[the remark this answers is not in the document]
pdrj: Nice, will do so 👍

PR #156, opened 2020-04-13
Fix main log ripples

### LINKED ISSUE #88: Ensure ripples are present everywhere on the main log
## Description
We need to ensure that the gorgeous input feedback is present

## Definition of done
- [ ] Flat and grouped time entries, when tapped, displays a ripple
- [ ] The continue button in Flat and grouped time entries, when tapped, displays a ripple
- [ ] The date/duration header, when tapped, *does not* display a ripple

daividssilverio:
### Summary
This PR simply fixes the usage of the backgrounds in the main log so that we have functional ripples.

### Technical considerations
The foreground property isn't working properly for our root view in the main log items. It should, but there seems to be a bug. AS does complain about the usage of `android:foreground`, so I've just added a view that serves as the "foreground view".

### Relationships
Closes #88

---
### Pull Request checklist

### Pointers
Check the first commit.
The second commit was just an auto-format (let's not bike shed over layout formatting please).

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it


heytherewill:
AS only complains about it because of Lollipop, right? 🤔

daividssilverio:
Just conforming, yep, everything works alright on api >= 23.
The changes here only affect lollipop. If we decide to move on with dropping lollipop, we can close this PR and issue.

heytherewill:
Closing as not needed

PR #157, opened 2020-04-13
Mark time entries as deleted when committed for deletion

### LINKED ISSUE #148: Mark time entries as deleted in the UI side after committing deletion
## Description
Currently, deleting a time entry (or a group) makes the deleted time entries reappear quickly before being actually removed from the time entries list.

That happens because the `entriesPendingDeletion` in the `TimerState` is cleared before the time entries are actually marked for deletion. 

## Dependencies
Depends on #135

## Definition of done
The solution can be as simple as marking entries with `isDeleted = true` when deletion is committed.
We need to that in two places:

- [ ] When handling `TimeEntriesLogAction.CommitDeletion` at the same time we update the state with `entriesPendingDeletion = setOf()`.
- [ ] When handling `handleDeletingSwipe` when there are already time entries to be deleted, similar to what's done above
- [ ] Related tests are added/fixed

daividssilverio:
### Summary
This PR marks as deleted the list of time entries that will be deleted after:
1. An entry/group is swiped (`TimeEntrySwiped|TimeEntryGroupSwiped`) and there are already pending time entries for deletion
2. The `CommitDeletion` effect is running.

Why? Otherwise the state is updated without the `entriesPendingDeletion` required to hide the time entries that soon will be deleted.

I'm pretty sure we can just delete the map of entries from the state as well, but just marking them as deleted locally seemed like a better idea when I wrote it. What do you think?

### Relationships
Closes #148

---
### Pull Request checklist

### Pointers
Just try it on master and then on this branch.
Try deleting entries/groups in secession before the timeout; try waiting for the deletion timeout to run out on its own as well.

### Review template
## Merge permissions
🦑 it


timer/src/test/java/com/toggl/timer/log/domain/CommitDeletionActionTests.kt
```diff
@@ -92,7 +92,7 @@ class CommitDeletionActionTests : FreeCoroutineSpec() {
                 effect shouldBe listOf()
             }
 
-            "should delete time entries and clear the pending list" - {
+            "should delete time entries, clear the pending list and optimistically update the time entries list that were deleted" - {
```
heytherewill, line 95: Wouldn't it be better to split this into 3 tests?

timer/src/test/java/com/toggl/timer/log/domain/TimeEntryGroupSwipedActionTests.kt
```diff
@@ -54,7 +54,7 @@ class TimeEntryGroupSwipedActionTests : FreeCoroutineSpec() {
             }
 
             "when swiping left" - {
-                "should delete TEs pending deletion (ignoring ids not in state) and put the swiped TEs to pending deletion in state" {
+                "should delete TEs pending deletion (ignoring ids not in state), put the swiped TEs to pending deletion in state, and optimistically update the time entries list that were deleted" {
```
heytherewill, line 57: Same question regarding splitting

timer/src/test/java/com/toggl/timer/log/domain/CommitDeletionActionTests.kt
daividssilverio: We can split the assertion in three parts. Let me do that.

PR #161, opened 2020-04-13
Document app architecture

### LINKED ISSUE #160: Add architecture documentation
## Description
We need to document our architecture, similar to what iOS did.

## Definition of done
- [ ] The readme.md is updated with information about the app, to make it easier for newcomers to understand what's going on

heytherewill:
### Summary
This adds documentation about our architecture, project structure and etc. It adds info about the shared architecture and also about the Android specific bits. It's largely inspired by the iOS app's docs

### Relationships
Closes #160

## Merge permissions
Only myself


Architecture.md
```diff
@@ -0,0 +1,259 @@
...
+
+// or
+
+store.state
+    .map { it.email }
+    .onEach { emailTextField.text = it }
+    .launchIn(scope)
+```
+
+The store can be "viewed into", which means that we'll treat a generic store as if it was a more specific one which deals with only part of the app state and a subset of the actions. More on STORE VIEWS.
```
pdrj, line 71: Is `STORE VIEWS` supposed to be a link?

Architecture.md
```diff
@@ -0,0 +1,259 @@
...
+
+```kotlin
+interface Reducer<State, Action> {
+    fun reduce(state: SettableValue<State>, action: Action): List<Effect<Action>>
+}
+```
+
+The idea is they take the state and an action and modify the state depending on the action and its payload.
+
+In order to dispatch actions asynchronously we will use `Effect`s. Reducers return an array of `Effect`s. The store waits for those effects and dispatches whatever action they emit, if any.
```
pdrj, line 116: Maybe `will` is not needed here, as we are already doing so 🤔

Architecture.md
```diff
@@ -0,0 +1,259 @@
...
+
+## Pullback
+
+There will be one app level reducer that gets injected into the store. This reducer will take the whole `AppState` and the complete set of `AppActions`.
```
pdrj, line 120: This (whole section and the next sections) too would read better in present tense, IMO.

Architecture.md
```diff
@@ -0,0 +1,259 @@
...
+            .map { effect -> effect.map { action -> action?.run(mapToGlobalAction) } }
+    }
+}
+```
+
+After we've transformed the reducer we can use `combine` to merge it with other reducers to create one single reducer that will be injected into the store.
+
+## Store Views
+
+Similarly to reducers and pullback, the store it self can be "mapped" into a specific type of store that only holds some part of the state and only handles some subset of actions. Only this operation is not exactly "map", so it's called `view`.
```
pdrj, line 152:
```suggestion
Similarly to reducers and pullback, the store itself can be "mapped" into a specific type of store that only holds some part of the state and only handles some subset of actions. Only this operation is not exactly "map", so it's called `view`.
```

Architecture.md
heytherewill: Not really, it should be lower case

PR #162, opened 2020-04-13
Add, Load and Persist Task models

### LINKED ISSUE #145: Add and persist the Task model
## Description
We need a task DAO

## Definition of done
- [x] The task Model is created in the models package
- [x] The DatabaseTask class is created in the database package
- [x] Tasks are persisted accordingly in the database
- [x] The Relationships between task and workspace are set
- [x] The Relationships between task and project are set
- [x] A load tasks effect should be created 
- [x] The load tasks effect should be added to the initial load of the app
- [ ] The [relevant documentation](https://www.notion.so/toggl/Loading-860ecdfe1e8249f4a5c493be9b42f2e6) is updated to indicate that the action is now implemented on android

daividssilverio:
### Summary
This PR does the following:
- [x] The task Model is created in the models package
- [x] The DatabaseTask class is created in the database package
- [x] Tasks are persisted accordingly in the database
- [x] The Relationships between task and workspace are set
- [x] The Relationships between task and project are set
- [x] A load tasks effect should be created 
- [x] The load tasks effect should be added to the initial load of the app

### Technical considerations
Nothing special here. It's pretty much the same as the other DAOs and the model has the same properties as iOS (except by the `estimatedSeconds` and `trackedSeconds`, should we have it?)

This is pretty much boilerplate.

### Relationships
Closes #145

### Pull Request checklist

### Pointers
Well, you can check the other models are loaded and the iOS implementation for comparison.

### Review template

## Merge permissions
🦑 it


heytherewill:
That's a good idea, Piotr. But rather than doing it now, I'll create an issue to compare our models with Xamarin ones after this PR is merged. This way we can consolidate everything all at once and stop touching the DB until the sync work starts :)

PR #167, opened 2020-04-14
Add AutocompleteSuggestion type

### LINKED ISSUE #149: Create the Autocomplete suggestion type
## Description
We need a discriminated union to represent all possible autocomplete suggestions that can appear for the user. Since the view is new, we won't copy the [Xamarin implementation](https://github.com/toggl/mobileapp/tree/develop/Toggl.Core/Autocomplete/Suggestions) 1:1. The possible cases are:

- Time Entry Suggestion
- Project Suggestion
- Task Suggestion
- Tag Suggestion
- Create Project Suggestion
- Create Tag Suggestion

It should be modeled as a sealed class

## Dependencies

We need the Task model added in #145 

## Definition of done
- [ ] The type is added to our codebase

daividssilverio:
### Summary
This adds the `AutocompleteSuggestion` sealed class.

### Technical considerations
It is as simple as it could be. It's using our models as a base (they are just data classes).

Feel free to suggest something else.

### Relationships
Closes #149

### Pull Request checklist

### Pointers
Think about the usage of the auto complete suggestions.

### Review template

## Merge permissions
🦑 it


models/src/main/java/com/toggl/models/common/AutocompleteSuggestion.kt
```diff
@@ -0,0 +1,10 @@
+package com.toggl.models.common
+
+sealed class AutocompleteSuggestion {
+    data class TimeEntry(val timeEntry: com.toggl.models.domain.TimeEntry) : AutocompleteSuggestion()
```
heytherewill, line 4: Maybe rename this to `TimeEntrySuggestion` to avoid using fully qualified types everywhere? 🤔

daividssilverio: I thought it would look too verbose. It's `AutocompleteSuggestion.TimeEntrySuggestion` vs `AutocompleteSuggestion.TimeEntry`. We are gonna use with the prefix, won't we?

daividssilverio: If we do import the members of the `AutocompleteSuggestion` it's better to have the more specific name.

daividssilverio: Easy to refactor if we don't like it as well.

PR #174, opened 2020-04-14
Add haptic feedback to some StartEditTimeEntry view actions

### LINKED ISSUE #137: Add haptic feedback to start/edit view events
## Description
As discussed with designers, we need to add that nice little 🐝🐝🐝🐝🐝 🅱️UZZ 🐝🐝🐝🐝🐝 whenever the user does some actions in the edit view
 
## Definition of done
- [ ] There's haptic feedback when the start/edit card opens
- [ ] There's haptic feedback when the user confirms the actions (presses done)
- [ ] There's haptic feedback when the user tries to dismiss the action but is presented with the dialog to confirm if they want to discard the changes

daividssilverio:
### Summary
This PR does the following:
- Adds haptic feedback when the start/edit card opens
- Adds haptic feedback when the user confirms the actions (presses done)

It doesn't:
- Add haptic feedback when the user tries to dismiss the action but is presented with the dialog to confirm if they want to discard the changes, **because we are not showing the confirmation dialog yet.**

### Technical considerations
The Vibrator APIs have some cool predefined effects available for use on Android Q... so I've used them and tried to fallback to similar approaches on earlier api levels.

Feel free to suggest something different. Do tell if you know another way.

### Relationships
Closes #137 (actually... only partially)

### Pull Request checklist

### Pointers
Click around to feel the vibration. You might not feel the haptic feedback if you are not holding the phone in your hand.

### Review template

## Merge permissions
🦑 it, but protest if you think this shouldn't close #137


common-android/src/main/java/com/toggl/common/VibratorExtensions.kt
```diff
@@ -0,0 +1,25 @@
...
+            this.vibrate(VibrationEffect.createPredefined(VibrationEffect.EFFECT_CLICK))
+        }
+        Build.VERSION.SDK_INT >= Build.VERSION_CODES.O -> {
+            this.vibrate(VibrationEffect.createOneShot(Constants.Vibration.defaultDurationInMillis, VibrationEffect.DEFAULT_AMPLITUDE))
+        }
+        else -> {
+            @Suppress("DEPRECATION")
+            this.vibrate(Constants.Vibration.oldApisVibrationPattern, -1)
+        }
+    }
```
heytherewill, line 24: IMO an if that early returns would be much more readable

timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogFragment.kt
```diff
@@ -96,6 +97,7 @@ class TimeEntriesLogFragment : Fragment(R.layout.fragment_time_entries_log) {
                 .distinctUntilChanged()
                 .onEach { isEditViewExpanded ->
                     if (isEditViewExpanded) {
+                        requireContext().performClickHapticFeedback()
```
heytherewill, line 100: I don't think we need to `requireContext` for an action that's so non-critical like a little vibration
```suggestion
                context?.performClickHapticFeedback()
```

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -183,6 +184,7 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
             }
 
             done_action.setOnClickListener {
+                requireContext().performClickHapticFeedback()
```
heytherewill, line 187: I don't think we need to `requireContext` for an action that's so non-critical like a little vibration
```suggestion
                context?.performClickHapticFeedback()
```

common-android/src/main/java/com/toggl/common/VibratorExtensions.kt
daividssilverio: I don't have a strong opinion here. Both look fine to me. I'll make the changes.

daividssilverio:
@heytherewill The compiler was complaining, so I made it happy.
I still used the fragment's context in the `TimeEntriesLogFragment` for sanity reasons... just so that we don't use any reference that might not be there for some reason.
On the dialog, the context is annotated with `@NonNull`.

Feel free to merge this unless you disagree with the changes.

PR #175, opened 2020-04-15
✨ Hook up billable button and update UI according to state

### LINKED ISSUE #153: Hook up the billable action and state to the view
## Description
We already have the action and the state in place, we just need to hook both up

## Definition of done
- [ ] The billable button disappears if the user is not in a pro workspace
- [ ] The billable button emits the appropriate action when tapped
- [ ] The billable button reflects the state of billable/non-billable as per [designs](https://toggl.invisionapp.com/d/main#/projects)

pdrj:
### Summary
This hooks up billable button and updates UI according to state

### Technical considerations
I added some animations because making the transitions just on/off looks pretty weak. These animations are for now inline, but we'll most likely want to reuse them for the other buttons.

I also subscribe to state in `attachBottomView` callback and I'm not sure I'm comfortable doing that (I suspect this might introduce leaks). In my testing, canary found no leaks related to that though 🤔 

I'm open to suggestions. 

### Relationships

Closes #153 

### Pull Request checklist

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -54,10 +62,19 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
 
     private val bottomSheetCallback = BottomSheetCallback()
 
+    private var activeButtonColor = 0
+    private var inactiveButtonColor = 0
+    private val buttonStateChangeDurationMs = 250L
+    private val fullAlpha = 255
+    private val noAlpha = 0
```
heytherewill, line 69: All that could be moved to within some `BottomControlPanelAnimator` class that can handle animating these buttons. This adds the modularity you mention in the PR description without polluting the fragment

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -75,11 +92,12 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
     override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
         return (super.onCreateDialog(savedInstanceState) as BottomSheetDialog).also { bottomSheetDialog: BottomSheetDialog ->
             store.state
-                .mapNotNull { it.editableTimeEntry }
+                .filter { it.editableTimeEntry != null }
+                .map { BottomControlPanelParams(it.editableTimeEntry!!, it.workspaces[it.editableTimeEntry.workspaceId]?.isPro() ?: false) }
```
heytherewill, line 96: This is really hard to read inline, would you mind making it a selector like the one that creates the list of the log?

timer/src/main/res/layout/bottom_control_panel_layout.xml
```diff
@@ -33,9 +34,11 @@
 
         <ImageView
             android:id="@+id/billable_action"
-            android:layout_width="wrap_content"
-            android:layout_height="wrap_content"
-            android:layout_margin="@dimen/grid_1"
+            android:layout_width="32dp"
+            android:layout_height="32dp"
```
heytherewill, line 38: Extract as dimen values

heytherewill:
Requesting @semanticer's review since he created the bottom bar code thingy and can comment better on your concerns

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
pdrj: I've simplified it a bit - see if you like it. If not, I can do a full-blown extraction to another file ;) I think that it's very specific to this dialog though and maybe it's alright to keep it here.

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -54,10 +62,16 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
 
     private val bottomSheetCallback = BottomSheetCallback()
 
+    private var bottomControlPanelAnimator: BottomControlPanelAnimator? = null
```
heytherewill, line 65: I like lateinit var better, since this will realistically never be null when we use it

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -154,16 +170,30 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
...
             this.findViewById<FrameLayout>(com.google.android.material.R.id.container)
         val bottomControlPanel = bottomSheetDialog.layoutInflater.inflate(layoutToAttach, null)
 
+        val billableButton = bottomControlPanel.findViewById<ImageView>(R.id.billable_action)
+        billableButton.isVisible = bottomControlPanelParams.isProWorkspace
+        billableButton.setOnClickListener {
+            store.dispatch(StartEditAction.BillableTapped)
+        }
+
+        store.state
```
semanticer, line 191: It does look a bit scary but technically it should be fine and I also haven't noticed any new leaks.

PR #177, opened 2020-04-15
✨ Add Project module

pdrj:
### Summary
This adds Project module.

### Technical considerations
Some copy/paste, some refactor and some testing.

### Why is it done like this and not some other way?
🤷 

### Relationships

Closes #169 

### Pull Request checklist

- [x] The code does not hinder the app's performance
- [x] The code does not introduce regressions
- [x] The code in this PR is tested thoroughly enough that changing the behavior to an incorrect one breaks tests

### Review template

- [ ] The code is thoroughly tested in well written and resilient manner
- [ ] The code is performant and leads to an optimal user experience
- [ ] The code is testable and does not cut corners for the sake of delivering
- [ ] The code considers older platforms well and will work in older devices just like it will in new ones

- [ ] The code considers light and dark theme
- [ ] The code considers RTL and does not mention left/right
- [ ] The code considers i18n
  - [ ] There are no hardcoded strings in the UI
  - [ ] All strings are included in every language we support OR there's an issue requesting new translations
- [ ] There are no hard-coded numbers referring to dimensions or font sizes
- [ ] The code considers UI elements that are similar and extracts their similarities into styles to avoid duplication

## Merge permissions
Anyone

PR #187, opened 2020-04-16
Create basic project layout

### LINKED ISSUE #170: Add the basic layout for the project view
## Description
We need to [add the basic layout](https://toggl.invisionapp.com/d/main#/console/19556473/408810922/preview) for the add project view. This is to make it simpler to add the actions and test them.

For now, let's not focus on the colors, since they are a bit harder. Let's just add the basicaly layout with the chips/pills and call it a day.

## Definition of done
- [ ] The basic layout is implemented

daividssilverio:
### Summary
This adds the basic layout for the "create project dialog".

### Technical considerations
The implementation is very simple and not final.
The colors and styles for chips are not final.
The icons are not final, I've used sample icons from AndroidStudio (Ping @AlicjaSuska, I couldn't get the new icons from invision, could you export them?)
The shape of the dialog isn't right yet.

From the description #170, I got that this PR should just to the minimal keep other issues moving.

### Relationships
Depends on #177 (I've branched off that PR's branch).
Closes #170

### Pull Request checklist

### Pointers
You can see the dialog by adding the following line somewhere in the `MainActivity`
```
ProjectDialogFragment().show(supportFragmentManager, "test")
```

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it either on develop after #177 is merged or change the PR to target `peter/projboil`.


timer/src/main/res/layout/fragment_dialog_project.xml
```diff
@@ -4,8 +4,141 @@
...
+        android:id="@+id/project_name_divider"
+        android:layout_width="0dp"
+        android:layout_height="1dp"
+        android:layout_marginTop="8dp"
+        android:background="@color/divider"
+        app:layout_constraintEnd_toEndOf="parent"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toBottomOf="@id/project_name_edit_text" />
+
+    <com.google.android.material.chip.Chip
```
pdrj, line 80: Should we add `app:chipBackgroundColor` using a color state list so that it changes color when selected?

daividssilverio: I've made the chip checkable for now. It changes the appearance a bit when selected.

PR #190, opened 2020-04-16
Improve state manipulation in reducers

### LINKED ISSUE #189: Improve the SettableValue class
## Description
Our current `SettableValue` class is not good enough. We can definitely improve on it to make it nicer to use. After some internal discussion we decided that there are a few things we wanna improve:

- Make the class use some kind of mutation context to avoid the annoying `state.value = state.value.copy()` boilerplate
- Make some kind of method to avoid the common case of mutating the state and then returning no effects

## Definition of done
- [ ] The above issues are addressed

heytherewill:
### Summary
This PR implements the things we discussed internally to improve our state manipulation in reducers

There are many possible paths here, so take a look at the code and be as naggy as you want

### Relationships
Closes #189 

### Pull Request checklist

### Pointers
The first Commit is just renaming, so feel free to skip it.

### Review template

## Merge permissions
Only myself


timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
```diff
@@ -26,49 +26,49 @@ class StartEditReducer @Inject constructor(
...
+                        it.copy(billable = !it.billable)
+                    }
                 }
-
-                noEffect()
             }
             StartEditAction.DoneButtonTapped -> {
-                val editableTimeEntry = state.value.editableTimeEntry!!
-                state.value = state.value.copy(editableTimeEntry = null)
+                val editableTimeEntry = state.mapState { editableTimeEntry ?: throw IllegalStateException() }
```
pdrj, line 66: Let's add a message for this exception

timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
```diff
@@ -21,54 +21,54 @@ class StartEditReducer @Inject constructor(
...
```
semanticer, line 66: I don't know
```kotlin
state.get().editableTimeEntry ?: throw IllegalStateException()
```
seems more readable to me. And even better with that invoke function:
```kotlin
state().editableTimeEntry ?: throw IllegalStateException()
```

daividssilverio: Gotta agree. This `state().editableTimeEntry` is faster to read than `mapState` when accessing just one field.

common-feature/src/main/java/com/toggl/common/feature/extensions/MutableValueExtensions.kt
```diff
@@ -0,0 +1,10 @@
+package com.toggl.common.feature.extensions
+
+import com.toggl.architecture.core.Effect
+import com.toggl.architecture.core.MutableValue
+import com.toggl.architecture.extensions.noEffect
+
+fun <T, Action> MutableValue<T>.mutateWithoutEffects(mutateFn: T.() -> T): List<Effect<Action>> {
```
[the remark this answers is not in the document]
daividssilverio: `just` could work 🤔 

```
when (action) {
            LoadingAction.StartLoading -> effects(
                LoadWorkspacesEffect(workspaceRepository, dispatcherProvider),
                LoadProjectsEffect(projectRepository, dispatcherProvider),
                LoadClientsEffect(clientRepository, dispatcherProvider),
                LoadTagsEffect(tagsRepository, dispatcherProvider),
                LoadTasksEffect(taskRepository, dispatcherProvider),
                LoadTimeEntriesEffect(timeEntryRepository, dispatcherProvider)
            )
            is LoadingAction.TimeEntriesLoaded -> state.just { copy(timeEntries = action.timeEntries) }
            is LoadingAction.WorkspacesLoaded -> state.just { copy(workspaces = action.workspaces) }
            is LoadingAction.ProjectsLoaded -> state.just { copy(projects = action.projects) }
            is LoadingAction.ClientsLoaded -> state.just { copy(clients = action.clients) }
            is LoadingAction.TagsLoaded -> state.just { copy(tags = action.tags) }
            is LoadingAction.TasksLoaded -> state.just { copy(tasks = action.tasks) }
        }
```

But it doesn't convey mutation I guess, except that it combos with the `copy` that happens right after it.

PR #191, opened 2020-04-16
✨ Handle NameEntered action

### LINKED ISSUE #171: Add the NameEntered action
## Description
We need to add the action that changes the name of the editable project as the user types

## Dependencies
Depends on both #169 and #170 

## Definition of done
- [ ] The NameEntered action is added to the list of Project actions
- [ ] The action is handled by the reducers
- [ ] The behavior is tested
- [ ] The action is emited by the UI

pdrj:
### Summary
This handles NameEntered action.

It's pretty straightforward.

### Relationships

Closes #171 

### Pull Request checklist

### Pointers
Trigger the fragment in `MainActivity`: `ProjectDialogFragment().show(supportFragmentManager, "test")`

Add some logging in the project reducer. Validate you receive the action when typing in the project name field.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


timer/src/main/java/com/toggl/timer/project/domain/ProjectReducer.kt
```diff
@@ -18,6 +18,10 @@ class ProjectReducer @Inject constructor(
         action: ProjectAction
     ): List<Effect<ProjectAction>> =
         when (action) {
-            is ProjectAction.NameEntered -> noEffect()
+            is ProjectAction.NameEntered -> {
+                val editableProject = state.value.editableProject!!
```
heytherewill, line 22: I'd rather see and `state.value.editableProject ?: throw InvalidStateOperation("Project can never be null")`, which is easier to track if we ever get to it than a null reference exception

timer/src/main/java/com/toggl/timer/project/domain/ProjectReducer.kt
```diff
@@ -18,6 +18,10 @@ class ProjectReducer @Inject constructor(
...
+                state.value = state.value.copy(editableProject = editableProject.copy(name = action.name))
```
heytherewill, line 23: Nested copying can be avoided by using arrow. See how in the start edit reducer

timer/src/main/java/com/toggl/timer/project/domain/ProjectReducer.kt
```diff
@@ -18,6 +18,12 @@ class ProjectReducer @Inject constructor(
         action: ProjectAction
     ): List<Effect<ProjectAction>> =
         when (action) {
-            is ProjectAction.NameEntered -> noEffect()
+            is ProjectAction.NameEntered -> {
+                state.value.editableProject ?: throw IllegalStateException("editableProject must not be null")
```
daividssilverio, line 22: Oh, gotta use the new notation. This should look like:
```
is ProjectAction.NameEntered -> state.mutateWithoutEffects {
                editableProject ?: throw IllegalStateException("editableProject must not be null")
                ProjectState.editableProject.modify(this) {
                    it.copy(name = action.name)
                }
            }
```

daividssilverio: Ok... I forgot to set an `editableProject` to the state before testing... it started crashing of course.

Issue #169, opened 2020-04-14
Add the project module boilerplate


heytherewill:
## Description
We need to add the boilerplate needed for the module responsible for adding new projects. We just need the boilerplate, no need to have concrete actions or handling.

One thing that needs to be added to the state is a `editableProject` type, similar to the editable time entry. That property is the one that will control the visibility of the project view

## Definition of done
- [ ] The ProjectReducer is added
- [ ] The ProjectState is added
- [ ] The ProjectDialogFragment is added
- [ ] The DI boilerplate needed is added
- [ ] All that is included in the `project` package inside the `:timer` module

PR #194, opened 2020-04-16
Nest project inside time entry

heytherewill:
### Summary
This is a tiny change to help us Make Illegal States Irrepresentable ™️ 

### Technical considerations
Given that you can only ever add a project when you are editing a time entry, nest these two is a logical conclusion

The other way allows for illegal state, which is a big nono

### Relationships
Refs #169 (LMAO)

### Pull Request checklist

### Pointers
Simple enough, just look at the changes

### Review template

## Merge permissions
Anyone, anytime


timer/src/main/java/com/toggl/timer/project/domain/ProjectState.kt
```diff
@@ -11,16 +11,20 @@ data class ProjectState(
...
 
         fun toTimerState(timerState: TimerState, projectState: ProjectState) =
-            timerState.copy(
-                projects = projectState.projects,
-                localState = timerState.localState.copy(
-                    editableProject = projectState.editableProject
+            with(timerState) {
+                copy(
+                    projects = projectState.projects,
+                    localState = localState.copy(
```
daividssilverio, line 22: Can't optics be used to do something about this? It's a bit unpleasant to see the nested copies.
I don't know how to use it properly yet 🤔

heytherewill:
@pdrj When that happens, we can certainly change this! Until then, I'd rather have that possibility irrepresentable and have the compiler ensure it :)

heytherewill:
@daividssilverio Good idea with the lenses. We can't apply `@optics` directly to the local state because of the internal visibility, but that doesn't prevent us from having our own `lens` :)

PR #196, opened 2020-04-17
Setup database testing infrastructure

### LINKED ISSUE #144: Setup database testing infrastructure

daividssilverio:
### Summary
This PR:
- Sets up robolectric and room database testing for the `database-android` module.
- Adds a basic TestUtil helper

### Technical considerations
The tests are running locally, not on actual devices, so there's no need for instrumentation yet.
Although that's not what's suggested on the android's developer documentation, it should be enough for our needs for now.

For room and roboeletric to recognized in the tests, I had to change our source sets so that they include generated sources.

Because it's faster to run locally and instrumentation tests are pain to run...

### Relationships
Closes #144

### Pull Request checklist

### Pointers
Try running the tests.

### Review template

## Merge permissions
🦑 it


daividssilverio:
Oh well... tests won't pass until roboeletric releases another version with the fixes for the error that's popping out. It seems like https://github.com/robolectric/robolectric/pull/5550 fixes it, but there isn't a 4.3.2 version and the 4.4-SNAPSHOT doesn't seem to be working.

```
Illegal reflective access by org.robolectric.util.ReflectionHelpers
```

Or we change our CI to use Java 8.

daividssilverio:
If we want to keep keep the tests and run them locally only for the time being, another option would be adding `@Ignore` to the database tests and running them from the IDE only until we manage to have the setup ready for use with our CI.

daividssilverio:
@pdrj 
> I'm completely OK with using java8 for our CI. We target v8 in our builds anyway 🤔

Cool

> That being said, how do you know we're using java10 in the CI pipeline at the moment? Their docs say that java8 is pre-installed and the dockerfile for android only seems to be downloading jdk7 for some reason...

When I saw the test failing, I run the gradle command used to run the tests locally and replicated the same set of exceptions.

```
com.toggl.database.BaseDatabaseTest > initializationError FAILED
    java.lang.Exception
```

I saw the workaround of running using jdk 8 to run gradle on an issue here: https://github.com/robolectric/robolectric/issues/4776 (it's so old... hmm... maybe they won't fix it)

> https://devcenter.bitrise.io/tips-and-tricks/android-tips-and-tricks/#installing--using-java-version-x
> https://hub.docker.com/r/bitriseio/docker-android/dockerfile

I'll take a look at it to see if the problem is somewhere else.

daividssilverio:
Hey @pdrj . You were right! Bitrise is already using jdk-7!
I run the tests locally using jdk-12 and got the errors I was talking about...
After setting my `JAVA_HOME` to target jdk-8, I managed to find the actual problem:
The base test class was not abstract and the JUnit4 runner wasn't finding any tests to run there, so it 💥ed . All it had to be done was to make the base test class abstract.

Now it works!

daividssilverio:
> I don't think that's done yet 🤔 Should we at least have the gradle changes described here: https://developer.android.com/training/data-storage/room/migrating-db-versions?

Yup, will do.

PR #198, opened 2020-04-20
✨ Add AutocompleteSuggestionTapped action

### LINKED ISSUE #179: Add the AutocompleteSuggestionTapped action
## Description
We need to add the action to handle when the user selects an autocomplete suggestion from the list
## Definition of done
- [ ] The action is added
- [ ] Proper handling will be done by other issues, so for now let's just return no effects and call it a day

pdrj:
### Summary
This adds [an empty] AutocompleteSuggestionTapped action

This is pretty straightforward.

### Relationships

Closes #179 

### Pull Request checklist

### Pointers
There isn't much to test. Just look at the code I guess.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
```diff
@@ -75,6 +75,8 @@ class StartEditReducer @Inject constructor(
             }
             is StartEditAction.AutocompleteSuggestionsUpdated ->
                 state.mutateWithoutEffects { copy(autocompleteSuggestions = action.autocompleteSuggestions) }
+            is StartEditAction.AutocompleteSuggestionTapped ->
+                state.mutateWithoutEffects { this }
```
semanticer, line 79: Why not just:
```suggestion
                noEffect()
```

timer/src/test/java/com/toggl/timer/startedit/domain/AutocompleteSuggestionTappedActionTests.kt
```diff
@@ -0,0 +1,29 @@
...
+    val reducer = StartEditReducer(repository, dispatcherProvider)
+
+    @Test
+    fun `should return no effect`() = runBlockingTest {
+        reducer.testReduce(
+            initialState,
+            action = StartEditAction.AutocompleteSuggestionTapped(mockk())
+        ) { state, effects ->
+            assertNoEffectsWereReturned(state, effects)
+        }
```
semanticer, line 27: Why not just

```suggestion
            action = StartEditAction.AutocompleteSuggestionTapped(mockk()),
            testCase = ::assertNoEffectsWereReturned
        )
```

timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
pdrj: I wanted to make it clear that it will soon be modified, but I can change that 👍

PR #199, opened 2020-04-20
✨ Include cursor position in DescriptionEntered action

### LINKED ISSUE #180: Update the DescriptionEntered action to include the cursor position
## Description
Since we are implementing the suggestion mechanism, the current position of the cursor is pretty important.
It doesn't need to be stored in the state, since it'll just be later used by an effect to calculate the suggestions
Emitting this action when the cursor changes should be handled in another issue.
## Definition of done
- [ ] The action is updated to match the docs
- [ ] The place where the action is dispatched is changed to return the cursor position as well 

pdrj:
### Summary
This includes cursor position in DescriptionEntered action

This seems simple enough.

### Relationships

Closes #180 

### Pull Request checklist

### Pointers
Add some logging in the reducer and modify the description. See that the position is seen by the reducer.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


heytherewill:
I'm putting this on-hold while we discuss the implementation internally

PR #200, opened 2020-04-20
✨ Display current duration on start/edit fragment

### LINKED ISSUE #181: Properly display the current duration
## Description
We need to ensure that the duration on the top right corner of the card is correct
## Definition of done
- [ ] If the duration is set, we just use it along with the existing format method used in the log
- [ ] If there's no duration, we calculate the duration based on the current date and then use the existing format method
- [ ] If there's no duration, we need to update the number every second

pdrj:
### Summary
This displays the current duration on start/edit fragment

I'm open to suggestions.

### Relationships

Closes #181 

### Pull Request checklist

### Pointers
Open the edit fragment for running / stopped / not yet persisted TE and validate the time indication is correct.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -235,4 +250,35 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
...
+    }
+
+    private fun StartEditState.getTimeEntryForEditable(): TimeEntry? {
+        val editableId = this.editableTimeEntry?.ids?.getOrNull(0)
+
+        editableId?.let {
+            return this.timeEntries[editableId]
+        }
+
+        return null
```
semanticer, line 282: How about just

```kotlin
timeEntries[editableId].takeIf { editableId != null } 
```

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -235,4 +250,35 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
             )
         }
     }
-}
+
+    private fun TextView.setDuration(timeEntry: TimeEntry?) {
```
semanticer, line 254: This method is a little bit hard to read & understand. I can't think of any clever suggestions right now but I still want to flag this observation for you and other reviewers in case anyone has ideas on how to improve this.

common/src/main/java/com/toggl/common/Constants.kt
```diff
@@ -2,6 +2,7 @@ package com.toggl.common
 
 object Constants {
     const val timeEntryDeletionDelayMs = 5000L
+    const val elapsedTimeIndicatorUpdateDelayMs = 1000L
```
semanticer, line 5: We could move these constants to our resource system as they are related to UI. Although in this case there is no big advantage in doing this, so I guess it's up to you 🤷‍♂️

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
pdrj: That's cleaner - thanks.

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
pdrj: Yep, some stuff can be extracted. I'll see what can be done ;)

common/src/main/java/com/toggl/common/Constants.kt
pdrj: Hm, I honestly don't know what's better. It's pretty simple to just have it here 😄

pdrj:
Yeah, we were in bad state basically and it has to be hidden. The fragment got recreated and the TextView restored, but the value didn't come from state, but from the usual Android restore function.

pdrj:
In other words, the `EditableTimeEntry` was null, but the fragment was showing and we can't have that.

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -235,4 +253,31 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
...
+        }
+        view.postDelayed(timeIndicatorScheduledUpdate, elapsedTimeIndicatorUpdateDelayMs)
+    }
+
+    private fun StartEditState.getTimeEntryForEditable(): TimeEntry? {
+        val editableId = this.editableTimeEntry?.ids?.getOrNull(0)
+
+        return timeEntries[editableId].takeIf { editableId != null }
+    }
+}
```
heytherewill, line 283: This method is incorrect. The `editableTimeEntry` type needs to have start and duration fields and these should either be set to the `now()` from the timeService (for non-started entries) or from the values from the entry you are editing

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -235,4 +253,31 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
...
+        this.text = durationToSet.formatForDisplaying()
+    }
+
+    private fun scheduleTimeEntryIndicatorUpdate(timeEntry: TimeEntry, view: TextView) {
+        timeIndicatorScheduledUpdate?.let { view.removeCallbacks(it) }
+        timeIndicatorScheduledUpdate = Runnable {
+            view.setDurationAndScheduleUpdates(timeEntry)
+        }
+        view.postDelayed(timeIndicatorScheduledUpdate, elapsedTimeIndicatorUpdateDelayMs)
+    }
```
heytherewill, line 276: I really dislike postDelayed, since it's hard on the eyes. Why not create a suspend fun that asynchronously waits for a second and then updates the UI, making cancellations trivial:

```
fun startDurationUpdates(): Job {
    return lifecycleScope.launch {  
        while(true) {
            // UI update goes here
            delay(1000)
        }
    }
}

// onCreate
updateJob = startDurationUpdates()


//onDestroy
updateJob.cancel()
}
```

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -149,20 +157,30 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
...
+        lifecycleScope.launchWhenStarted {
+            store.state
+                .filter { it.editableTimeEntry == null }
+                .distinctUntilChanged()
+                .onEach {
+                    if (dialog?.isShowing == true) {
+                        findNavController().popBackStack()
+                    }
+                }
+                .collect()
```
heytherewill, line 175: Why the collect?

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -149,20 +157,30 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
...
             .distinctUntilChanged()
-            .onEach {
-                if (isVisible) {
-                    findNavController().popBackStack()
-                }
-            }
+            .onEach { time_indicator.setDurationAndScheduleUpdates(it) }
             .launchIn(lifecycleScope)
+
+        lifecycleScope.launchWhenStarted {
```
heytherewill, line 166: Why does this need `launchWhenStarted` unlike the rest?

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
pdrj: To me `while(true)` is even harder on the eyes. I agree that cancelling is a bit easier.

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
pdrj: Because otherwise the flow is cold and doesn't run.

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
pdrj: Because we're doing a fragment transaction in there. I'd rather be on the safe side.

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
pdrj: OK, so EditableTimeEntry is basically missing some info.

I don't see how the method is incorrect though 🤔 If we don't find the corresponding time entry in state, then we'll display a 'zero' duration (the TE is not started nor saved yet). This case should not happen though, yet. Otherwise, if there's a 'Duration' in the TE, we use it. If there isn't, we compute the duration based on the current time.

It just occurred to me that the method is wrong for groups, so maybe that's what you meant 😄As for the changes you proposed, I don't think we should be setting a start time on EditableTimeEntry by ourselves 🤔. I think this would happen in case of new manual TEs, but right now the start should always be copied over from the 'real' TE, or be null in case of groups. Am I correct?

pdrj:
@heytherewill I've made the changes. In case you're wondering - I'm using `launchWhenCreated` so the `Job` is cancelled automatically.

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -235,4 +256,29 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
...
+        if (this.text != newDurationText) {
+            this.text = newDurationText
+        }
+    }
+
+    private fun EditableTimeEntry.getDurationForDisplaying() =
+        when (this.duration) {
+            null -> {
+                Duration.between(this.startTime, timeService.now())
+            }
```
heytherewill, line 271:
```suggestion
            null -> Duration.between(this.startTime, timeService.now())
```

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -235,4 +256,29 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
...
+
+    private fun TextView.setDurationIfDifferent(duration: Duration) {
+        val newDurationText = duration.formatForDisplaying()
+        if (this.text != newDurationText) {
+            this.text = newDurationText
+        }
+    }
+
+    private fun EditableTimeEntry.getDurationForDisplaying() =
+        when (this.duration) {
```
heytherewill, line 268: Don't you need to check if there is a startTime and return 0 otherwise?

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -235,4 +256,29 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
...
+            null -> {
+                Duration.between(this.startTime, timeService.now())
+            }
+            else -> this.duration
```
heytherewill, line 272: Do we even need to schedule and tick this every second if there's a duration?

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
daividssilverio: I'm assuming so, the app crashed with `this.startTime` being null on my first try.

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -235,4 +256,29 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
...
+    private fun TextView.setDurationIfDifferent(duration: Duration) {
+        val newDurationText = duration.formatForDisplaying()
+        if (this.text != newDurationText) {
+            this.text = newDurationText
+        }
+    }
+
+    private fun EditableTimeEntry.getDurationForDisplaying() =
+        when (this.duration) {
+            null -> {
```
daividssilverio, line 269:
```suggestion
        when {
            startTime == null -> Duration.ZERO
            duration == null -> {
```

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -235,4 +256,29 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
...
+                Duration.between(this.startTime, timeService.now())
+            }
+            else -> this.duration
+        }
+
+    private fun scheduleTimeEntryIndicatorUpdate(editableTimeEntry: EditableTimeEntry) {
+        timeIndicatorScheduledUpdate?.cancel()
+        timeIndicatorScheduledUpdate = lifecycleScope.launchWhenCreated {
```
daividssilverio, line 277: 🤔 I wonder if using a chronometer text view wouldn't be simpler 🤔
We wouldn't need to handle the ticks manually, we'd just need to update it when the `startTime` changes or a duration is set.

https://developer.android.com/reference/android/widget/Chronometer

Plus, we'd get to see auto completion with [easter eggs](https://developer.android.com/reference/android/widget/Chronometer#isTheFinalCountDown())

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
pdrj: Oh right, sorry about that. I forgot about the case where you press the description field.

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
pdrj: No, in fact I had implemented it so that it only happens if there's no duration. This however made the store subscription a tad too complicated (there was onEach, then a map and another onEach). I thought it's best to call update even on TEs whose duration won't change (and only set the text if it changed, to not do unnecessary redraws).

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
pdrj: Hm, I don't know... With the current approach (material chip) we get a lot of styling for free. We also can just set a string on it, whereas chronometer would also need some format handling. I'm not sure what is better - what do you think?

pdrj:
>Don't you need to cancel the timeIndicatorScheduledUpdate when destroying the fragment?

Nope:
> In case you're wondering - I'm using launchWhenCreated so the Job is cancelled automatically.

I'd have to cancel it if just `launch` was used.

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
daividssilverio: I didn't notice that we were using a chip for that.

I think it's worth trying to simplify all the logic that's being used to show the timer, but since it's done and working, I think we can create an issue and try to improve it later on.

We can get some of the styling magic using a `ChipDrawable`. Since I'm not particularly experienced with the chronometer myself, I won't request changes for that. But do you mind creating and issue for that so that we give it another look at some point?

pdrj: I just gave `ChipDrawable` a go and I don't think that's the way to go honestly. I might be using it wrong, but to me it seems like we'd have to modify its state (like bounds and stuff) on each chronometer tick, it's quite ugly IMO 🤷I think it's meant to be used for stuff that gets wrapped in a chip shape and then the content stays the same. If you have experience using that and I'm doing it wrong, please let me know. We could alternatively just use a hidden chronometer and its callback to set text on a chip, but that's bad too.

Also, `Chronometer` uses `postDelayed` internally and @heytherewill doesn't like that 😄 https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/widget/Chronometer.java#345

Of course, we could just style `Chronometer` to look like a chip, but then we'd also have to add these nifty animations that action chip gets on presses...

Let me know what you think @daividssilverio

timer/src/main/java/com/toggl/timer/common/domain/EditableTimeEntry.kt
```diff
@@ -2,32 +2,40 @@ package com.toggl.timer.common.domain
...
                 workspaceId = timeEntry.workspaceId,
                 description = timeEntry.description,
+                startTime = timeEntry.startTime,
+                duration = timeEntry.duration,
                 billable = timeEntry.billable,
                 editableProject = null
             )
 
-        fun fromGroup(ids: List<Long>, groupSample: TimeEntry) =
+        fun fromGroup(ids: List<Long>, groupSample: TimeEntry, timeEntries: Map<Long, TimeEntry>) =
```
heytherewill, line 33: This feels really wasteful. Why not pass, instead of a list of ids, a list of TimeEntry? This way we don't need that weird validation inside a model while maintaining the correct logic for calculating duration at a single point

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
heytherewill: Agreed

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
daividssilverio: It looks like too much trouble using the `ChipDrawable` + `Chronometer`, let's just use your first idea, since it works already.

> Also, Chronometer uses postDelayed internally and @heytherewill doesn't like that 😄 https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/widget/Chronometer.java#345

well, look at that!

timer/src/main/java/com/toggl/timer/common/domain/EditableTimeEntry.kt
pdrj: I did it like so because I didn't want to construct any temporary lists on the fly, but I do it anyway in a `map` 😄 I changed it so that in now takes a list of TimeEntry only and computes the other properties based on that. It blends in nicely with how we handle the Editable based on a single time entry.

timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogReducer.kt
```diff
@@ -28,108 +28,105 @@ class TimeEntriesLogReducer @Inject constructor(
         state: MutableValue<TimeEntriesLogState>,
         action: TimeEntriesLogAction
     ): List<Effect<TimeEntriesLogAction>> =
-            when (action) {
```
heytherewill, line 31: Please refrain from making formatting changes unless that's the focus of the PR. This makes reviewing really hard

pdrj: Sure, sorry about that. I made them in a separate commit, but this doesn't help in this case, does it? 🙈

timer/src/test/java/com/toggl/timer/common/domain/EditableTimeEntryTests.kt
```diff
@@ -0,0 +1,38 @@
...
+
+        val ids = listOf<Long>(1, 2)
+        val editable = EditableTimeEntry.fromGroup(timeEntries.filter { it.id in ids })
+
+        editable.duration shouldBe Duration.ofMinutes(8)
+        editable.startTime shouldBe null
+    }
+
+    @Test
+    fun `for single TEs, should copy over the duration and start time`() {
```
daividssilverio, line 29: what about running time entries?

PR #201, opened 2020-04-21
✨ Add and dispatch tag button pressed action

### LINKED ISSUE #182: Add the TagButtonTapped action
## Description
This is the action dispatched when the tag button in the bottom bar is tapped.
## Definition of done
- [ ] The Action is added
- [ ] The action is dispatched when you tap the bottom bar tag button
- [ ] When the action is handled, it adds a `#` character to the description
- [ ] Tests ensure behavior is correct

pdrj:
### Summary
This adds and dispatches tag button pressed action

### Technical considerations
It's branched off the project button pressed branch.

This seems simple enough.

### Relationships

Closes #182 

### Pull Request checklist

### Pointers
Type in the description field quickly, interfere that with pressing the tag button and validate '#' appear at correct places. Do it for running/new/old time entries.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -217,6 +216,14 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
                 context.performClickHapticFeedback()
                 store.dispatch(StartEditAction.DoneButtonTapped)
             }
+
+            project_action.setOnClickListener {
+                store.dispatch(StartEditAction.ProjectButtonTapped)
+            }
+
+            tag_action.setOnClickListener {
+                store.dispatch(StartEditAction.TagButtonTapped)
```
semanticer, line 225: Turns out neither of these listeners (`project_action` and `tag_action`) work for me 🤔. They are set but never actually triggered :-(

pdrj: If that happens on new time entries only, then check out this branch with the latest changes. Specifically 9198ea7

common-android/src/main/java/com/toggl/common/EditTextExtensions.kt
```diff
@@ -0,0 +1,24 @@
+package com.toggl.common
+
+import android.text.Editable
+import android.text.TextWatcher
+import android.widget.EditText
+import androidx.core.widget.doAfterTextChanged
+
+fun EditText.setSafeText(newText: String) {
+    val lastDispatchedDescription = tag as? String
```
heytherewill, line 9: `this.tag` would really improve readability here

PR #220, opened 2020-04-22
Add the MathUtil for the wheel math

### LINKED ISSUE #203: Port Math utils used in the wheel from the old apps
## Description
This should be a straight port from the current implementation.
Let's take a look on what's being used, port it and be done with it.

## Dependencies
None.

## Definition of done
Basically, we need:
- [x] The helper methods used in all wheel components are ported:
  - [x] ToAngleOnTheDial
  - [x] ToAngle
  - [x] ToPositiveAngle
  - [x] AngleToTime
  - [x] AngleBetween
  - [x] DistanceSq
  - [x] PointOnCircumference
  - [x] IsBetween
  - [x] PingPongClamp
  - [ ] ToRadians
- [x] The constants used in the wheel are also ported:
  - [x] FullCircle
  - [x] MinutesInAnHour
  - [x] SecondsInAMinute

daividssilverio:
### Summary
This is a [port of the Math utils](https://github.com/toggl/mobileapp/blob/develop/Toggl.Shared/Math.cs) used to build the wheel components.

### Technical considerations
The port isn't 1:1 because the APIs are a bit different.
`ToRadians` isn't needed for the wheel stuff by the way. It's only used in the donut chart view.

### Relationships
Closes #203

### Pull Request checklist

### Pointers
Compare it with the old implementation.

### Review template

☝️ about that, not sure if writing tests for these utils are worth it or not... although possible, I have to say they would be a matter of testing the functions with some values and writing tests with those values to make sure they won't change in the future.

## Merge permissions
🦑 it


timer/src/main/java/com/toggl/timer/startedit/util/MathUtil.kt
```diff
@@ -0,0 +1,72 @@
...
+
+fun angleBetween(a: PointF, b: PointF): Float = kotlin.math.atan2(a.y - b.y, a.x - b.x)
+
+fun distanceSq(a: PointF, b: PointF): Float {
+    val dx = a.x - b.x
+    val dy = a.y - b.y
+    return dx * dx + dy * dy
+}
+
+fun PointF.updateWithPointOnCircumference(center: PointF, angle: Double, radius: Double) {
```
pdrj, line 43: This returns a new point in the old implementation - just making sure it's intended to do a `set` here instead.

Issue #207, opened 2020-04-21
Add the wheel widget


daividssilverio:
## Description
The Wheel™ is going to be ported almost 1:1 from the Xamarin apps (api differences apply).
The Wheel™ is composed of the following components:
- The [WheelBackgroundView](https://github.com/toggl/mobileapp/blob/develop/Toggl.Droid/Views/EditDuration/WheelBackgroundView.cs) which handles drawing the fixed background of the wheel (clockdial and empty wheel background).
- The [WheelForegroundView](https://github.com/toggl/mobileapp/blob/develop/Toggl.Droid/Views/EditDuration/WheelForegroundView.cs) (the main component that handles most of the user interaction, drawing and exposes most of the API used to control the widget).
- The [WheelDurationInputView](https://github.com/toggl/mobileapp/blob/develop/Toggl.Droid/Views/EditDuration/WheelDurationInput.cs) which handles manual duration input in minutes.

## Dependencies
It needs the base layout that should be added by #202 
Wheel Background View #204 
Wheel Foreground View #205 
Wheel Input Duration View #206

## Definition of done
- [ ] The WheelBackgroundView is added
- [ ] The WheelForegroundView is added
- [ ] The WheelDurationInputView is added
- [ ] The components are combined into one and the placeholder added by #202 is replaced with the actual Wheel™

PR #221, opened 2020-04-22
Add the wheel background view

### LINKED ISSUE #204: Create the WheelBackground view
## Description

The [WheelBackgroundView](https://github.com/toggl/mobileapp/blob/develop/Toggl.Droid/Views/EditDuration/WheelBackgroundView.cs) that draws the base wheel background, including the clock dial and the "empty wheel".

## Dependencies
Needs the Math utils from #203

## Definition of done
- [ ] The wheel background view was ported to kotlin
- [ ] There's no expensive drawing/painting elements being created in the draw loop
- [ ] The clock dial is being drawn
- [ ] The empty wheel's background is being drawn

daividssilverio:
### Summary
This is the port for the WheelBackgroundView from the old android app: https://github.com/toggl/mobileapp/blob/develop/Toggl.Droid/Views/EditDuration/WheelBackgroundView.cs

### Technical considerations
There's nothing special going on.
I've added some niceties since we use AndroidStudio now :)
After running the project once, you can use the layout preview together with the custom attributes to preview the view.

I've added default colors and dimensions for the view inside the timer module.
I guess we can use proper colors considering dark mode when styling the whole thing on #207 

Feel free to make suggestions to improve how it's done.

### Relationships
Closes #204 

### Pull Request checklist

### Pointers
Please shout if you are seeing something that can be improved.

You can replace the `extended_options`'s `TextView` in the `timer/src/main/res/layout/fragment_dialog_start_edit.xml` layout file with the snipped below to preview the wheel background 

```xml
 <com.toggl.timer.startedit.ui.editduration.WheelBackgroundView
        android:id="@+id/extended_options"
        android:layout_width="300dp"
        android:layout_height="300dp"
        app:layout_constraintBottom_toBottomOf="parent"
        app:layout_constraintStart_toStartOf="parent"
        app:layout_constraintTop_toBottomOf="@+id/bottom_divider" />
```

It looks like this:
Screen Shot 2020-04-22 at 7 10 15 PM

If you open the StartEdit dialog view with the snippet above, it looks wonky for some reason, but I rather figure out that later on #207 
image

### Review template
- [ ] The code considers i18n

## Merge permissions
Only me. This PR needs to be rebased with the changes from #220 before being merged


timer/src/main/java/com/toggl/timer/startedit/ui/editduration/shapes/ClockDial.kt
```diff
@@ -0,0 +1,47 @@
...
+
+    private val paint = Paint(Paint.ANTI_ALIAS_FLAG).apply {
+        textAlign = Paint.Align.LEFT
+        setTextSize(textSize)
+        color = textColor
+        typeface = Typeface.create("sans-serif", Typeface.NORMAL)
+    }
+
+    fun onDraw(canvas: Canvas?) = canvas?.run {
+        for (minute in (0 until minutesInAnHour) step 5) {
```
pdrj, line 33: :sogood:

timer/src/main/java/com/toggl/timer/startedit/ui/editduration/shapes/Wheel.kt
```diff
@@ -0,0 +1,33 @@
+package com.toggl.timer.startedit.ui.editduration.shapes
+
+import android.graphics.Canvas
+import android.graphics.Paint
+import android.graphics.RectF
+import androidx.annotation.ColorInt
+
+class Wheel(
```
pdrj, line 8: Just wondering - since this is just a wheel, can't we use an xml resource and set it as a background, rather than having this custom class? I don't know if `setHidden` is needed though and if it is, then we'd have to modify the bg from code, which is a bit ugly.

daividssilverio: The `Wheel` is also used in the `WheelForegroundView`, we hide it when drawing overlapping arcs.

When there are 1.0+ loops in the foreground view, the background loop is just a `Wheel` and the foreground loop is an arc. We keep changing its color to match the previous colour of the rainbow loop.

When there's less than 1 loop completed, the `WheelBackgroundView`'s wheel still appears on the screen, and the `WheelForegroundView` wheel is hidden.

And of course, we can try replacing it with an xml resource, but I'm not sure if it's worth loading a resource, having to tint it manually, that kind of stuff.

common-android/src/main/java/com/toggl/common/ContextExtensions.kt
```diff
@@ -1,10 +1,17 @@
...
     val vibrator = getSystemService<Vibrator>()
     vibrator?.performClickEffect()
-}
\ No newline at end of file
+}
+
+@Suppress("DEPRECATION")
+fun Context.getSafeColor(@ColorRes colorResId: Int) =
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) getColor(colorResId)
+    else resources.getColor(colorResId)
```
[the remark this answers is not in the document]
daividssilverio: Yup! That's exactly the same code! Good catch.

timer/src/main/java/com/toggl/timer/startedit/ui/editduration/shapes/Wheel.kt
pdrj: Thanks for the explanation. In this case, it's not worth going xml.

PR #223, opened 2020-04-23
📐 Add a basic edit layout

### LINKED ISSUE #202: Create the edit duration base layout
## Description
This issue is here to get things moving as fast as possible.
Let's add the basic layout ahead of time so that we can bind the reducer actions more easily as we implement them.

Designs instructions for the [expanded view here](https://toggl.invisionapp.com/d/main?origin=v7#/console/19556473/409766523/preview?scrollOffset=5527).

## Dependencies
None.

## Definition of done
- [ ] The base layout for the edit duration view was added to the expanded StartEdit view
- [ ] The title labels for start/end were added
- [ ] The start time clickable label was added
- [ ] The start date clickable label was added
- [ ] The end time clickable label was added
- [ ] The end date clickable label was added
- [ ] The "stop" button label that appears for running time entries was added
- [ ] A placeholder for wheel widget was put in place

pdrj:
### Summary
This adds a basic edit layout

### Technical considerations
Screenshot 2020-04-23 at 14 34 02

- I had to wrap the bottom line of ☝️ in linear layout because I couldn't [in a reasonable amount of time] make all 3 elements to be centered related to the header with just a flat constraint layout 😢If you know a way to do it, please share :) 
- The second commit wraps the whole thing [except the close button, handle and time indicator] in a nested scroll view - otherwise the wheel placeholder gets cut off. I don't like it, but I don't think there's a way around that. 
- We have `AlphaSlideAction` in the fragment. I don't think it worked for me before (or I can't see it), but now it just can't work as it's referencing a `Group` and setting group's alpha has no effect. **Should we just remove the animation? Can you see it working on your device?** 
- I have not handled changing the labels on start/end time labels in case the TE has no duration or is not running yet. For this to be implemented, I'd have to have my changes regarding displaying the correct time on the indicator. I could rebase on that branch and add those features, I just don't know what the expectation on this issue is, since it's just an `S`.

I'm open to suggestions.

### Relationships

Closes #202 

### Pull Request checklist

### Pointers
Use this with keyboard on/off; expand, collapse. See if it behaves correctly.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


timer/src/main/res/layout/fragment_dialog_start_edit.xml
```diff
@@ -36,78 +36,187 @@
...
-            android:id="@+id/chip_group"
-            android:layout_width="wrap_content"
-            android:layout_height="wrap_content"
-            app:singleLine="true">
+                    <com.google.android.material.chip.Chip
+                        android:id="@+id/add_project"
+                        style="@style/Widget.MaterialComponents.Chip.Action"
+                        android:layout_width="wrap_content"
+                        android:layout_height="wrap_content"
+                        android:text="Add a Project"
```
semanticer, line 87: Maybe it's time to extract these strings? Not sure what the plan is @heytherewill ?

daividssilverio:
> I have not handled changing the labels on start/end time labels in case the TE has no duration or is not running yet. For this to be implemented, I'd have to have my changes regarding displaying the correct time on the indicator. I could rebase on that branch and add those features, I just don't know what the expectation on this issue is, since it's just an S.

Oh, the idea was to have that done in this PR, but since I didn't notice that dependency, I think we can be flexible with that.
I can create another issue to handle the stop timer as an `size-XS` that can be handled after #200 is merged.

➿ @heytherewill

timer/src/main/res/layout/fragment_dialog_start_edit.xml
```diff
@@ -36,78 +36,187 @@
         app:layout_constraintEnd_toEndOf="parent"
         app:layout_constraintTop_toTopOf="parent" />
 
-    <com.google.android.material.textfield.TextInputLayout
-        android:id="@+id/time_entry_description_input_layout"
+    <androidx.core.widget.NestedScrollView
```
daividssilverio, line 39: We can simplify a bit this layout by making `NestedScrollView` its root and removing the extra nested `ConstraintLayout`, what do you think?

daividssilverio:
> I had to wrap the bottom line of ☝️ in linear layout because I couldn't [in a reasonable amount of time] make all 3 elements to be centered related to the header with just a flat constraint layout 😢If you know a way to do it, please share :)

As a matter of fact we managed to do it in flat way on the old app. It would look like something like this:

https://github.com/toggl/mobileapp/blob/2fba1cdea3fa55f107f331d83972f3a677e5509b/Toggl.Droid/Resources/layout/EditDurationActivity.axml#L22-L181

Basically, the idea is:
- to have a vertical guideline that divides the screen in two
- each header (start | stop) is centered on its side of the divider (start or end)
- each [`date text` <-> `round icon` <-> `start time`] chain is centered on its side of the divider
- the `stop` label is centered likewise on its side of the divider

It works quite alright.

timer/src/main/res/layout/fragment_dialog_start_edit.xml
pdrj: Not really, then the 'x' button and the handle would also scroll and we don't want that. Am I missing something @daividssilverio?

timer/src/main/res/layout/fragment_dialog_start_edit.xml
pdrj: Yeah... For sure that needs to be done at some point. Extracting is not much work, so I'll just do it.

pdrj:
@daividssilverio @semanticer 
I've re-enabled the animation - I didn't notice it before because it's quite subtle and you have to expand/collapse slowly. It works now and is a nice touch 👍 

I've also removed the nested linear layout as suggested by Daivid.

I've also handled the visibility and state of start/stop labels in different scenarios. This is quite messy TBH, if you've got better ideas - I'm open to suggestions. The commit is https://github.com/toggl/android/pull/223/commits/881507d3016521b1752a57c679a275d235b6474f.

I've added handling for the billable chip too - it now dispatches the action and is checked/unchecked according to state.

pdrj:
Oh, and I've merged my unmerged PR into this so I can use the new fields of EditableTimeEntry.

heytherewill:
I didn't review, but a bunch of string were added; don't forget to add the translations for the languages we support as well :)

pdrj:
I'll add these that can be exported form the old app @heytherewill. There's also one that I came up with as a sort of placeholder - I asked in the unification project channel how the view should actually look in the state that the added string is describing.

timer/src/main/res/layout/fragment_dialog_start_edit.xml
daividssilverio: Oh, I didn't notice that. Damn, forget this request then.

pdrj:
@heytherewill 
We're missing:
- Add a project
- End
- Time entry not started yet [potentially, this one probably won't be used anyway]

Should I create 'translation issues' for the first two?

daividssilverio:
@pdrj 
> I've also handled the visibility and state of start/stop labels in different scenarios. This is quite messy TBH, if you've got better ideas - I'm open to suggestions. The commit is 881507d.

Stop label handling is all good, but what about those "time entry not started yet"? That is not how we handle those cases, right? At least the start time should be set at any point 🤔 even if defaulting to "now".

Although I think that's fine for now, we need to go through designs and specs to decide if we are gonna keep that. ping @heytherewill

pdrj:
Related thread https://toggl.slack.com/archives/CS25QL11P/p1587728792000600

common-android/src/main/java/com/toggl/common/ViewExtensions.kt
```diff
@@ -0,0 +1,13 @@
...
+import android.view.View
+
+fun View.addInterceptingOnClickListener(action: () -> Unit) {
+    this.setOnTouchListener { _, event ->
+        if (event.action == MotionEvent.ACTION_UP) {
+            action()
+            true
+        } else false
+    }
+}
```
heytherewill, line 13: Why do we need this to only happen in touch up? 🤔

common-android/src/main/res/values-fr/strings.xml
```diff
@@ -13,6 +13,10 @@
     <string name="continue_this">Continuer</string>
     <string name="delete">Effacer</string>
     <string name="undo">Annuler</string>
+    <string name="add_tags">Ajoutez les labels</string>
+    <string name="billable">Facturable</string>
+    <string name="start">Début</string>
+    <string name="stop_timer">Stop timer</string>
```
heytherewill: This is in english

common-android/src/main/res/values-pt/strings.xml
```diff
@@ -13,6 +13,11 @@
     <string name="continue_this">Continuar</string>
     <string name="delete">Excluir</string>
     <string name="undo">Desfazer</string>
+    <string name="add_project">Add a project</string>
```
heytherewill, line 16: This is english

common-android/src/main/res/values-ja/strings.xml
```diff
@@ -13,6 +13,10 @@
     <string name="continue_this">続ける</string>
     <string name="delete">削除</string>
     <string name="undo">戻す</string>
+    <string name="add_tags">タグを追加</string>
+    <string name="billable">請求可能</string>
+    <string name="start">開始</string>
+    <string name="stop_timer">Stop timer</string>
```
heytherewill: English

common-android/src/main/res/values/colors.xml
```diff
@@ -19,4 +19,6 @@
   <color name="button_active">#06AAF5</color>
   <color name="button_active_background">#2606AAF5</color>
   <color name="button_inactive">#B5BCC0</color>
+  <color name="clickable_text_primary">#328FFF</color>
+  <color name="clickable_text_secondary">#2196F3</color>
```
heytherewill, line 23: Primary and secondary are enabled and disabled? If so, we should consider renaming it

timer/src/main/java/com/toggl/timer/running/domain/RunningTimeEntryReducer.kt
```diff
@@ -31,7 +33,7 @@ class RunningTimeEntryReducer @Inject constructor(
                 state.mutateWithoutEffects {
                     val entryToOpen = timeEntries.runningTimeEntryOrNull()
                         ?.run(EditableTimeEntry.Companion::fromSingle)
-                        ?: EditableTimeEntry.empty(defaultWorkspaceId())
+                        ?: EditableTimeEntry.withStartTime(defaultWorkspaceId(), timeService.now())
```
heytherewill, line 36: Wait a second, this is not correct. The start time should not be set when we open the card, but rather when the entry is started. I understand that you need the `now` time for displaying, but this is not the correct way of doing it because then the user might get unexpected start times.

common-android/src/main/res/values-fr/strings.xml
pdrj: These English strings are gone in the latest commit

common-android/src/main/java/com/toggl/common/ViewExtensions.kt
pdrj: I usually use this one to mean 'a click'. I know it's not exactly the same as click though :) - would you like me to track how long the press took, what the coordinates were, etc?

common-android/src/main/res/values/colors.xml
pdrj: Nope, the primary is the time label, the secondary is the date label. I named it like so because the time one seems 'bolder' in the designs.
Screenshot 2020-04-27 at 14 07 54

timer/src/main/java/com/toggl/timer/running/domain/RunningTimeEntryReducer.kt
pdrj: I thought that was the decision from https://toggl.slack.com/archives/CS25QL11P/p1587728792000600 Otherwise, we'd have to update the start label each minute?

daividssilverio: Yeah... I think we can start the `EditableTimeEntry` with a null `startTime` and handle displaying "now" from the UI side until the user interacts with the wheel or one of the time pickers. What do you think @heytherewill @pdrj ?

pdrj: Yep that's how it's done now @daividssilverio [we're obviously missing the editing part].

pdrj:
Scrolling issue: #236 
Translations: #235

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -104,7 +109,7 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
         return (super.onCreateDialog(savedInstanceState) as BottomSheetDialog).also { bottomSheetDialog: BottomSheetDialog ->
             store.state
                 .filter { it.editableTimeEntry != null }
-                .map { BottomControlPanelParams.fromState(it) }
+                .map { BottomControlPanelParams(it.editableTimeEntry!!, it.isEditableInProWorkspace()) }
```
daividssilverio, line 112: I wish the compiler could be smarter here... those `!!` make me anxious.

heytherewill: Well, what we should be doing here is throwing the appropriate exception and letting the compiler infer that it can't be null 🤔

pdrj: You mean in `map` @heytherewill ? Like 
```
                .map {
                    it.editableTimeEntry ?: throw EditableTimeEntryShouldNotBeNullException()
                    BottomControlPanelParams(it.editableTimeEntry, it.isEditableInProWorkspace()) 
                }
```

pdrj: That would let us omit the !!, but I think @daividssilverio means that the compiler should already know that it can't be null, thanks to the filter.

pdrj:
I agree @daividssilverio, applied the suggestion.

PR #225, opened 2020-04-24
Add Effect to calculate autocomplete suggestions

### LINKED ISSUE #184: Add the UpdateAutocompleteSuggestionsEffect
## Description
This is the effect that will update the suggestions in the background as the user types.
It should receives the current description, the current cursor position, and all timeEntries, projects, tasks, clients and tags and return the list of AutocompleteSuggestion
## Dependencies
Depends on #183 
## Definition of done
- [ ] The effect is added
- [ ] For now, it only returns an empty list of AutocompleteSuggestions

heytherewill:
### Summary
This adds the effect needed to update our autocomplete suggestions

### Technical considerations
This Is mostly boilerplate, but it's better done in its own issue because it touches too many parts of the codebase

### Relationships
Closes #184

### Pull Request checklist

### Pointers
Please go commit by commit, as that will make things simpler to wrap your head around

### Review template

## Merge permissions
Anyone, anytime


timer/src/main/java/com/toggl/timer/startedit/domain/UpdateAutocompleteSuggestionsEffect.kt
```diff
@@ -0,0 +1,26 @@
...
+    private val query: String,
+    private val cursorPosition: Int,
+    private val tags: Map<Long, Tag>,
+    private val tasks: Map<Long, Task>,
+    private val clients: Map<Long, Client>,
+    private val projects: Map<Long, Project>,
+    private val timeEntries: Map<Long, TimeEntry>
+) : Effect<StartEditAction.AutocompleteSuggestionsUpdated> {
+    override suspend fun execute(): StartEditAction.AutocompleteSuggestionsUpdated? =
+        withContext(dispatcherProvider.io) {
```
daividssilverio, line 23: This is more likely to be `default` right? or is it gonna do any IO?

heytherewill: No IO at all, but we wanna do this in another thread, so yeah, good catch

PR #228, opened 2020-04-27
✨ Add DateTimePickMode enum

### LINKED ISSUE #208: Create DateTimePickMode enum and add it as a property to handle editing time entry start/end times to the StartEdit state
## Description
We need a new property added to the StartEdit state to help handling time/date/barrel pickers.
- `datetimePickMode ` should be an enum (DateTimePickMode) with the following possible values:
  - None -> indicates that we are not any picking time/date
  - StartTime -> for the start time picker
  - StartDate -> for the start date picker
  - EndTime -> for the end time picker
  - EndDate -> for the end date picker

## Dependencies
None.

## Definition of done
- [ ] The enum `DateTimePickMode` for the `datetimePickMode` is created
- [ ] The `datetimePickMode` property is added to the StartEdit state

pdrj:
### Summary
This adds DateTimePickMode enum to start edit state

I followed the suggestions from the issue.

### Relationships

Closes #208 

### Pull Request checklist

### Pointers
Well there's not much to test here. Validate the app builds and works fine.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone

### DIFF
--- a/timer/src/main/java/com/toggl/timer/startedit/domain/DateTimePickMode.kt
+++ b/timer/src/main/java/com/toggl/timer/startedit/domain/DateTimePickMode.kt
@@ -1,0 +1,5 @@
+package com.toggl.timer.startedit.domain
+
+enum class DateTimePickMode {
+    None, StartTime, EndTime, StartDate, EndDate
+}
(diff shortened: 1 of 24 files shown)


timer/src/main/java/com/toggl/timer/startedit/domain/DateTimePickMode.kt
```diff
@@ -0,0 +1,5 @@
+package com.toggl.timer.startedit.domain
+
+enum class DateTimePickMode {
+    None, StartTime, EndTime, StartDate, EndDate
```
heytherewill, line 4: I'd prefer these to be one per line for readability

PR #229, opened 2020-04-27
✨ Add and handle PrivateProjectSwitchTapped action

### LINKED ISSUE #185: Add the PrivateProjectSwitchTapped action
## Description
This action is emitted when the user taps the row in the project view that controls whether the project is private or not.

## Dependencies
Depends on #170 

## Definition of done
- [ ] The Action is added
- [ ] The action is dispatched when you tap the row in the project view
- [ ] When the action is handled, it toggles the property
- [ ] The UI should reflect the change
- [ ] Tests ensure behavior is correct

pdrj:
### Summary
This adds and handles PrivateProjectSwitchTapped action

In case anyone is wondering, we can't use onclick because then the chip state is not driven from state ;).
### Relationships

Closes #185 

### Pull Request checklist

### Review template
Apply the following patch:
```

diff --git a/timer/src/main/java/com/toggl/timer/common/domain/EditableTimeEntry.kt b/timer/src/main/java/com/toggl/timer/common/domain/EditableTimeEntry.kt
index edae3d2..8242342 100644
--- a/timer/src/main/java/com/toggl/timer/common/domain/EditableTimeEntry.kt
+++ b/timer/src/main/java/com/toggl/timer/common/domain/EditableTimeEntry.kt
@@ -12,7 +12,7 @@ data class EditableTimeEntry(
     val editableProject: EditableProject?
 ) {
     companion object {
-        fun empty(workspaceId: Long) = EditableTimeEntry(listOf(), workspaceId, "", false, null)
+        fun empty(workspaceId: Long) = EditableTimeEntry(listOf(), workspaceId, "", false, EditableProject.empty(1))
 
         fun fromSingle(timeEntry: TimeEntry) =
             EditableTimeEntry(
diff --git a/timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt b/timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
index e9baa95..b3a0100 100644
--- a/timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
+++ b/timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
@@ -37,6 +37,7 @@ import com.toggl.models.domain.WorkspaceFeature
 import com.toggl.timer.R
 import com.toggl.timer.common.domain.EditableTimeEntry
 import com.toggl.timer.di.TimerComponentProvider
+import com.toggl.timer.project.ui.ProjectDialogFragment
 import com.toggl.timer.startedit.domain.StartEditAction
 import com.toggl.timer.startedit.domain.StartEditState
 import kotlinx.android.synthetic.main.bottom_control_panel_layout.*
@@ -156,6 +157,9 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
                 }
             }
             .launchIn(lifecycleScope)
+
+        ProjectDialogFragment().show(parentFragmentManager, "test")
+
     }
 
     override fun onDestroyView() {
```

and then tap the bottom bar. You should get the project fragment to open. Try using the switch.

- [ ] The code considers i18n

## Merge permissions
Anyone


pdrj:
In https://github.com/toggl/android/pull/223/commits/d2581b24c8f4f1f1468b6373d771567eededea23 I introduced an extension which I'd like to use here as well. When that PR gets merged I'll modify this one to use the extension.

timer/src/test/java/com/toggl/timer/project/domain/PrivateProjectSwitchTappedActionTests.kt
```diff
@@ -0,0 +1,31 @@
...
+import org.junit.jupiter.api.Test
+
+@ExperimentalCoroutinesApi
+@DisplayName("The PrivateProjectSwitchTapped action")
+internal class PrivateProjectSwitchTappedActionTests : CoroutineTest() {
+    val reducer = ProjectReducer(mockk(), dispatcherProvider)
+
+    @Test
+    fun `should toggle the private flag and return no effect`() = runBlockingTest {
+        val initialState = ProjectState(EditableProject.empty(1), emptyMap())
```
daividssilverio, line 21: It would read better if you made explicit that the project is private. This could be enough:

```suggestion
        val initialState = ProjectState(EditableProject.empty(1).copy(isPrivate = true), emptyMap())
```

daividssilverio:
> In case anyone is wondering, we can't use onclick because then the chip state is not driven from state ;).

I've never even considered this 🤔 I'd just use a `setOnCheckedChangeListener`, double check if the state in the `chip` matched and dispatch the action.
It would look like this:
```kotlin
private_chip.setOnCheckedChangeListener { buttonView, isChecked ->  
            if (buttonView.isChecked != isChecked) {
                store.dispatch(ProjectAction.PrivateProjectSwitchTapped)
            }
        }
```

But, the more I look at your solution, the more it makes sense.

pdrj:
Wouldn't that have the possibility of a mismatch between the state and the UI @daividssilverio? Do I get this correctly that you'd dispatch when the UI already presents the 'changed' state?

daividssilverio:
Yeah, that's why your solution keeps looking better @pdrj . What you said is correct, using the `setOnCheckedChangeListener` would mean that the UI is optimistic.

timer/src/main/java/com/toggl/timer/project/ui/ProjectDialogFragment.kt
```diff
@@ -53,6 +61,21 @@ class ProjectDialogFragment : BottomSheetDialogFragment() {
             store.dispatch(action)
         }
 
+        private_chip.setOnTouchListener { _, event ->
+            if (event.action == MotionEvent.ACTION_UP) {
+                store.dispatch(ProjectAction.PrivateProjectSwitchTapped)
+                true
+            } else false
+        }
```
heytherewill, line 69: Change this to use the new extension method

timer/src/main/java/com/toggl/timer/project/ui/ProjectDialogFragment.kt
```diff
@@ -53,6 +61,21 @@ class ProjectDialogFragment : BottomSheetDialogFragment() {
...
+
+        store.state
+            .mapNotNull { it.editableProject?.isPrivate }
+            .distinctUntilChanged()
+            .onEach {
+                private_chip.isChecked = it
+            }
```
heytherewill, line 76: This could be a single line for easier scanning

PR #230, opened 2020-04-27
Dispatch cursor position changes

### LINKED ISSUE #183: Dispatch the DescriptionEntered action when the cursor position changes
## Description
We need to dispatch this action whenever the cursor position changes, since that might change the current autocomplete suggestions
Looking into our Xamarin implementation is a good way of getting insight on all the problems that might arise from doing this manipulation, so I do recommend checking that battle hardened implementation instead of trying to come up with a naive solution that possible ignores corner cases. Needless to say, don't just blindly copy the Xamarin solution.
## Dependencies
Depends on #180 
## Definition of done
- [ ] The action is dispatched whenever the cursor position changes

heytherewill:
### Summary
This makes the description changed action be dispatched when the user changes the position of the cursor

- There's no reliable way to listen to cursor position changes without using a sub-class
- The `requestFocus` method removes an ugly lambda from the activity
- We need the three constructors there because using `@JvmOverloads` (which creates a constructor that calls `super(context, null, 0)`) is not the same as calling `super(context)` for `EditText`s

### Relationships
Closes #183 

### Pull Request checklist

### Pointers
Type `LoggingReducer` in logcat and see the magic 

### Review template

## Merge permissions
Anyone


timer/src/main/java/com/toggl/timer/startedit/ui/autocomplete/AutocompleteEditText.kt
```diff
@@ -0,0 +1,39 @@
...
+
+    fun requestFocus(onFocusChanged: () -> Unit) {
+        setOnFocusChangeListener { _, _ ->
+            post { onFocusChanged() }
+            onFocusChangeListener = null
+        }
+        requestFocus()
+    }
+
+    fun setDescriptionChangedListeners(dispatch: () -> Unit) {
```
pdrj, line 25: Why the plural? It's always just one listener, right?

heytherewill: You set one for when the text changes and one for when the cursor changes

pdrj: OK, this is still the same listener though 😄 It just gets called in two places. I'm good with having that extra 's' in there then.

pdrj:
> Besides the cosmetic comment, I've noticed that the listener would fire 3 times when you enter the edit fragment for a time entry that has description:
> Screenshot 2020-04-27 at 12 58 28
> 
> Are we fine with that?

What about this one @heytherewill?

timer/src/main/java/com/toggl/timer/startedit/ui/autocomplete/AutocompleteEditText.kt
```diff
@@ -5,14 +5,30 @@ import android.text.TextWatcher
...
+@FlowPreview
+@ExperimentalCoroutinesApi
 class AutocompleteTextInputEditText : TextInputEditText {
     constructor(context: Context) : super(context)
     constructor(context: Context, attrs: AttributeSet) : super(context, attrs)
     constructor(context: Context, attrs: AttributeSet, defStyleAttr: Int) : super(context, attrs, defStyleAttr)
 
-    private var dispatch: (() -> Unit)? = null
     private var descriptionChangeListener: TextWatcher? = null
+    // HACK: onSelectionChanged runs before the constructor, so this needs to be nullable
```
pdrj, line 22: :wtf:

heytherewill: I cry everytime 😭

PR #226, opened 2020-04-25
Add the WheelForegroundView

daividssilverio:
### Summary
This is the port for the WheelForegroundView from the old android app: https://github.com/toggl/mobileapp/blob/develop/Toggl.Droid/Views/EditDuration/WheelForegroundView.cs

### Technical considerations
~Still WIP, I intendo to add some niceties like it was done in #221~ 
I've changed some static fields to become styleables to simplify stylizing the wheel.
I've added default colors and dimensions for the view inside the timer module.
I guess we can use proper colors considering dark mode when styling the whole thing on #207 

### Why is it done like this and not some other way?
Feel free to make suggestions to improve how it's done.

### Relationships
Closes #205 

### Pull Request checklist
- [x] The code does not hinder the app's performance
- [x] The code does not introduce regressions
- [x] The code in this PR is tested thoroughly enough that changing the behavior to an incorrect one breaks tests

### Pointers
Please shout if you are seeing something that can be improved.

You can replace the `extended_options`'s `TextView` in the `timer/src/main/res/layout/fragment_dialog_start_edit.xml` layout file with the snipped below to preview the wheel background 

```xml
 <com.toggl.timer.startedit.ui.editduration.WheelForegroundView
        android:id="@+id/extended_options"
        android:layout_width="300dp"
        android:layout_height="300dp"
        app:layout_constraintBottom_toBottomOf="parent"
        app:layout_constraintEnd_toEndOf="parent"
        app:layout_constraintStart_toStartOf="parent"
        app:layout_constraintTop_toBottomOf="@+id/bottom_divider" />
```

Unfortunately, I couldn't make the foreground wheel IDE preview work like the background view. Not sure if it's worth trying for now.

For now, the bottom sheet dialog is intercepting the touches from the wheel during its interaction, so we'll need to add some extra handling there, but we can handle that on #207.

### Review template
- [ ] The code is thoroughly tested in well written and resilient manner
- [ ] The code is performant and leads to an optimal user experience
- [ ] The code is testable and does not cut corners for the sake of delivering
- [ ] The code considers older platforms well and will work in older devices just like it will in new ones
- [ ] The code considers light and dark theme
- [ ] The code considers RTL and does not mention left/right
- [ ] The code considers i18n
  - [ ] There are no hardcoded strings in the UI
  - [ ] All strings are included in every language we support OR there's an issue requesting new translations
- [ ] There are no hard-coded numbers referring to dimensions or font sizes
[...]

PR #234, opened 2020-04-28
Add the WheelDurationInputView

### LINKED ISSUE #206: Create the WheelDurationInputView
## Description
We need to port the wheel duration input to kotlin to compose the Wheel™.
The duration input view:
- Draws a duration in "edit mode" and "display mode"
- Handles touch interaction to let the user manually input minutes
- Provides interfaces to control the widget and its represented value

## Dependencies
None.

## Definition of done
Drawing:
- [ ] The input duration view draws the duration set to when in display mode
- [ ] The input duration view draws the current duration in minutes when the user is inputting values

Input handling:
- [ ] It's possible obtain focus when clicking on the input view
- [ ] It's only possible to input digits and they work like the[ current implementation](https://github.com/toggl/mobileapp/blob/develop/Toggl.Droid/Views/EditDuration/WheelDurationInput.cs).

daividssilverio:
## Info
~WIP~ branched off #226

### Summary
This is the port of old apps's [WheelDurationInput view](https://github.com/toggl/mobileapp/blob/develop/Toggl.Droid/Views/EditDuration/WheelDurationInput.cs)

### Technical considerations
This port also includes the port of the `DurationFieldInfo` struct and its tests, which is actually handling the input.

Feel free to make suggestions. Start reviewing already.

### Relationships
Closes #206 

### Pull Request checklist

## Reviewing
~The first commit on this branch actually is https://github.com/toggl/android/pull/234/commits/b145961091c4ec65466002f229e74641eb563747~

### Pointers
I'll add more reviewing hints after the UI part is done. For now, check the tests.
You can check how everything works by applying this diff:

```
diff --git a/timer/src/main/java/com/toggl/timer/startedit/ui/editduration/WheelDurationInputView.kt b/timer/src/main/java/com/toggl/timer/startedit/ui/editduration/WheelDurationInputView.kt
index 922cc66..24ed48d 100644
--- a/timer/src/main/java/com/toggl/timer/startedit/ui/editduration/WheelDurationInputView.kt
+++ b/timer/src/main/java/com/toggl/timer/startedit/ui/editduration/WheelDurationInputView.kt
@@ -62,6 +62,7 @@ class WheelDurationInputView @JvmOverloads constructor(
         setOnTouchListener(this)
         transformationMethod = null
         filters = arrayOf(WheelDurationInputFilter(::onDigitEntered, ::onDeletionDetected))
+        setDuration(duration)
     }
 
     override fun onEditorAction(actionCode: Int) {
diff --git a/timer/src/main/res/layout/fragment_time_entries_log.xml b/timer/src/main/res/layout/fragment_time_entries_log.xml
index a44b763..db1d533 100644
--- a/timer/src/main/res/layout/fragment_time_entries_log.xml
+++ b/timer/src/main/res/layout/fragment_time_entries_log.xml
@@ -7,6 +7,7 @@
     android:layout_height="match_parent">
 
     <androidx.recyclerview.widget.RecyclerView
+        android:visibility="gone"
         android:id="@+id/recycler_view"
         android:layout_height="match_parent"
@@ -15,6 +16,30 @@
         app:layout_behavior="@string/appbar_scrolling_view_behavior"
         tools:listitem="@layout/time_entries_log_item" />
 
+    <com.toggl.timer.startedit.ui.editduration.WheelBackgroundView
+        android:id="@+id/WheelBackground"
+        android:layout_width="296dp"
+        android:layout_height="296dp"
+        android:layout_gravity="center" />
+
+    <com.toggl.timer.startedit.ui.editduration.WheelForegroundView
+        android:id="@+id/WheelForeground"
+        android:layout_width="296dp"
+        android:layout_height="296dp"
+        android:layout_gravity="center" />
+
+    <com.toggl.timer.startedit.ui.editduration.WheelDurationInputView
+        android:id="@+id/WheelDurationInput"
+        android:textSize="28dp"
+        android:background="@null"
+        android:textColor="#328fff"
+        app:fadedTextColor="#cecece"
+        android:inputType="numberPassword"
+        android:fontFamily="sans-serif-light"
+        android:layout_gravity="center"
+        android:layout_width="wrap_content"
+        android:layout_height="wrap_content" />
+
     <com.google.android.material.appbar.AppBarLayout
         android:id="@+id/app_bar_layout"

```

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it only after I rebase this after #226 PR is merged.


daividssilverio:
> The input duration view draws the current duration in minutes when the user is inputting values

Why not? You can see the duration while typing, right?
When on focus:
image

When out of focus:
image

daividssilverio:
> Also, I'm wondering why is the wheel foreground and background split? They're always the same dimension and having them next to each other in layout files is quite weird. Shouldn't they be merged and the background painted automatically?

For two reasons @pdrj 
1. It would mean bloating the foreground view with the logic to draw the clock dial.
2. It would also mean redrawing the background view on every invalidation in the foreground view, and that's not ideal. @kornelijepetak once said that there must be a smarter way to do that as well to that it background view is drawn only once using the same canvas, but I didn't find the time to update it and went with what we have right now. Do you know the way to do what Kornelije suggested?

Here are some of the `onDraw` calls made when you interact with the wheel.
```
2020-04-29 10:06:17.277 21782-21782/com.toggl.aurora.debug D/xxaa: background onDraw
2020-04-29 10:06:17.284 21782-21782/com.toggl.aurora.debug D/xxaa: foreground onDraw
2020-04-29 10:07:12.964 21782-21782/com.toggl.aurora.debug D/xxaa: foreground onDraw
2020-04-29 10:07:13.894 21782-21782/com.toggl.aurora.debug D/xxaa: foreground onDraw
2020-04-29 10:07:13.918 21782-21782/com.toggl.aurora.debug D/xxaa: foreground onDraw
2020-04-29 10:07:14.871 21782-21782/com.toggl.aurora.debug D/xxaa: foreground onDraw
2020-04-29 10:07:14.995 21782-21782/com.toggl.aurora.debug D/xxaa: foreground onDraw
2020-04-29 10:07:15.769 21782-21782/com.toggl.aurora.debug D/xxaa: foreground onDraw
2020-04-29 10:07:15.787 21782-21782/com.toggl.aurora.debug D/xxaa: foreground onDraw
2020-04-29 10:07:15.912 21782-21782/com.toggl.aurora.debug D/xxaa: foreground onDraw
2020-04-29 10:07:15.931 21782-21782/com.toggl.aurora.debug D/xxaa: foreground onDraw
2020-04-29 10:07:16.924 21782-21782/com.toggl.aurora.debug D/xxaa: foreground onDraw
2020-04-29 10:07:16.979 21782-21782/com.toggl.aurora.debug D/xxaa: foreground onDraw
2020-04-29 10:07:17.750 21782-21782/com.toggl.aurora.debug D/xxaa: foreground onDraw
```

pdrj:
> Why not? You can see the duration while typing, right?

It was unclear to me if that means that the indicators on the wheel should update too. But I guess they should not update yet.

Regarding the bg view. I think this could be achieved by having a delegate to handle the bg drawing and only call it when needed? That being said, I'm sure it's not just as straightforward as this and requires some thinking and testing, so we should leave it for now.

common/src/main/java/com/toggl/common/ComparableExtensions.kt
```diff
@@ -0,0 +1,7 @@
+package com.toggl.common
+
+fun <T : Comparable<T>> T.clamp(min: T, max: T): T {
+    if (this < min) return min
+    if (this > max) return max
+    return this
+}
```
heytherewill, line 7: I thought we had deleted this

timer/src/main/java/com/toggl/timer/startedit/ui/editduration/WheelDurationInputView.kt
```diff
@@ -0,0 +1,200 @@
...
+import kotlinx.coroutines.flow.asFlow
+import org.threeten.bp.Duration
+
+@ExperimentalCoroutinesApi
+@FlowPreview
+class WheelDurationInputView @JvmOverloads constructor(
+    context: Context,
+    attrs: AttributeSet? = null,
+    defStyleAttr: Int = 0
+) : AppCompatEditText(context, attrs, defStyleAttr), TextWatcher, View.OnTouchListener {
```
heytherewill, line 35: Don't use `@JvmOverloads` constructors with `EditText`s, since all hell might break loose (like it did in #230). Explicitly define 3 constructors instead

common/src/main/java/com/toggl/common/ComparableExtensions.kt
daividssilverio: This branch needs a rebase after the #226 is merged

timer/src/main/java/com/toggl/timer/startedit/ui/editduration/WheelDurationInputView.kt
daividssilverio: I rather not, are you ok with using the correct default value for the `defStyleAttr` instead? I've done that here https://github.com/toggl/android/pull/234/commits/cb66be57fbcf8d7b89499b75e9418430f92ce4b5

daividssilverio: Sorry for not explaining why. I rather not do that because the initialization code becomes more cluttered, that's it :)

PR #237, opened 2020-04-29
🛠 Add file template for creating new feature

### LINKED ISSUE #222: Create feature template files
## Description
Android version of https://github.com/toggl/ios/issues/86

This template should create the whole feature structure based on the given `FeatureName` right after selecting this template in Android Studio. 

Structure of newly created feature files

    .
    ├── ...
    ├── domain
    │   ├── FeatureState.kt             
    │   ├── FeatureReducer.kt              
    │   ├── FeatureAction.kt             
    ├── ui
    │   ├── FeatureStoreViewModel.kt        
    │   ├── FeatureFragment.kt

Ideally, we should also create `fragment_feature.xml` resource for our layout. 

The above files should have the basic implementation of the boilerplate code (based on features already created in the project). The generated boilerplate code should also include the needed imports. It should also include the given `FeatureName` in the generated code where needed.

## Dependencies
none

## Definition of done
- [ ] Create all the files as described above
- [ ] Put a copy of the working templates in the folder `Templates` in the root dir of the repo

semanticer:
## Info
This adds a file template for easier feature creation as described in #222. 

### Summary
Screen Shot 2020-04-29 at 10 33 56
Screen Shot 2020-04-29 at 10 34 13
Screen Shot 2020-04-29 at 10 34 24

### Technical considerations
I used FreeMarker, the same technique that is used for Android Templates created by Google.

This is probably the only way but please do let me know if you have some alternative solutions

### Relationships

Closes #222

### Pull Request checklist

### Pointers
Follow the instructions from the README file and see if you can successfully copy this template to your AS. Try creating a new feature and review created code.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


.idea/templates/file/aurora/Feature/root/src/app_package/Fragment.kt.ftl
```diff
@@ -0,0 +1,42 @@
...
+import android.view.View
+import androidx.fragment.app.Fragment
+import androidx.fragment.app.viewModels
+import androidx.lifecycle.ViewModelProvider
+import androidx.lifecycle.lifecycleScope
+import ${packageName}.domain.${actionName}
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.flow.launchIn
+import kotlinx.coroutines.flow.onEach
+import javax.inject.Inject
```
daividssilverio, line 14: You could also add the R import as well: `import ${packageName}.R`

semanticer: The problem with this is when you have a feature that is not in the root of the module. Like all the features in our Time Module right now.  Then your `${packageName}` is actually different than the root package from which you need the `R`.

semanticer: We could probably create two different feature templates or add boolean setting to cover both of these cases. The root module feature could also generate Dagger files maybe 🤔
I'm just not sure it is worth the effort.

semanticer:
I haven't noticed any `.DS_Store` but I added them to `gitignore` just to be safe. 

Some script that would update both **file** and **live** templates would be awesome. I am actually pretty disappointed that there is no better way provided by **Intellij** :-(

daividssilverio:
@semanticer I was actually suggesting you to write a script like that 🤔. But we can have that later.

semanticer:
I'm not 100% sure how to approach this script thing to make it cross-platform etc. But I will definitely think about it.

PR #238, opened 2020-04-29
✨ Add and dispatch PickerTapped action

### LINKED ISSUE #209: Add the PickerTapped action to the StartEdit reducer
## Description
- It should take in a `DateTimePickMode` to tell if we are changing the (start|end) (date|time) values.
- It should be reduced to change the state to set the `datetimePickMode`.

## Dependencies
It needs the enum defined for `datetimePickMode` on #208
Depends on #202 

## Definition of done
- [x] The action is added
- [x] The action is dispatched when you tap the labels for the `start time`, `start date`, `end time` and `end date`
- [x] The action is reduced, setting the `datetimePickMode` to the state
- [x] No effects are produced
- [x] Tests were added
- [ ] The [documentation](https://www.notion.so/toggl/PickerTapped-40a7f5fed38c4908bbb3111acae715d4) is updated to reflect the changes done in this PR

pdrj:
### Summary
This adds and dispatches PickerTapped action

This seems simple enough.

### Relationships

Closes #209 

### Pull Request checklist

### Pointers
Try to tap start/end labels for TEs that are stopped/ongoing/not started and observe the output of LoggingReducer.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone

### DIFF
--- a/timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
+++ b/timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
@@ -200,12 +200,7 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
             stop_time_label to DateTimePickMode.EndTime,
             start_date_label to DateTimePickMode.StartDate,
             stop_date_label to DateTimePickMode.EndDate
-        ).onEach {
-            val (label, action) = it
-            label.setOnClickListener {
-                if (isStartStopLabelClickable(label)) store.dispatch(StartEditAction.PickerTapped(action))
-            }
-        }
+        ).onEach { it.setActionOnLabel() }
 
         val bottomSheetBehavior = (dialog as BottomSheetDialog).behavior
         with(bottomSheetBehavior) {
@@ -371,6 +366,16 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
         if (this.text != newText) {
             this.text = newText
         }
+    }
+
+    private fun Map.Entry<TextView, DateTimePickMode>.setActionOnLabel() {
+        val (label, action) = this
+        label.setOnClickListener {
+            if (isStartStopLabelClickable(label)) {
+                store.dispatch(StartEditAction.PickerTapped(action))
+            }
+        }
+        return
     }
 
     private data class BottomControlPanelParams(val editableTimeEntry: EditableTimeEntry, val isProWorkspace: Boolean)
--- a/timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
+++ b/timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
@@ -321,7 +321,7 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
                 stop_time_label to DateTimePickMode.EndTime,
                 start_date_label to DateTimePickMode.StartDate,
                 stop_date_label to DateTimePickMode.EndDate
-            ).onEach { it.setActionOnLabel() }
+            ).onEach { it.setPickerTappedActionOnLabel() }
 
             setTextOnStartTimeLabels(startTime)
 
@@ -364,7 +364,7 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
         }
     }
 
-    private fun Map.Entry<TextView, DateTimePickMode>.setActionOnLabel() {
+    private fun Map.Entry<TextView, DateTimePickMode>.setPickerTappedActionOnLabel() {
         val (label, action) = this
         label.setOnClickListener {
             store.dispatch(StartEditAction.PickerTapped(action))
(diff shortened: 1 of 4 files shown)


timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -190,6 +195,18 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
...
+            start_time_label to DateTimePickMode.StartTime,
+            stop_time_label to DateTimePickMode.EndTime,
+            start_date_label to DateTimePickMode.StartDate,
+            stop_date_label to DateTimePickMode.EndDate
+        ).onEach {
+            val (label, action) = it
+            label.setOnClickListener {
+                if (isStartStopLabelClickable(label)) store.dispatch(StartEditAction.PickerTapped(action))
+            }
+        }
```
heytherewill, line 208: I'm not a fan of using `forEach`/`onEach` and then passing a multi-line lambda. This is useful when we have a reusable function; otherwise a simple for loop would be more readable

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -190,6 +195,18 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
...
 
+        mapOf(
+            start_time_label to DateTimePickMode.StartTime,
+            stop_time_label to DateTimePickMode.EndTime,
+            start_date_label to DateTimePickMode.StartDate,
+            stop_date_label to DateTimePickMode.EndDate
+        ).onEach {
+            val (label, action) = it
+            label.setOnClickListener {
+                if (isStartStopLabelClickable(label)) store.dispatch(StartEditAction.PickerTapped(action))
```
heytherewill, line 206: Ifs that don't return anything should really have braces so we can scan them faster when reading

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -329,10 +346,17 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
         }
     }
 
+    private fun isStartStopLabelClickable(label: View) = when (label) {
+        in startLabels -> startLabels.all { it.isVisible }
```
heytherewill, line 350: Why do you need to verify the visibility of all view in the group? 🤔

pdrj: Because only when both parts are visible, either of them is clickable [in the context of this PR]. Otherwise we're dealing with the other cases, when only one label is visible and it functions differently then - for instance for stopping a running TE.

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -373,5 +368,15 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
...
     }
 
+    private fun Map.Entry<TextView, DateTimePickMode>.setActionOnLabel() {
+        val (label, action) = this
+        label.setOnClickListener {
+            if (isStartStopLabelClickable(label)) {
+                store.dispatch(StartEditAction.PickerTapped(action))
+            }
+        }
+        return
```
heytherewill: ?

heytherewill: PLease remove this if it's not needed

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -349,5 +364,12 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
         }
     }
 
+    private fun Map.Entry<TextView, DateTimePickMode>.setActionOnLabel() {
```
daividssilverio, line 367: A more descriptive name could be `setPickerTappedActionOnLabel()`

timer/src/test/java/com/toggl/timer/startedit/domain/PickerTappedActionTests.kt
```diff
@@ -0,0 +1,43 @@
...
+@DisplayName("The PickerTapped action")
+internal class PickerTappedActionTests {
+    val testDispatcher = TestCoroutineDispatcher()
+    val dispatcherProvider = DispatcherProvider(testDispatcher, testDispatcher, Dispatchers.Main)
+    val repository = mockk<TimeEntryRepository>()
+    val initialState = createInitialState()
+    val reducer = StartEditReducer(repository, dispatcherProvider)
+
+    @ParameterizedTest
+    @MethodSource("allPickerModes")
```
daividssilverio, line 28: For enums you can use `@EnumSource(DateTimePickMode::class)` and 🔥 the companion object.

pdrj: Oh nice, thanks.

PR #240, opened 2020-04-29
✨🚧 Reimplement Main log to use Epoxy

### LINKED ISSUE #224: 🚧Refactor Main Log using Epoxy library
## Description
We will refactor the Main Log recycler view using [Epoxy](https://github.com/airbnb/epoxy). During this process, we should set up the basic infrastructure for possible future Epoxy usage. We should be able to do this with minimal layout changes and re-use most of our [ViewHolder code](https://github.com/airbnb/epoxy/wiki/ViewHolder-Models).

## Dependencies
None

## Definition of done
- [ ] We have the infrastructure for Epoxy and it's further usage ready
- [ ] Main log is refactored using epoxy with re-used [ViewHolder code](https://github.com/airbnb/epoxy/wiki/ViewHolder-Models)
- [ ] Swipe actions are refactored to use https://github.com/airbnb/epoxy/wiki/Touch-Support
- [ ] Old adapter is removed

semanticer:
### Summary
Reimplementation of main log using https://github.com/airbnb/epoxy

Key takeaways so far:
- No Adapter code
- Super nice **updateList** declarative method

**Swiping is not implemented yet**

### Technical considerations
TODO

TODO

### Relationships
Closes #224

### Pull Request checklist

### Review template

- [ ] The code considers i18n

## Merge permissions
Me


timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogFragment.kt
```diff
@@ -62,95 +71,76 @@ class TimeEntriesLogFragment : Fragment(R.layout.fragment_time_entries_log) {
...
+    private fun onLogItemSwiped(model: EpoxyModel<Any>, swipeDirection: SwipeDirection) {
+        when (model) {
+            is TimeEntryItemModel_ -> TimeEntriesLogAction.TimeEntrySwiped(model.timeEntry.id, swipeDirection)
+            is TimeEntryGroupModel_ -> TimeEntriesLogAction.TimeEntryGroupSwiped(model.timeEntryGroup.timeEntryIds, swipeDirection)
+            else -> throw IllegalStateException("Swiping unsupported model $model")
+        }.let(store::dispatch)
 
         if (swipeDirection == SwipeDirection.Right) {
-            adapter.notifyItemChanged(adapterPosition)
+            // recycler_view.adapter?.notifyItemChanged(position) we need this to revert continue animation but it's it is forbidden to call this in epoxy
```
semanticer, line 138: This is the issue we have with the continue swipe animation not being reverted.

timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogFragment.kt
```diff
@@ -135,8 +140,8 @@ class TimeEntriesLogFragment : Fragment(R.layout.fragment_time_entries_log) {
         }.let(store::dispatch)
 
         if (swipeDirection == SwipeDirection.Right) {
-            // recycler_view.adapter?.notifyItemChanged(position) we need this to revert continue animation but it's it is forbidden to call this in epoxy
-            recycler_view.requestModelBuild()
+            touchHelper?.attachToRecyclerView(null)
```
daividssilverio, line 142: well... it works... but this is so odd...

semanticer: Yeah, and the reverse animation doesn't really play correctly, it just jumps to its original position 😢

semanticer:
I did some more research over the weekend. The main problem is that ItemTouchHelper [is not designed to handle swipe actions that don't remove the item from the list](https://stackoverflow.com/a/31872357/509681) 

The fact that it works with the `adapter.notifyItemChanged(position)` and the view is animated back to its position seems to be just a lucky coincidence/hack. 

We have 4 options:

1. I found some nicer ways to reset the view than `touchHelper?.attachToRecyclerView(null)` but the result is the same = the animation is not played backward, it just jumps to the original position :-( 

2. We find some other way to implement swipe to continue. Not allowing the full swipe and change the threshold so that the view is always returned to its original place. 

3. We will create a custom TouchHelper as [suggested](https://stackoverflow.com/questions/31787272/android-recyclerview-itemtouchhelper-revert-swipe-and-restore-view-holder/31872357#31872357) but that feels 🤮

4. We won't use epoxy, or we will use it for some other RecyclerViews than this one

PR #242, opened 2020-04-29
Add first analytics events

### LINKED ISSUE #241: Add first round of events
## Description
After [going through all events in the old apps](https://www.notion.so/toggl/Old-App-Events-that-need-mapping-5ffa065959f1405a8b5fc7128fb23840) I managed to compile a list of events that can be implemented with the code we already have. We should do so now and then implement future events as they come.

## Definition of done
- [ ] The `EditViewClosed` event is implemented with the same parameters as the Xamarin app
- [ ] The `EditViewOpened` event is implemented with the same parameters as the Xamarin app
- [ ] The `TimeEntryStopped` event is implemented with the same parameters as the Xamarin app
- [ ] The `DeleteTimeEntry` event is implemented with the same parameters as the Xamarin app

heytherewill:
### Summary
This adds the first round of analytics events to get our codebase up to speed. Other events can later be added alongside with actions.

### Relationships
Closes #241

### Pull Request checklist
// I'm not a fan of testing Analytics events. It kinda becomes silly to the point where you copy the analytics code to check that a method was called. For that reason I'm not adding tests for this.

### Pointers
Check commit by commit to reducer clutter. The last commit is the important one.

### Review template
- [x] The code is thoroughly tested in well written and resilient manner
- [x] The code is performant and leads to an optimal user experience
- [x] The code is testable and does not cut corners for the sake of delivering
- [X] The code considers older platforms well and will work in older devices just like it will in new ones

## Merge permissions
Anyone, anytime


environment/src/main/java/com/toggl/environment/services/analytics/parameters/EditViewOpenReason.kt
```diff
@@ -0,0 +1,9 @@
+package com.toggl.environment.services.analytics.parameters
+
+enum class EditViewOpenReason {
+    RunningTimeEntryCard,
+    SingleTimeEntry,
+    GroupHeader,
+    GroupTimeEntry,
```
daividssilverio, line 7: Interesting, we are not tracking this one yet 🤔 We might need to change the code a bit to enable tracking it, right?

heytherewill: Yeah, it requires a couple of changes because we don't differentiate taps

environment/src/main/java/com/toggl/environment/services/analytics/parameters/TimeEntryStopOrigin.kt
```diff
@@ -0,0 +1,11 @@
+package com.toggl.environment.services.analytics.parameters
+
+enum class TimeEntryStopOrigin {
+    Manual,
+    Deeplink,
+    Widget,
+    EditView,
+    Wheel,
+    CalendarContextualMenu,
+    Keyboard
```
daividssilverio, line 10: Not sure if the `Keyboard` entries will ever make sense on the android apps. Not until we start supporting tablets or chromebooks officially 🤔

PR #244, opened 2020-04-29
Add the DateTimePicked action

### LINKED ISSUE #211: Add the DateTimePicked action to the StartEdit reducer
## Description
This action takes in the value obtained from the picker.
This action should be emitted when a datetime is picked from one of the time/date/barrels has a value selected.
The reducer should validate the input from the action to prevent inconsistencies.
The reducer should set the `editableTimeEntry`'s start time/duration accordingly.

*On Android, this action should also set the `datetimePickMode` to `None` after executing successfully.
*On iOS, this action does do anything else.

## Dependencies
Depends on #208 

## Definition of done
- [x] The action is added and takes in a datetime parameter
- [x] The `editableTimeEntry`s start time and duration is updated accordingly based on the current `datetimePickMode`
- [x] On android, the action input needs to be validated to make sure the `start time` and `duration` are valid. An exception should be thrown if the input is invalid. Check the [old source code for reference](https://github.com/toggl/mobileapp/blob/2fba1cdea3fa55f107f331d83972f3a677e5509b/Toggl.Core.UI/ViewModels/EditDurationViewModel.cs#L275-L307).
- [x] The behavior is tested 
- [ ] The [documentation](https://www.notion.so/toggl/DateTimePicked-8c45d89481fc4fd99381771c46090065) is updated to reflect the changes done in this PR

daividssilverio:
### Summary
This PR adds the DateTimePicked action, reduces and tests its behaviour.

### Technical considerations
~Handling the exceptions is looking a bit weird in my head right now, not really sure at the moment how to handle it properly.~
I've had to make some extra changes to the state and the reducer to handle date time picking exceptions:
- A new property has been added to the state to help handle temporal inconsistencies
- A new effect has been added to reopen a date/time picker when a temporal inconsistency is detected

Feel free to make suggestions, specially about handling the temporal inconsistencies.

### Relationships
Closes #211

### Pull Request checklist

### Pointers
~WIP:
For now, please check the tests.~
Check the [old implementation](https://github.com/toggl/mobileapp/blob/2fba1cdea3fa55f107f331d83972f3a677e5509b/Toggl.Core.UI/ViewModels/EditDurationViewModel.cs#L275-L307) for reference.
Try reading the [tests](https://github.com/toggl/android/pull/244/commits/6c9018e553d8141f2d72f9c383483215bbd0158f).

### Review template

## Merge permissions
🦑 it ~after the draft becomes a PR~


timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
```diff
@@ -103,7 +109,108 @@ class StartEditReducer @Inject constructor(
...
+        }
+
+    private fun StartEditState.handleEndTimeEdition(
+        editableTimeEntry: EditableTimeEntry,
+        action: StartEditAction.DateTimePicked
+    ): StartEditState = copy(
+        dateTimePickMode = DateTimePickMode.None,
+        temporalInconsistency = TemporalInconsistency.None,
+        editableTimeEntry = editableTimeEntry.copy(
+            duration = editableTimeEntry.startTime!!.absoluteDurationBetween(action.dateTime)
```
heytherewill, line 146: Let's avoid force unwrapping and use the exceptions to properly throw adequate errors instead

timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
```diff
@@ -103,7 +109,108 @@ class StartEditReducer @Inject constructor(
...
+        return if (originalDuration == null) {
+            copy(
+                dateTimePickMode = DateTimePickMode.None,
+                temporalInconsistency = TemporalInconsistency.None,
+                editableTimeEntry = editableTimeEntry.copy(
+                    startTime = action.dateTime
+                )
+            )
+        } else {
+            val originalStartTime = editableTimeEntry.startTime!!
```
heytherewill, line 164: Let's avoid force unwrapping and use the exceptions to properly throw adequate errors instead

timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
```diff
@@ -103,7 +109,108 @@ class StartEditReducer @Inject constructor(
...
+                } else {
+                    val endTime = startTime + editableTimeEntry.duration
+                    if (newDateTime > endTime)
+                        return TemporalInconsistency.StartTimeAfterStopTime
+                    if (endTime.absoluteDurationBetween(newDateTime) > maxDuration)
+                        return TemporalInconsistency.DurationTooLong
+                }
+            }
+            return TemporalInconsistency.None
+        } else if (mode.targetsEnd()) {
```
heytherewill, line 204: Does this have to be an else if? What else will it target, if it doesn't target neither? 🤔

timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
```diff
@@ -103,7 +109,108 @@ class StartEditReducer @Inject constructor(
...
+                        return TemporalInconsistency.StartTimeAfterCurrentTime
+                } else {
+                    val endTime = startTime + editableTimeEntry.duration
+                    if (newDateTime > endTime)
+                        return TemporalInconsistency.StartTimeAfterStopTime
+                    if (endTime.absoluteDurationBetween(newDateTime) > maxDuration)
+                        return TemporalInconsistency.DurationTooLong
+                }
+            }
+            return TemporalInconsistency.None
```
heytherewill, line 203: There are two `return TemporalInconsitency.None` at the end of elses and I find this a bit confusing, why aren't you just letting the one at the end return None?

timer/src/main/java/com/toggl/timer/startedit/domain/ReopenPickerEffect.kt
```diff
@@ -0,0 +1,7 @@
+package com.toggl.timer.startedit.domain
+
+import com.toggl.architecture.core.Effect
+
+class ReopenPickerEffect(val dateTimePickMode: DateTimePickMode) : Effect<StartEditAction> {
+    override suspend fun execute(): StartEditAction? = StartEditAction.PickerTapped(dateTimePickMode)
+}
```
semanticer, line 7: Why is this "useless" effect needed? You can just extract the behavior from `StartEditAction.PickerTapped ->` into some method and use this method instead of `effect(ReopenPickerEffect(mode))` ? Or am I missing something?

timer/src/main/java/com/toggl/timer/common/domain/TimerState.kt
```diff
@@ -26,14 +27,16 @@ data class TimerState(
         internal val expandedGroupIds: Set<Long>,
         internal val entriesPendingDeletion: Set<Long>,
         internal val autocompleteSuggestions: List<AutocompleteSuggestion>,
-        internal val dateTimePickMode: DateTimePickMode
+        internal val dateTimePickMode: DateTimePickMode,
+        internal val temporalInconsistency: TemporalInconsistency
```
semanticer, line 31: Can you maybe elaborate a bit more on why we need to persist this inconsistency in our state? Is it because of some error reporting in UI or?

timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
semanticer, line 203: This method is very hard to understand. Do you think we can make this a single `when()` like:

```kotlin
val maxDuration = Duration.ofHours(999)
val now = timeService.now()
// TODO other important vals
return when (x) {
    longCondition -> TemporalInconsistency.DurationTooLong
    longCondition -> TemporalInconsistency.StartTimeAfterCurrentTime
    longCondition -> TemporalInconsistency.StartTimeAfterStopTime   
    longCondition -> TemporalInconsistency.StopTimeBeforeStartTime
    else -> TemporalInconsistency.None
}
```

With this approach, conditions could be very long and repetitive but I would prefer the clarity where you can simply go through all the things that need to be true for each individual TemporalInconsistency enum to be returned rather than this deeply nested if/else tree.

timer/src/test/java/com/toggl/timer/common/HelperFunctions.kt
```diff
@@ -35,7 +35,7 @@ fun <T> T.toMutableValue(setFunction: (T) -> Unit) =
 suspend fun <State, Action> Reducer<State, Action>.testReduce(
     initialState: State,
     action: Action,
-    testCase: suspend (State, List<Effect<Action>>) -> Unit
+    testCase: suspend (State, List<Effect<Action>>) -> Unit = { _, _ -> }
```
semanticer, line 38: What about making this even more explicit. Keep the variant with non-optional testCase and rather add another `testReduce` variant with the expected exception as the last argument, instead of the `testCase`.
With this we could replace:

```kotlin
       assertThrows<EditableTimeEntryDoesNotHaveAStartTimeSetException> {
            runBlockingTest {
                reducer.testReduce(
                    initialState,
                    StartEditAction.DateTimePicked(mockk())
                )
            }
        }
```

with: 
```kotlin
         runBlockingTest {
                reducer.testReduce(
                    initialState,
                    StartEditAction.DateTimePicked(mockk()),
                    EditableTimeEntryDoesNotHaveAStartTimeSetException
                )
            }
```

I prefer less nested code and it would be safer to hide this inside of our testing API if we ever decide to switch testing framework again - but it's up to you.

timer/src/main/java/com/toggl/timer/startedit/domain/ReopenPickerEffect.kt
daividssilverio: It sounded very expressive to write it like that.
Sure it can be a method, but it will become harder to test and a bit weird to understand from my point of view.

Also, launching an effect isn't exactly the same as running some statement on the spot, or did I get something wrong? The effect will go to the store queue and might not run immediately, right?

Anyway, what I wanted to express is:
When a temporal inconsistency is detected, we need to show a flash message (toast) about the the problem and quickly reopen the DateTime picker.

timer/src/main/java/com/toggl/timer/common/domain/TimerState.kt
daividssilverio: Exactly that. We need to show a toast or some other kind of error message when an invalid date time is picked.
We need that because our `Time` and `Date` pickers are separated and it's possible to pick invalid start/end `DateTime` from both pickers.

timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
daividssilverio: I can try to improve that. I did try it with a `when` before, but it ended up very convoluted... but let's see if we can improve it now.

timer/src/test/java/com/toggl/timer/common/HelperFunctions.kt
daividssilverio: I like that. Let's do it.

timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
daividssilverio: I'll try to address your comment alongside with @semanticer 's comment about using a `when`.

timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
daividssilverio: Same.

daividssilverio: Oh man... that was getting gross... so I've split the logic into a few `whens` and new functions.
Which led my tests to fail and a question to be asked 🤔Is it legal to set an end time (duration) from a picker? ping @heytherewill 

```
private fun detectTemporalInconsistencies(
        editableTimeEntry: EditableTimeEntry,
        mode: DateTimePickMode,
        newDateTime: OffsetDateTime
    ): TemporalInconsistency {
        val maxDuration = Duration.ofHours(999)
        val now = timeService.now()
        val startTime = editableTimeEntry.startTime
        val duration = editableTimeEntry.duration
        val endTime = (startTime ?: now) + (duration ?: Duration.ZERO)

        return when {
            startTime == null -> detectTemporalInconsistenciesOnEmptyTimeEntry(mode, newDateTime, now, maxDuration)
            duration == null -> detectTemporalInconsistenciesOnRunningTimeEntry(mode, newDateTime, now, maxDuration)
            else -> detectTemporalInconsistenciesOnStoppedTimeEntry(mode, startTime, endTime, newDateTime, maxDuration)
        }
    }

    private fun detectTemporalInconsistenciesOnStoppedTimeEntry(
        mode: DateTimePickMode,
        startTime: OffsetDateTime,
        endTime: OffsetDateTime,
        newDateTime: OffsetDateTime,
        maxDuration: Duration
    ): TemporalInconsistency = when {
        mode.targetsStart() && newDateTime > endTime -> StartTimeAfterStopTime
        mode.targetsStart() && endTime.absoluteDurationBetween(newDateTime) > maxDuration -> DurationTooLong
        mode.targetsEnd() && newDateTime < startTime -> StopTimeBeforeStartTime
        mode.targetsEnd() && newDateTime.absoluteDurationBetween(startTime) > maxDuration -> DurationTooLong
        else -> TemporalInconsistency.None
    }

    private fun detectTemporalInconsistenciesOnRunningTimeEntry(
        mode: DateTimePickMode,
        newDateTime: OffsetDateTime,
        now: OffsetDateTime,
        maxDuration: Duration
    ): TemporalInconsistency = when {
        mode.targetsStart() && now.absoluteDurationBetween(newDateTime) > maxDuration -> DurationTooLong
        mode.targetsStart() && newDateTime > now -> StartTimeAfterCurrentTime
        mode.targetsEnd() -> throw EditableTimeEntryDoesNotHaveAStartTimeSetException()
        else -> TemporalInconsistency.None
    }

    private fun detectTemporalInconsistenciesOnEmptyTimeEntry(
        mode: DateTimePickMode,
        newDateTime: OffsetDateTime,
        now: OffsetDateTime,
        maxDuration: Duration
    ): TemporalInconsistency = when {
        mode.targetsStart() && now.absoluteDurationBetween(newDateTime) > maxDuration -> DurationTooLong
        mode.targetsEnd() -> throw EditableTimeEntryDoesNotHaveAStartTimeSetException()
        else -> TemporalInconsistency.None
    }
```

semanticer: I quite like this solution. Another alternative would be to declare those three helper methods inside of the `detectTemporalInconsistencies` to limit the scope but it kinda looks uglier usually :-)

PR #245, opened 2020-04-30
✨ Show date/time pickers when respective label is clicked

### LINKED ISSUE #217: Handle displaying date/time pickers for the edit duration view
## Description
We need to display the date & time pickers when the StartEdit's state say so.

## Dependencies
Depends on #202 
Depends on #208 
Depends on #210 

## Definition of done
- [ ] A time picker is displayed when the `StartEdit`'s state `datetimePickMode` is set to `StartTime` or `EndTime`.
- [ ] A date picker is displayed when the `StartEdit`'s state `datetimePickMode` is set to `StartDate ` or `EndDate`.
- [ ] The date picker has its max and min dates set based on the `editableTimeEntry`'s start time and end time (start time + duration).
- [ ] The preset date/time is set based on the `datetimePickMode` and the `editableTimeEntry`'s `startTime` and `duration`.
- [ ] No picker should be visible when `datetimePickMode` is set to `None`.
- [ ] The `DateTimePickingCancelled` action is dispatched when the picker's cancel button is clicked.

pdrj:
### Summary
This shows date/time pickers when respective label is clicked. It also dispatches picking cancelled event when the dialog is cancelled.

### Technical considerations
There are many edge cases with this, especially when setting bounds on the min/max date 😐 It works like this:
- for stopped time entries, it works as you'd suppose, I guess 😄 
- for not-yet-started entries, ~~it lets you set the start date forward/backward however you want~~ it lets you only set the start date to a past one - there's an ongoing discussion on this subject https://toggl.slack.com/archives/C02APMYTM/p1588250620470100 we'll need some more work on this in any case
- for running time entries, it lets you set the start date to the past only

I don't like the naming I used `..orNow` - if you can think of something better, let me know.

### Relationships

Closes #217 

### Pull Request checklist

### Pointers
Click on the start/end time/date labels for a running/stopped/not-started TE and see if the initial values are set correctly. For date pickers, validate the max/min is set correctly.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -280,6 +296,58 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
...
+            DatePickerDialog(requireContext(), null, initialTime.year, initialTime.monthValue - 1, initialTime.dayOfMonth)
+        val datePicker = datePickerDialog.datePicker
+
+        maxTime?.let { datePicker.maxDate = maxTime.toEpochMillisecond() }
+        minTime?.let { datePicker.minDate = minTime.toEpochMillisecond() }
+
+        datePickerDialog.setOnCancelListener(dispatchingCancelListener)
+
+        datePickerDialog.show()
+        editDialog = datePickerDialog
```
semanticer, line 348: How about something like:
```suggestion
        with(DatePickerDialog(
            requireContext(),
            null,
            initialTime.year,
            initialTime.monthValue - 1,
            initialTime.dayOfMonth
        )) {
            maxTime?.let { datePicker.maxDate = maxTime.toEpochMillisecond() }
            minTime?.let { datePicker.minDate = minTime.toEpochMillisecond() }
            setOnCancelListener(dispatchingCancelListener)
            show()
            editDialog = this
        }
```

semanticer: Or maybe better:
```suggestion
        editDialog = DatePickerDialog(
            requireContext(),
            null,
            initialTime.year,
            initialTime.monthValue - 1,
            initialTime.dayOfMonth
        ).apply {
            maxTime?.let { datePicker.maxDate = maxTime.toEpochMillisecond() }
            minTime?.let { datePicker.minDate = minTime.toEpochMillisecond() }
            setOnCancelListener(dispatchingCancelListener)
            show()
        }
```

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -350,15 +418,23 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
...
+        this.duration != null -> this.duration
+        this.isNotStarted() && this.startTime == null -> Duration.ZERO
+        this.startTime != null -> Duration.between(this.startTime, timeService.now())
+        else -> throw IllegalStateException("Editable time entry must either have a duration, a start time or not be started yet (have no ids)")
+    }
 
-    private fun TextView.setTextIfDifferent(newText: String) {
+    private fun EditableTimeEntry.startTimeOrNow(): OffsetDateTime = this.startTime ?: timeService.now()
+    private fun EditableTimeEntry.endTimeOrNow(): OffsetDateTime? = when {
+        this.isStopped() -> this.startTime!!.plus(this.duration)
```
semanticer, line 432: It would be really cool to use [contract](https://github.com/Kotlin/KEEP/blob/master/proposals/kotlin-contracts.md) in `isStopped` to get rid of the `!!` but contracts are still experimental and it would have to be top-level function so it's not worth it at this point.

pdrj:
> This is probably a correct behavior in the context of this PR, but I just wanted to flag this to confirm.

Yes, 'OK' should be handled in the other callback (onTimeSet or whatever it's called), where `null` is passed ATM.

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -280,6 +296,58 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
...
+        editDialog = timePickerDialog
+    }
+
+    private fun startEditingDate(
+        initialTime: OffsetDateTime,
+        minTime: OffsetDateTime? = null,
+        maxTime: OffsetDateTime? = null
+    ) {
+        val datePickerDialog =
+            DatePickerDialog(requireContext(), null, initialTime.year, initialTime.monthValue - 1, initialTime.dayOfMonth)
```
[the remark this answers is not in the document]
pdrj: I don't know either 😄 I can test it out and see if it's any different. The default picker looks like in designs though (on my device) - that's why I just went ahead with it.

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
pdrj: I like the second suggestion 👍

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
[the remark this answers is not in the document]
pdrj: Maybe I'll just leave it for now? We are having conversations about using the barrel stuff ATM, so I think we can just go with the simple dialogs here.

Issue #219, opened 2020-04-22
Dispatch wheel related actions from the Wheel component


daividssilverio:
## Description
We need to dispatch the events from the interactions with the Wheel component.
That includes the wheel's handles, arc and the manual duration input.

## Dependencies
Needs the wheel widget to be completed #207 
Needs the actions to be there already: #212 #213 #214 #215 

## Definition of done
- [ ] The DurationInputted is dispatched from the manual duration input
- [ ] The Wheel emits the following actions:
  - [ ] WheelChangedEndTime
  - [ ] WheelChangedStartTime
  - [ ] WheelOffsettedStartTime

PR #251, opened 2020-05-05
Add the wheel widget to the start edit view

### LINKED ISSUE #207: Add the wheel widget

daividssilverio:
### Summary
This PR adds the widget components to the start edit view.

### Technical considerations
What's missing:
- [ ] Proper styling
- [x] Setting the values (start time, end time, duration)
- [x] Handling the focus properly for the numeric input view
- [x] Auto scrolling to focus on the numeric input when inputting duration manually

Feel free to make change requests.

### Relationships
Closes #207 

### Pull Request checklist

### Pointers
Try it out. It works already, more or less.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it after the DRAFT becomes a PR.


timer/src/main/java/com/toggl/timer/startedit/ui/editduration/WheelForegroundView.kt
```diff
@@ -300,6 +303,7 @@ class WheelForegroundView @JvmOverloads constructor(
     private fun touchesBegan() {
         if (touchInteractionIsValid()) {
             isDragging = true
+            editionStateChannel.offer(isDragging)
```
heytherewill, line 306: Do you even ned both a property and the channel?

timer/src/main/res/layout/fragment_dialog_start_edit.xml
```diff
@@ -205,16 +206,37 @@
...
 
-            <View
-                android:id="@+id/wheel_placeholder"
-                android:layout_width="150dp"
-                android:layout_height="150dp"
+            <com.toggl.timer.startedit.ui.editduration.WheelBackgroundView
+                android:id="@+id/wheel_background"
+                android:layout_width="296dp"
+                android:layout_height="296dp"
+                android:layout_marginTop="@dimen/edit_elements_margin"
```
heytherewill, line 213: Extract as dimens

timer/src/main/res/layout/fragment_dialog_start_edit.xml
```diff
@@ -205,16 +206,37 @@
...
+                app:layout_constraintEnd_toEndOf="parent"
+                app:layout_constraintStart_toStartOf="parent"
+                app:layout_constraintTop_toBottomOf="@id/start_time_label" />
+
+            <com.toggl.timer.startedit.ui.editduration.WheelForegroundView
+                android:id="@+id/wheel_foreground"
+                android:layout_width="296dp"
+                android:layout_height="296dp"
```
heytherewill, line 221: Extract as dimens

timer/src/main/res/layout/fragment_dialog_start_edit.xml
```diff
@@ -205,16 +206,37 @@
...
-                android:background="@color/button_active"
                 app:layout_constraintEnd_toEndOf="parent"
                 app:layout_constraintStart_toStartOf="parent"
                 app:layout_constraintTop_toBottomOf="@id/start_time_label" />
 
+            <com.toggl.timer.startedit.ui.editduration.WheelDurationInputView
+                android:id="@+id/wheel_duration_input"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:textSize="28dp"
```
heytherewill, line 231: Extract as dimens and make these SP

timer/src/main/res/layout/fragment_dialog_start_edit.xml
```diff
@@ -205,16 +206,37 @@
...
+                android:layout_width="296dp"
+                android:layout_height="296dp"
+                android:layout_marginTop="@dimen/edit_elements_margin"
+                app:layout_constraintEnd_toEndOf="parent"
+                app:layout_constraintStart_toStartOf="parent"
+                app:layout_constraintTop_toBottomOf="@id/start_time_label" />
+
+            <com.toggl.timer.startedit.ui.editduration.WheelForegroundView
+                android:id="@+id/wheel_foreground"
+                android:layout_width="296dp"
```
heytherewill, line 220: Wouldn't it be nicer to constraint this to the sides with a fixed padding and then use `layout_constraintDimensionRatio="1:1"` to use as much space as the device can offer?

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -133,7 +133,9 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
             stop_time_label,
             stop_divider,
             stop_date_label,
-            wheel_placeholder
+            wheel_background,
+            wheel_foreground,
+            wheel_duration_input
```
heytherewill, line 138: Also, these are getting a bit out of hand. I get the list for adding the scroll alpha listener, that looks nice, but do we really need the list for controlling visibility? Woulnd't a [`Group`](https://developer.android.com/reference/androidx/constraintlayout/widget/Group) be much nicer?

timer/src/main/res/layout/fragment_dialog_start_edit.xml
daividssilverio: Actually... for better control of this widget, I rather leave with dps. Using SPs can cause the numeric input widget grow past the wheel, it can break the view.

timer/src/main/res/layout/fragment_dialog_start_edit.xml
daividssilverio: We don't want to use as much space as we can, the idea is that wheel handles caps can interacted with one hand, just swirling a finger. I didn't do any research on that though. We can try using the aspect ratio and see if people complain about it later.

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
daividssilverio: I think that was implemented like that because of a limitation in the framework. But since I'm not sure, I'll give it a try.

daividssilverio: Yeah, that doesn't work. We can set the alpha value of a `Group`, but that doesn't change the alpha of the views referenced by the group.

timer/src/main/res/layout/fragment_dialog_start_edit.xml
daividssilverio: Let's see how it goes. I've changed it to use a aspect ratio with some reasonable margins.

timer/src/main/res/layout/fragment_dialog_start_edit.xml
daividssilverio: I've extracted the size, but I've kept it as DP.

timer/src/main/java/com/toggl/timer/startedit/ui/editduration/WheelForegroundView.kt
daividssilverio: Oh, it turns out we don't need the prop.

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
heytherewill: Like I said, I'm okay with creating the list here to control the alpha, but we are storing it and then iterating over it to change the visibility of fields when needing to hide then. Is that really the best way?

daividssilverio: It's the simplest way for now. 
Other two ways I see:
- To use data binding + some custom binding adapter (do we want that now?)
- Use a group just to keep reference of the ids + `findViewById` when needed (this doesn't sound like a good idea)

What's the concern about using a list to reference the views here? It doesn't seem to be causing any leaks. If we find a leak on that, we could just use weak references, but I rather not do that until we detect a leak for sure.

pdrj: I don't remember exactly, but I had the following reasons.

- With a group you still have to handle billable visibility separately.
- As soon as you make one item in a group visible (or not visible, I don't remember), then the whole group takes that visibility. This then doesn't work for us, as we sometimes only need to hide the date labels and time/date separators (the dot).

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -217,6 +217,13 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
             .distinctUntilChanged()
             .onEach { nested_scrollview.requestDisallowInterceptTouchEvent(it) }
             .launchIn(lifecycleScope)
+
+        wheel_duration_input.setOnFocusChangeListener { numericInput, hasFocus ->
+            if (hasFocus) {
+                activity?.getSystemService<InputMethodManager>()
+                    ?.showSoftInput(numericInput, InputMethodManager.SHOW_FORCED)
```
heytherewill, line 223: Something like `activity.tryShowingKeyboard()` would be a nice extension, since this is the second place where we include this gross boilerplate

daividssilverio:
@pdrj I'm trying to fix it. Without success so far... I've tried scrolling the root dialog view and the nested scrollview, but that didn't work. Any ideas?

daividssilverio:
@pdrj 

> This should not yet be wired to actually update the editable time entry, right?

yep, not yet. #219 should handle that o/

I've made some changes to the soft input when the numeric input focus changes. The input appears when the view has focus and the keyboard is hidden when the user tries to interact with the bottom sheet or the numeric input loses focus. Please try it out.

PR #264, opened 2020-05-07
✨ Add the StartEdit.WheelChangedEndTime action

### LINKED ISSUE #213: Add the WheelChangedEndTime action to the StartEdit reducer
## Description
- it should simply change the `editableTimeEntry`'s duration
- it shouldn't make any changes if the end time (startTime + duration) is before the start time or the total duration surpasses 999 hours

## Dependencies
None.

## Definition of done
- [x] The action is added
- [x] The reducer handles setting the `editableTimeEntry`'s duration
- [x] Invalid end time (startTime + duration) should be filtered out:
  - [x] startTime + duration < startTime
  - [x] duration > 999 hours
- [x] No effects are produced
- [x] Tests are added
- [x] The [documentation](https://www.notion.so/toggl/WheelChangedEndTime-4eee91ef22044ba58e48a3bf83da9718) is updated to reflect the changes done in this PR

daividssilverio:
### Summary
This PR adds the action that will be emitted by the wheel when interacting with the end handle cap.

### Technical considerations
This action shouldn't be emitted for new and running time entries.
Exceptions are being thrown for those cases.

Feel free to make your suggestion.
We can also just do nothing for new/running time entries.

### Relationships
Closes #213 

### Pull Request checklist

### Pointers
The implementation is very similar to the `DurationInputted` action, you might wanna check that.
The tests are basically the same.

### Review template

## Merge permissions
🦑 it


timer/src/test/java/com/toggl/timer/startedit/domain/WheelChangedEndTimeActionTests.kt
```diff
@@ -0,0 +1,150 @@
...
+        val initialStartTime = now - Duration.ofHours(1)
+        val initialDuration = Duration.ofMinutes(30)
+        val initialTimeEntry = createEditableTimeEntry(initialStartTime, initialDuration)
+        val durationInputted = Duration.ofHours(1).plusMinutes(30)
+
+        reducer.testReduceState(
+            initialState.copy(editableTimeEntry = initialTimeEntry),
+            action = StartEditAction.WheelChangedEndTime(durationInputted)
+        ) {
+            it shouldBe it.copy(
```
semanticer, line 77: I think you meant this?
```suggestion
            it shouldBe initialState.copy(
```

timer/src/test/java/com/toggl/timer/startedit/domain/WheelChangedEndTimeActionTests.kt
```diff
@@ -0,0 +1,150 @@
...
+        val initialStartTime = now - Duration.ofHours(1)
+        val initialDuration = Duration.ofMinutes(30)
+        val initialTimeEntry = createEditableTimeEntry(initialStartTime, initialDuration)
+        val durationInputted = Duration.ofSeconds(-1)
+
+        reducer.testReduceState(
+            initialState.copy(editableTimeEntry = initialTimeEntry),
+            action = StartEditAction.WheelChangedEndTime(durationInputted)
+        ) {
+            it shouldBe it
```
semanticer, line 97: ❓ 
This is probably not what you wanted

timer/src/test/java/com/toggl/timer/startedit/domain/WheelChangedEndTimeActionTests.kt
```diff
@@ -0,0 +1,150 @@
...
+        val initialStartTime = now - Duration.ofHours(1)
+        val initialDuration = Duration.ofMinutes(30)
+        val initialTimeEntry = createEditableTimeEntry(initialStartTime, initialDuration)
+        val durationInputted = Duration.ofHours(Constants.TimeEntry.maxDurationInHours).plusSeconds(1)
+
+        reducer.testReduceState(
+            initialState.copy(editableTimeEntry = initialTimeEntry),
+            action = StartEditAction.WheelChangedEndTime(durationInputted)
+        ) {
+            it shouldBe it
```
semanticer, line 112: And here

timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
```diff
@@ -283,6 +298,8 @@ class StartEditReducer @Inject constructor(
 
     private fun EditableTimeEntry.shouldStart() = this.ids.isEmpty()
     private fun EditableTimeEntry.isRunningOrNew() = this.startTime == null || this.duration == null
```
semanticer, line 300: this could be a composite of two methods below

timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
```diff
@@ -283,6 +298,8 @@ class StartEditReducer @Inject constructor(
...
+    private fun EditableTimeEntry.isNew() = this.startTime == null
+    private fun EditableTimeEntry.isRunning() = this.duration == null
```
semanticer, line 302: We have a lot of universally useful methods here. Some of them are repeated in `StartEditDialogFragment` and probably elsewhere. I'd consider moving them to `EditableTimeEntry` file and use them where ever we can. Let's reserve private extension methods for specific use-cases in specific contexts.

timer/src/test/java/com/toggl/timer/startedit/domain/WheelChangedEndTimeActionTests.kt
daividssilverio: YES! I fixed this on the other branch, but I've wrote a bunch of those. Let me fix them all.

timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
daividssilverio: I agree, this is starting to look out of place.

PR #287, opened 2020-05-11
✨ Dispatch DateTimePicked with proper date time

### LINKED ISSUE #218: Dispatch DateTimePicked actions when picking dates and times
## Description
We need to dispatch `DateTimePicked` actions when dates and times are picked.

## Dependencies
Depends on #217

## Definition of done
- [ ] When the date picker picks a date, a `DateTimePicked` action is dispatched with its `DateTime` value combining the current `editableTimeEntry`'s startTime time part.
- [ ] When the time picker picks a time, a `DateTimePicked` action is dispatched with its `DateTime` value combining the current `editableTimeEntry`'s startTime date part.
- [ ] Clarification: Both `start` and `end` date and time pickers should dispatch their respective date times.

semanticer:
### Summary
This PR implements everything described in #218 

### Technical considerations
`startEditingDate` started to look too overloaded so I created some higher abstraction methods to make everything more explicit and hopefully easier to read.

Feel free to suggest any alternatives

### Relationships

Closes #218 

### Pull Request checklist

### Pointers
Run this and check if you can change start time & start date. 

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -350,28 +350,65 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
...
+    private fun startEditingStartTime(initialTime: OffsetDateTime) =
+        startEditingTime(initialTime, TimePickerDialog.OnTimeSetListener { _, hourOfDay, minute ->
+            val changedTime = initialTime
+                .withHour(hourOfDay)
+                .withMinute(minute)
+            store.dispatch(StartEditAction.DateTimePicked(changedTime))
+        })
+
+    private fun startEditingEndTime(initialTime: OffsetDateTime) =
+        startEditingTime(initialTime, null)
```
daividssilverio, line 362: I was wondering why picking the end time wasn't working, gotta set this listener.

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -350,28 +350,65 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
...
+        startEditingDate(initialTime, DatePickerDialog.OnDateSetListener { _, year, month, dayOfMonth ->
+            val changedTime = initialTime
+                .withYear(year)
+                .withMonth(month + 1)
+                .withDayOfMonth(dayOfMonth)
+            store.dispatch(StartEditAction.DateTimePicked(changedTime))
+        }, maxTime = maxTime)
+
+    private fun startEditingEndDate(initialTime: OffsetDateTime, minTime: OffsetDateTime) =
+        startEditingDate(initialTime, null, minTime = minTime)
```
daividssilverio, line 389: Same for the end date.

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
semanticer: @daividssilverio I thought this shouldn't be a part of this issue https://github.com/toggl/android/issues/218.

daividssilverio: I see, I've made a ©️🍝 when writing the definition of done of this issue. 

The second line was supposed to point to the `duration` and not the `startTime` again.

Please let me know if you see something weird like that again.

daividssilverio: Sorry for the misleading description. I've updated the issue.

timer/src/main/java/com/toggl/timer/startedit/ui/StartEditDialogFragment.kt
```diff
@@ -391,27 +391,54 @@ class StartEditDialogFragment : BottomSheetDialogFragment() {
...
+            onTimeSetListener,
+            initialTime.hour,
+            initialTime.minute,
+            true
+        )
+        with(dialog) {
+            setOnCancelListener(dispatchingCancelListener)
+            show()
+            editDialog = this
+        }
```
daividssilverio, line 412: What do you think about: 
```suggestion
        TimePickerDialog(
            requireContext(),
            onTimeSetListener,
            initialTime.hour,
            initialTime.minute,
            true
        ).run {
            setOnCancelListener(dispatchingCancelListener)
            show()
            editDialog = this
        }
```

It's just a small suggestion to avoid creating that val, but it doesn't make much of a difference.
If you accept this suggestion, do the same for the date picker.

PR #293, opened 2020-05-12
Implement time entry suggestions

### LINKED ISSUE #259: Make the UpdateAutocompleteSuggestionsEffect return time entry suggestions
## Description
We need to update the effect that calculates autocomplete suggestions to make it handle time entry suggestions.
What we need to do is mimic the way the Xamarin app does the search. We basically split the query by spaces and make sure that every word matches some part of the time entry. This includes the description, project names, task names, tags and clients. So for example, the query `wo mob` would match an entry `Working` with the project `mobile app`, and would also match a time entry with just the description `aaasdadasmobdasdasdsawoaowowodadasdmo`

This case is to be handled last in the effect, whenever we can't match something more specific like projects/tasks or tags.

## Definition of done
- [ ] The effect is updated to return a list of time entry autocomplete suggestions
- [ ] Tests ensure that it matches the correct TEs

heytherewill:
### Summary
This adds time entry suggestions as the user types in the description text field

### Technical considerations
I'm not implementing caching atm. Let's wait and see if it's really needed.
I've also throttled the emission of new events by 300ms. It's close but not quite like the current implementation, which *debounces* by 300ms and feels like it's lagging behind rather than updating in realtime.

### Relationships
Closes #259 

### Pull Request checklist

### Pointers
Go commit by commit and make sure the tests are correct

### Review template
- [ ] The code considers i18n

## Merge permissions
Only myself


timer/src/main/java/com/toggl/timer/startedit/domain/UpdateAutocompleteSuggestionsEffect.kt
```diff
@@ -19,8 +21,46 @@ class UpdateAutocompleteSuggestionsEffect(
     private val projects: Map<Long, Project>,
     private val timeEntries: Map<Long, TimeEntry>
 ) : Effect<StartEditAction.AutocompleteSuggestionsUpdated> {
+
+    private val wordSeparator = " "
```
daividssilverio, line 24: I wonder how well this works for languages that don't use spaces to separate words like Japanese.

pdrj:
Maybe we can use https://developer.android.com/reference/kotlin/java/text/BreakIterator#getWordInstance(java.util.Locale) for the eastern languages.

PR #294, opened 2020-05-12
☸️ Wheel actions dispatching

### LINKED ISSUE #219: Dispatch wheel related actions from the Wheel component

semanticer:
### Summary
This PR is an attempt to implement #219 ~even with the problematic `WheelOffsettedStartTime` action~ the existence of which is questionable. 

### Technical considerations
~To properly implement `WheelOffsettedStartTime` I had to change the way we dispatch events from `WheelForegroundView`. Rather than having two separate flows for `startTime` and `endTime` which are updated independently, we now have single `startAndEndTimeFlow` backed by a new single source of truth `startAndEndTimeChannel`.  
`startTime` and `endTime` properties are also backed by `startAndEndTimeChannel`. 
This is making time inconsistencies impossible but more importantly, it lets us do this:~
```kotlin
        if (updateType == WheelUpdateType.EditBothAtOnce) {
            // ... code ... // 
            startAndEndTime = nextStartTime to nextEndTime
        }
```
~This is updating start and end time in one event to make `WheelOffsettedStartTime` dispatching possible in our Fragment.~

I refactored `WheelForegroundView` slightly to reduce possible redundancies. Start and end times are now backed only by channel values.

~We could simply eliminate `WheelOffsettedStartTime` and everything would be fine. We would lose some semantic value but no functionality. If we decide to do this, we don't need the whole `Wheel` refactor although I would argue that some of these changes are useful anyway.~
We decided to drop `WheelOffsettedStartTime` as it is not necessary and makes everything harder to implement.

### Relationships

Closes #219

### Pull Request checklist

## Reviewing
Check dispatched actions in the log to see if proper actions are dispatched without redundancies. Especially when changing the offset, the wheel should be emitting only `WheelOffsettedStartTime` actions.
Make sure I didn't break anything. There are still some weird glitches when the wheel suddenly jumps somewhere but I'm not sure if it's me or some other broken code. 

### Review template

- [ ] The code considers i18n

## Merge permissions
Me


daividssilverio:
Gotta say, I like your changes.
If we refactor the wheel the way you are proposing, we could later further change it to work directly in terms of `start time` and `duration` instead of `start time` and `end time`.

Also, not having to worry about inconsistency is a plus.

If we rather not do that, we could simply go with:
```
wheel_foreground.startTimeFlow
            .onEach {
                store.dispatch(StartEditAction.WheelChangedStartTime(it))
            }.launchIn(lifecycleScope)

        wheel_foreground.endTimeFlow
            .onEach {
                val duration = wheel_foreground.startTime.absoluteDurationBetween(it)
                store.dispatch(StartEditAction.WheelChangedEndTime(duration))
            }.launchIn(lifecycleScope)
```

and of course, drop the `WheelOffsettedStartTime` which doesn't make sense with the current implementation.

Pinging @juanlaube as to gather more opinions on this since the iOS foreground wheel is being implemented right now. Do you think this could be beneficial on iOS as well?

timer/src/main/java/com/toggl/timer/startedit/ui/editduration/WheelForegroundView.kt
```diff
@@ -106,42 +107,52 @@ class WheelForegroundView @JvmOverloads constructor(
...
+        }
 
-            endTimeAngle = endTime.timeOfDay().toAngleOnTheDial().toPositiveAngle()
-            endTimePosition.updateWithPointOnCircumference(center, endTimeAngle, handleCapsPositionRadius)
-            arc.update(startTimeAngle, endTimeAngle)
-            wheelHandleDotIndicator.update(startTimeAngle, endTimeAngle)
-            endTimeChannel.offer(endTime)
-            invalidate()
+    var endTime: OffsetDateTime
+        get() = startAndEndTimeChannel.value.second
```
daividssilverio, line 150: The old ` if (field < startTime) startTime else field` was in place to avoid having negative durations being represented in the wheel. I think that is still needed.

timer/src/main/java/com/toggl/timer/startedit/ui/editduration/WheelForegroundView.kt
```diff
@@ -106,42 +107,52 @@ class WheelForegroundView @JvmOverloads constructor(
     var maximumEndTime: OffsetDateTime = OffsetDateTime.MAX
 
     private var startTimeAngle = 0.0
-    var startTime: OffsetDateTime = OffsetDateTime.now()
+    private var endTimeAngle = 0.0
+
+    private var startAndEndTimeChannel = ConflatedBroadcastChannel<Pair<OffsetDateTime, OffsetDateTime>>(OffsetDateTime.now() to OffsetDateTime.now())
+
+    private var startAndEndTime: Pair<OffsetDateTime, OffsetDateTime>
```
daividssilverio, line 114: Some points about this:
1. We still need to update the backing fields even if `center.x.isNaN()`. The wheel will use the start/end backing fields to recreate the drawing delegates and other internal fields on `onLayout`
2. What we can't/shouldn't do is to update the `startTimeAngle`, `startTimePosition`, `endTimeAngle`, `endTimePosition `.
3. I think we don't need to offer a new event to the `startAndEndTimeChannel` if both the start and end times didn't change.

daividssilverio: What about making this pair a data class instead? It's more pleasant to read `startTime` and `endTime` than `first` and `second`.

timer/src/main/java/com/toggl/timer/startedit/ui/editduration/WheelForegroundView.kt
```diff
@@ -100,49 +100,49 @@ class WheelForegroundView @JvmOverloads constructor(
...
 
             if (center.x.isNaN()) return
 
-            startTimeAngle = startTime.timeOfDay().toAngleOnTheDial().toPositiveAngle()
+            startTimeAngle = newStartTime.timeOfDay().toAngleOnTheDial().toPositiveAngle()
             startTimePosition.updateWithPointOnCircumference(center, startTimeAngle, handleCapsPositionRadius)
             arc.update(startTimeAngle, endTimeAngle)
             wheelHandleDotIndicator.update(startTimeAngle, endTimeAngle)
-            startTimeChannel.offer(startTime)
+            startTimeChannel.offer(newStartTime)
```
daividssilverio, line 123: We need to set the backing field even when `center.x.isNan()` because of the logic to update the layout drawing delegates & other backing fields that happen on `onLayout`. Same for `endTime`.

daividssilverio:
Discussion that led to this implementation: https://toggl.slack.com/archives/C02APMYTM/p1589363090283600

PR #295, opened 2020-05-12
🔨 Reduce AutocompleteSuggestionTapped for time entries

### LINKED ISSUE #258: Update the StartEditReducer's AutocompleteSuggestionTapped action to handle TimeEntry suggestions
## Description
When the user picks a suggested a time entry, we change the details of the entry being edited/created to match the selected suggestion
Bear in mind that this method should never alter the id(s) of the running time entry, just all all other fields
## Definition of done

daividssilverio:
### Summary
This handles the `AutocompleteSuggestionTapped` action for Time Entry suggestions.

### Technical considerations
None.

Feel free to make a suggestion.

### Relationships
Closes #258

### Pull Request checklist

### Pointers
Check the tests.

### Review template

## Merge permissions
🦑 it


timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
```diff
@@ -116,8 +117,14 @@ class StartEditReducer @Inject constructor(
             }
             is StartEditAction.AutocompleteSuggestionsUpdated ->
                 state.mutateWithoutEffects { copy(autocompleteSuggestions = action.autocompleteSuggestions) }
-            is StartEditAction.AutocompleteSuggestionTapped ->
-                noEffect()
+            is StartEditAction.AutocompleteSuggestionTapped -> {
+                state.mutateWithoutEffects {
+                    when (action.autocompleteSuggestion) {
+                        is AutocompleteSuggestion.TimeEntry -> modifyWithTimeEntrySuggestion(action.autocompleteSuggestion)
+                        else -> this
```
heytherewill, line 124: Maybe don't `else`, but rather just fill all possibilities and throw while they are not implemented

timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
```diff
@@ -236,6 +243,18 @@ class StartEditReducer @Inject constructor(
...
 
+    private fun StartEditState.modifyWithTimeEntrySuggestion(autocompleteSuggestion: AutocompleteSuggestion.TimeEntry): StartEditState =
+        StartEditState.editableTimeEntry.modify(this) {
+            val timeEntrySuggestion = autocompleteSuggestion.timeEntry
+            it.copy(
+                workspaceId = timeEntrySuggestion.workspaceId,
+                description = timeEntrySuggestion.description,
+                projectId = timeEntrySuggestion.projectId,
+                tagIds = timeEntrySuggestion.tagIds,
+                billable = timeEntrySuggestion.billable
```
heytherewill, line 254: You're missing taskId

daividssilverio: 🤔 the editableTimeEntry doesn't have a `taskId` yet. I'll add one.

PR #303, opened 2020-05-14
🔨 Handle auto complete project suggestion taps

### LINKED ISSUE #256: Update the StartEditReducer's AutocompleteSuggestionTapped action to handle Project suggestions
## Description
When you select a project suggestion, it sets the project of the editable time entry.
If the project is in a workspace different than the one of the time entry, the time entry workspace is changed as well
Selecting a project suggestion should also remove the word that starts with an @ from the description (do so by checking the cursor position as well)

## Definition of done

daividssilverio:
### Summary
This handles auto complete project suggestions taps.

### Technical considerations
🤔 I had to add the cursor position to the state in order to figure out what part of the current description of the editable time entry to remove when a suggestion is accepted. I've used an extension from my other PR to do that (10490b728afcece0d7878faee3bf3d0974f1a539 was 🍒 ⛏️ ed).

By all means, feel free to suggest something else. 

Another way of doing the replacement would be keeping track of the query that generated the suggestion. We could add that query to the `AutocompleteSuggestion` data objects. It could look like this:
```kotlin
sealed class AutocompleteSuggestion {
    data class TimeEntry(val timeEntry: com.toggl.models.domain.TimeEntry) : AutocompleteSuggestion()
    data class Project(val project: com.toggl.models.domain.Project, val sourceQuery: String) : AutocompleteSuggestion()
    data class Task(val task: com.toggl.models.domain.Task, val sourceQuery: String) : AutocompleteSuggestion()
    data class Tag(val tag: com.toggl.models.domain.Tag, val sourceQuery: String) : AutocompleteSuggestion()
    data class CreateProject(val name: String, val sourceQuery: String) : AutocompleteSuggestion()
    data class CreateTag(val name: String, val sourceQuery: String) : AutocompleteSuggestion()
}
```
We would then use that `sourceQuery` directly to remove its last occurrence in the `editableTimeEntry`'s description.

Any suggestion?

### Relationships
Closes #256 

### Pull Request checklist

Still gotta add corner case tests. But I'm waiting on some opinions about the query handling.

### Pointers
Start reviewing. Check the tests. Check my other PR #302 .

### Review template

## Merge permissions
🦑 it after te draft is done.


timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
```diff
@@ -247,6 +249,19 @@ class StartEditReducer @Inject constructor(
         }
     }
 
+    private fun StartEditState.modifyWithProjectSuggestion(autocompleteSuggestion: AutocompleteSuggestion.Project): StartEditState =
+        StartEditState.editableTimeEntry.modify(this) {
+            val projectToken = '@'
```
heytherewill, line 254: Better to have this as a constant somewhere and reuse it in all places where we check for projects

timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
```diff
@@ -51,6 +52,7 @@ class StartEditReducer @Inject constructor(
                 state.mutate {
                     editableTimeEntry ?: throw EditableTimeEntryShouldNotBeNullException()
                     StartEditState.editableTimeEntry.modify(this) {
+                        copy(cursorPosition = action.cursorPosition)
                         it.copy(description = action.description)
```
heytherewill, line 56: Is this code even correct? The first `copy`'s return is being ignored, innit?

heytherewill: Nonetheless, we need to test that the description is being adequately updated when this action is received

daividssilverio: Forgot about the test, thonks.

PR #305, opened 2020-05-14
✨Autocomplete suggestion task selected

### LINKED ISSUE #254: Update the StartEditReducer's AutocompleteSuggestionTapped action to handle Task suggestions
## Description
When you select a task suggestion, it sets the project of the editable time entry and the task of the editable time entry as well.
If the project is in a workspace different than the one of the time entry, the time entry workspace is changed as well
Selecting a project suggestion should also remove the word that starts with an @ from the description (do so by checking the cursor position as well)

## Definition of done

semanticer:
### Summary
This PR should implement everything in #254. 

### Technical considerations
I still need to fix some tests but the main issue I want to discuss is that we don't have access to `cursorPosition` in our reducer. See my code comments below. 

Please let me know if you have any ideas

### Relationships
Closes #254

### Pull Request checklist

### Pointers
Make sure that those tests are actually testing the right behavior because I'm not 100% sure.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
```diff
@@ -256,6 +258,26 @@ class StartEditReducer @Inject constructor(
...
+                editableProject = EditableProject(
+                    name = project.name,
+                    color = project.color,
+                    active = project.active,
+                    isPrivate = project.isPrivate,
+                    billable = project.billable,
+                    workspaceId = project.workspaceId,
+                    clientId = project.clientId
+                ),
+                description = it.description.clearLastSubstringFromAnyTokenToPosition(charArrayOf('@')),
```
semanticer, line 275: `charArrayOf('@')` should be probably referenced from somewhere but the bigger issue here is that I'm using the default `cursorPosition` but I should be using the same `cursorPosition` that was used in `UpdateAutocompleteSuggestionsEffect`. Should we persist `cursorPosition` in our state?

daividssilverio: That's what I did here https://github.com/toggl/android/pull/303/commits/4dc1e4f0dadc03b0bee9c85e18e14dca85dd7181

timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
```diff
@@ -256,6 +258,26 @@ class StartEditReducer @Inject constructor(
             it.copy(editableProject = EditableProject(name = autocompleteSuggestion.name, workspaceId = 1))
         }
 
+    private fun StartEditState.modifyWithTaskSuggestion(autocompleteSuggestion: AutocompleteSuggestion.Task): StartEditState =
+        StartEditState.editableTimeEntry.modify(this) {
+            val taskSuggestion = autocompleteSuggestion.task
+            val project = this.projects[taskSuggestion.projectId] ?: throw ProjectDoesNotExistException()
+            it.copy(
+                editableProject = EditableProject(
```
daividssilverio, line 266: I was under the impression that the `EditableProject` was something used for new projects only.
Shouldn't we simply set the `projectId` here? If that's not the case, #303 is wrong. I actually set the `editableProject` to null when a project suggestion was picked.

semanticer: Ouu... I think you're right

daividssilverio: That was the idea, right @pdrj and @heytherewill (you've created the `EditableProject`).

PR #308, opened 2020-05-15
🚫 Ignore .idea/misc.xml

daividssilverio:
### Summary
This just ~commits~ adds the damn `.idea/misc.xml` file that keeps changing on people's macs to the project's .gitignore.

### Technical considerations
~Do we wan't to commit this?~ Should we just add it to .gitignore instead?

It's either that or adding it to .gitignore.
How does the file look on your machine @pdrj @semanticer @heytherewill ?

### Relationships
Closes #Nothing

### Pull Request checklist

### Pointers
Open the project with AndroidStudio, do something, see if the file changes.

### Review template

## Merge permissions
🦑 it when everybody agrees on what to do.


semanticer:
Mine looks like this

```xml
<?xml version="1.0" encoding="UTF-8"?>
<project version="4">
  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" project-jdk-name="1.8" project-jdk-type="JavaSDK" />
</project>
```
I think we should rather add this to `.gitignore` in order to be platform agnsotic.

daividssilverio:
Done, added it to `.gitignore`.

semanticer:
I'm gonna merge this since @pdrj is not here and he wasn't against when it was discussed on slack.

PR #310, opened 2020-05-18
🏷 Tag suggestion tapped

### LINKED ISSUE #261: Update the StartEditReducer's AutocompleteSuggestionTapped action to handle tag suggestions
## Description
When you select a tag suggestion, it adds the tag to the list of tags of the editable time entry.
Selecting a tag suggestion should also remove the word that starts with an # from the description (do so by checking the cursor position as well)

## Definition of done

semanticer:
### Summary
This PR implements everything from #261 
It also implements proper fetch of tag suggestions which is not mentioned in related issue but was probably required.
~Description manipulation was also unified across our suggestion tap handlers.~

### Technical considerations
Some of the tests were shamelessly copy-pasted from other suggestion handler tests. We should probably somehow unify the testing of our description clearing.

Let me know

### Relationships

Closes #261 

### Pull Request checklist

### Pointers
Commit by commit should work here

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it


semanticer:
By using `it.description.substringBeforeLast(delimiter)` across all of our tap suggestion handlers, I broke some tests. Now I'm not 100% sure if requirements have changed and I should update the tests or if we wanted this behavior only in some of the tap suggestion handlers and I should, therefore, rollback the changes.

PR #316, opened 2020-05-20
✨ Add the CalendarService boilerplate

### LINKED ISSUE #267: Add the Calendar service boilerplate
## Description

Add the calendar service boilerplate. This service will be used to fetch the existing calendars and calendar events using EventKit. For now just add the service boilerplate and define the `Calendar` and `CalendarEvent` structs, actual integration with EventKit will happen later.

The calendar service protocol should contain the following methods

```
func getAvailableCalendars() -> [Calendar]
func getCalendarEvents(between startDate: Date, and: endDate: Date, in calendars: [Calendar]) -> [CalendarEvent]
```

Calendar properties:
- id: String
- name: String
- sourceName: String

CalendarEvent properties:
- id: String
- startTime: Date
- duration: TimeInterval
- description: String
- color: Color (String?)
- calendarId: String

daividssilverio:
### Summary
This adds the services and related classes requested on #267.
Basically the `CalendarService` boilerplate.

### Technical considerations
~I've added the classes based on the [old implementation.](https://github.com/toggl/mobileapp/tree/develop/Toggl.Core/Calendar).~
Did you have something else in mind @juanlaube ?

This could be a chance to change how we are doing things in the calendar.

### Relationships
Closes #267

### Pull Request checklist

### Pointers
Check the old code. Think about how this is going to be used.

### Review template

## Merge permissions
🦑 it


environment/src/main/java/com/toggl/environment/services/calendar/CalendarEvent.kt
```diff
@@ -0,0 +1,18 @@
...
+
+data class CalendarEvent(
+    val id: String,
+    val source: CalendarSource,
+    val duration: Duration?,
+    val startTime: OffsetDateTime,
+    val description: String,
+    val project: String = "",
+    val task: String = "",
+    val client: String = "",
```
semanticer, line 14: The fact that these are defined as strings makes this `data class` UI-level in my eyes, meaning that I would expect this data class to be constructed somewhere in the Calendar module with the help of 'real' project, task, and client objects.

environment/src/main/java/com/toggl/environment/services/calendar/CalendarEvent.kt
```diff
@@ -0,0 +1,18 @@
+package com.toggl.environment.services.calendar
+
+import org.threeten.bp.Duration
+import org.threeten.bp.OffsetDateTime
+
+data class CalendarEvent(
```
semanticer, line 6: If I understand this correctly, this data class is designed to hold either Calendar Event data or Time Entry data based on the source. This seems to be the ideal place for a sealed class hierarchy with one parent sealed class and two inheriting data classes representing Time Entry and Calendar Event with 'hard' non-optional properties. This will ensure that we won't have some inconsistent data with properties of Time Entry but Calendar Event as a source etc.

environment/src/main/java/com/toggl/environment/services/calendar/CalendarSource.kt
```diff
@@ -0,0 +1,6 @@
+package com.toggl.environment.services.calendar
+
+enum class CalendarSource {
```
semanticer, line 3: This enum should be completely unnecessary given what I suggested above.

environment/src/main/java/com/toggl/environment/services/calendar/CalendarEvent.kt
```diff
@@ -0,0 +1,18 @@
...
+    val duration: Duration?,
+    val startTime: OffsetDateTime,
+    val description: String,
+    val project: String = "",
+    val task: String = "",
+    val client: String = "",
+    val color: String = "",
+    val timeEntryId: Long? = null,
+    val calendarId: String = ""
+)
```
heytherewill, line 18: Wouldn't it be nicer to make this a sealed class that wraps either a time entry or a calendar event? That'd simplify this massively and make pattern matching better

environment/src/main/java/com/toggl/environment/services/calendar/CalendarEvent.kt
juanlaube: No, calendar events are not either event data or time entries, for that we will have a `CalendarItem` that will hold either a `CalendarEvent` or a `TimeEntry`.

environment/src/main/java/com/toggl/environment/services/calendar/CalendarEvent.kt
juanlaube, line 14: These are not properties of a calendar event and they do not belong in the `CalendarEvent` class.

environment/src/main/java/com/toggl/environment/services/calendar/CalendarEvent.kt
```diff
@@ -0,0 +1,18 @@
...
+    val id: String,
+    val source: CalendarSource,
+    val duration: Duration?,
+    val startTime: OffsetDateTime,
+    val description: String,
+    val project: String = "",
+    val task: String = "",
+    val client: String = "",
+    val color: String = "",
+    val timeEntryId: Long? = null,
```
juanlaube, line 16: Same here.

environment/src/main/java/com/toggl/environment/services/calendar/CalendarService.kt
```diff
@@ -0,0 +1,12 @@
+package com.toggl.environment.services.calendar
+
+import org.threeten.bp.OffsetDateTime
+
+interface CalendarService {
+    fun getAvailableCalendars(): List<Calendar>
+    fun getCalendarEvents(
+        fromStartDate: OffsetDateTime,
+        toEndDate: OffsetDateTime,
+        fromCalendar: Calendar
```
juanlaube, line 10: You can get events from multiple calendars:
```suggestion
        fromCalendars: List<Calendar>
```

PR #318, opened 2020-05-20
🛠 Billable tooltip

### LINKED ISSUE #188: Display a tooltip when billable state changes
## Description
When changing the billable state of a time entry using the '$' button above keyboard, a tooltip should appear describing the new state.

Ie. when a TE is not billable and users presses the '$' button, a tooltip with 'Billable' text appears.

Screenshot 2020-04-14 at 14 14 54

Spec here: [invision](https://toggl.invisionapp.com/d/main#/console/19556473/409562660/inspect)
- 14sp font
- -0.15 spacing

Screenshot 2020-04-16 at 12 43 14
Screenshot 2020-04-16 at 12 43 06

## Definition of done

- [ ] a tooltip, as described above, appears on each billable state change

semanticer:
### Summary
This PR displays a tooltip above billable toggle as described in #188 
It also makes this functionality reusable and modifiable for future use.

### Technical considerations
This solution is using simple [PopupWindow](https://developer.android.com/reference/android/widget/PopupWindow) with custom card layout and `handler` that dismisses the window after a given period of time. 

```kotlin
showTooltip(
    textResId = tooltipText,
    duration = Duration.ofSeconds(5)
).above(billableButton)
```

`showTooltip` is provided as an extension function on any `LifecycleOwner`. This `LifecycleOwner` is then used to unsubscribe handlers to prevent memory leaks automatically without any effort on the client's side.

I had to move `billableButton`'s `onClickListener` setup into the billable state listener in order to display correct tooltip text for both transition states.

To get the right shape look I had to reorganize and set up our shape resources for easier future use.

Some of the resources like colors, text sizes, etc are still preliminary since we don't have the final spec yet.

### Relationships

Closes #188 

### Pull Request checklist

### Pointers
This should be fairly easy to review as a whole, no need to go commit by commit (but they are well separated so it's up to you)

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it


common-android/src/main/res/values/strings.xml
```diff
@@ -15,6 +15,7 @@
     <string name="add_project">Add a project</string>
     <string name="add_tags">Add tags</string>
     <string name="billable">Billable</string>
+    <string name="non_billable">Non Billable</string>
```
heytherewill, line 18: Please either add translations or create a translation request issue

common-android/src/main/java/com/toggl/common/Tooltip.kt
```diff
@@ -0,0 +1,90 @@
...
+            h.removeCallbacksAndMessages(targetView)
+        }
+    })
+}
+
+data class BasicTooltipParams(val lifecycleOwner: LifecycleOwner, @StringRes val textResId: Int, val duration: Duration)
+
+enum class Position {
+    ABOVE,
+    BELOW
```
heytherewill, line 89: Let's not use caps for enum classes
```suggestion
    Above,
    Below
```

common-android/src/main/java/com/toggl/common/Tooltip.kt
```diff
@@ -0,0 +1,90 @@
...
+        animationStyle = R.style.PopupAnimation
+    }
+
+    with(tooltipView) {
+        findViewById<TextView>(R.id.tooltip_text).setText(textResId)
+        measure(0, 0)
+    }
+
+    val yOffset =
+        if (position == Position.ABOVE) -(targetView.height + tooltipView.measuredHeight + res.getDimensionPixelSize(R.dimen.grid_1))
```
heytherewill, line 65:
```suggestion
        if (position == Position.Above) -(targetView.height + tooltipView.measuredHeight + res.getDimensionPixelSize(R.dimen.grid_1))
```

common-android/src/main/java/com/toggl/common/Tooltip.kt
```diff
@@ -0,0 +1,90 @@
...
+    duration: Duration = Duration.ofSeconds(2)
+) = BasicTooltipParams(this, textResId, duration)
+
+fun BasicTooltipParams.above(targetView: View) =
+    showTooltipFor(
+        targetView,
+        textResId,
+        lifecycleOwner,
+        duration,
+        Position.ABOVE
```
heytherewill, line 26:
```suggestion
        Position.Above
```

common-android/src/main/java/com/toggl/common/Tooltip.kt
```diff
@@ -0,0 +1,90 @@
...
+    )
+
+fun BasicTooltipParams.below(targetView: View) =
+    showTooltipFor(
+        targetView,
+        textResId,
+        lifecycleOwner,
+        duration,
+        Position.BELOW
```
heytherewill, line 35:
```suggestion
        Position.Below
```

common-android/src/main/java/com/toggl/common/Tooltip.kt
```diff
@@ -0,0 +1,90 @@
...
+    )
+
+fun showTooltipFor(
+    targetView: View,
+    @StringRes textResId: Int,
+    lifecycleOwner: LifecycleOwner,
+    duration: Duration = Duration.ofSeconds(2),
+    position: Position = Position.ABOVE
```
heytherewill, line 43:
```suggestion
    position: Position = Position.Above
```

common-android/src/main/java/com/toggl/common/Tooltip.kt
```diff
@@ -0,0 +1,90 @@
...
+        LinearLayout.LayoutParams.WRAP_CONTENT,
+        false
+    ).apply {
+        elevation = targetView.context.resources.getDimension(R.dimen.plane_08)
+        animationStyle = R.style.PopupAnimation
+    }
+
+    with(tooltipView) {
+        findViewById<TextView>(R.id.tooltip_text).setText(textResId)
+        measure(0, 0)
```
daividssilverio, line 61: `measure(0, 0)` does what?

semanticer: `tooltipView.measuredHeight` and `tooltipView.measuredWidth` would return `0` if not for this. Let me know if there is a better way.

common-android/src/main/java/com/toggl/common/Tooltip.kt
```diff
@@ -0,0 +1,90 @@
...
+
+    popupWindow.showAsDropDown(targetView, xOffset, yOffset)
+
+    val h = Handler().apply {
+        postDelayed(duration.toMillis(), targetView) {
+            popupWindow.dismiss()
+        }
+    }
+
+    lifecycleOwner.lifecycle.addObserver(object : DefaultLifecycleObserver {
```
daividssilverio, line 78: Is it alright to add this observer and forget about it?

common-android/src/main/java/com/toggl/common/Tooltip.kt
daividssilverio: I'm more curious about the meaning of that. It reads like: "measure with 0 width and 0 height", but it actually means "measure with unspecified measure specs". Can we use the flags instead? `View.MeasureSpec.UNSPECIFIED`?

common-android/src/main/java/com/toggl/common/Tooltip.kt
semanticer: Should be https://github.com/googlecodelabs/android-lifecycles/issues/5 
Otherwise, the whole lifecycle magic wouldn't make much sense.

common-android/src/main/java/com/toggl/common/Tooltip.kt
semanticer: You are right, that will be certainly less confusing.

common-android/src/main/res/values/strings.xml
semanticer: I found translations in Xamarin app

PR #320, opened 2020-05-21
Fix autocomplete behavior

### LINKED ISSUE #314: Fix up the token replacement logic when tapping autocomplete suggestions
## Description
This might need some corrections and is up to discussion.

We need to define and decide what happens to the StartEdit's description field when an auto complete suggestion is tapped.

### Current behaviour
Currently as described in the tests added by #303 and #305 we are:
- Using the `cursorPosition`
  - To change the query used to provide suggestions
  - To select **up to** which part of the description in the StartEdit's description should be replaced when a suggestion is accept
  - The last shortcut token up to the `cursorPosition` will be used to tell if we are suggesting projects or tags.

Here are a series of examples of the current behaviour:
Assume `$` is the cursor.

- Example 1: 
  Starting description:
  ```
  "Creating GH issues #1 #2 #3 @project $"
  ```
  Project suggestions will appear, and they will be based on a query **"project "**.
  Tapping on a suggestion would make the description become:
  ```
  "Creating GH issues #1 #2 #3 $"
  ```

- Example 2:
  ```
  "Creating GH issues #1 #2 #3 @pro$ject"
  ```
  Project suggestions will appear, and they will be based on a query **"pro".**
  Tapping on a suggestion would make the description become:
  ```
  "Creating GH issues #1 #2 #3 $ject"
  ```

- Example 3:
  ```
  "Creating GH issues #1 #2 #3 $@project"
  ```
  Tag suggestions will appear, and they will be based on a query **"3 "**.
  Tapping on a suggestion would make the description become:
  ```
  "Creating GH issues #1 #2 $@project"
  ```

- Example 4:
  ```
  "Creating GH issues$ #1 #2 #3 @project"
  ```
No tag or project suggestions will appear.

<hr>

### Expected/Desired behaviour
What we have isn't actually that different, we basically want to truncate everything after the token that is currently providing the query that is generating the suggestions.

- We want to use the `cursorPosition`:
  - To change the query used to provide suggestions
  - To select **from which** shortcut token of the description in the StartEdit's description should be truncated
  - The last shortcut token up to the `cursorPosition` will be used to tell if we are suggesting projects or tags.

Here are a series of examples for the correct/expected behaviour:
Assume `$` is the cursor.

- Example 1: 
  Starting description:
  ```
  "Creating GH issues #1 #2 #3 @project $"
  ```
  Project suggestions will appear, and they will be based on a query **"project "**.
  Tapping on a suggestion would make the description become:
  ```
  "Creating GH issues #1 #2 #3 $"
  ```
(Same as the current behaviour).

- Example 2:
  ```
  "Creating GH issues #1 #2 #3 @pro$ject"
  ```
  Project suggestions will appear, and they will be based on a query **"pro"**. (Or **"project"**, I'm not sure sure)
  Tapping on a suggestion would make the description become:
  ```
  "Creating GH issues #1 #2 #3 $"
  ```
(Notice how everything is truncated after the last detected shortcut).

- Example 3:
  ```
  "Creating GH issues #1 #2 #3 $@project"
  ```
  Tag suggestions will appear, and they will be based on a query **"3 "** (or **"3 @project"**, I'm not sure).
  Tapping on a suggestion would make the description become:
  ```
  "Creating GH issues #1 #2 "
  ```
(Notice how everything is truncated after the last detected shortcut).

- Example 4:
  ```
  "Creating GH issues$ #1 #2 #3 @project"
  ```
No tag or project suggestions will appear.

## Definition of done
- [ ] We've decided how the `cursorPosition` should be used to provide the query used to generate suggestions. Check Example 2 and 3 from the Expected/Desired behaviour section.
- [ ] We've fixed the tests for all actions related to taps on suggestions 
- [ ] We've added tests to exemplify the behaviour agreed upon
- [ ] The iOS issues are updated and their logic/tests matches the Android counterparts

heytherewill:
### Summary
This fixes the tiny oopsies in our autocomplete replacement behavior

### Technical considerations
We were using the token position to replace only a few words and we were parsing the string in a way that wouldn't allow the user to type emails as part of the project. This is also inconsistent with the behavior of the current app. Now we are looking for @s only at the beginning of words, which is how the current app parses things.

The plural of regex is regrets, so let me know if you have a better way

### Relationships
Closes #314

### Pull Request checklist

### Pointers
Go at it commit by commit. Also feel free to suggest more tests if you think something is not clear.

### Review template

## Merge permissions
Only myself


timer/src/main/java/com/toggl/timer/startedit/util/TextUtils.kt
```diff
@@ -1,19 +1,18 @@
 package com.toggl.timer.startedit.util
 
-fun String.lastSubstringFromTokenToPosition(token: Char, position: Int = this.length): Pair<Char?, String> =
-    this.lastSubstringFromAnyTokenToPosition(charArrayOf(token), position)
+fun String.findTokenAndQueryMatchesForAutocomplete(token: Char, cursorPosition: Int = this.length): Pair<Char?, String> =
```
daividssilverio, line 3: Do you mind adding a comment explaining what's going on with this regex? It's not obvious.

PR #325, opened 2020-05-22
Create the calendar items selector & translate calendar layout calculator

### LINKED ISSUE #268: Create the visible items selector
## Description

We need to add a selector to merge time entries and calendar events into a collection of group of calendar items. These items are grouped in a way they don't overlap with each other when rendering them in a collection view.
The selector needs to also filter the time entries and calendar events for the selected date.

## Definition of done
- [x] The CalendarItem enum is added with two cases, .timeEntry and .calendarEvent
- [x] The selector returns items for the selected date only
- [x] The logic to group items is translated from the CalendarLayoutCalculator in the old app
- [x] Tests are created for the selector

Depends on #267

daividssilverio:
### Summary
This PR does two things as requested on #268:
- Translates the [calendar layout attributes calculator from the old app](https://github.com/toggl/mobileapp/blob/develop/Toggl.Core.UI/Calendar/CalendarLayoutCalculator.cs)
- Adds the calendarItemSelector

### Technical considerations
None that I can remember now.

Feel free to make suggestions. This is the time. I'd suggest that iOS does splits this work into two PRs when they do it @juanlaube .

### Relationships
Closes #268

### Pull Request checklist

### Pointers
Read the tests mate.

### Review template

## Merge permissions
🦑 it


calendar/src/main/java/com/toggl/calendar/ui/CalendarLayoutCalculator.kt
```diff
@@ -0,0 +1,175 @@
...
+    /**
+     * Calculates the layout attributes for the indexed calendar items in a bucket.
+     * The calculation is done minimizing the number of columns.
+     *
+     * @param bucket
+     * @return An list of indexed calendar attributes
+     */
+    private fun calculateLayoutAttributesFor(bucket: CalendarItemGroup): List<Pair<Int, CalendarItemLayoutAttributes>> {
+        val left = bucket.filter { indexedItem -> indexedItem.second is CalendarItem.CalendarEvent }.toMutableList()
+        val right = bucket.filter { indexedItem -> indexedItem.second is CalendarItem.TimeEntry }.toMutableList()
```
semanticer, line 69:
```suggestion
        val left = bucket.filter { (_, item) -> item is CalendarItem.CalendarEvent }.toMutableList()
        val right = bucket.filter { (_, item)  -> item is CalendarItem.TimeEntry }.toMutableList()
```

This is always preferable to `.first` or `.second` for readability reasons imo.

calendar/src/main/java/com/toggl/calendar/ui/CalendarLayoutCalculator.kt
```diff
@@ -0,0 +1,175 @@
...
+
+import com.toggl.environment.services.time.TimeService
+import org.threeten.bp.Duration
+import org.threeten.bp.OffsetDateTime
+import org.threeten.bp.temporal.ChronoUnit
+import javax.inject.Inject
+import kotlin.math.absoluteValue
+
+typealias CalendarItemGroup = MutableList<Pair<Int, CalendarItem>>
+typealias CalendarItemGroups = MutableList<CalendarItemGroup>
```
semanticer, line 11: I would much rather see this whole thing written without mutable lists - especially because the rest of the code is nicely `functional` but of course there is some minor performance cost that comes with it.

calendar/src/main/java/com/toggl/calendar/ui/CalendarLayoutCalculator.kt
```diff
@@ -0,0 +1,175 @@
+package com.toggl.calendar.ui
```
heytherewill, line 1: This layout calculator class is exactly what should be done in the selector: Transforming calendar items in something the UI can display. Why is it split in two files when the other one is so tiny?

calendar/src/main/java/com/toggl/calendar/ui/CalendarItem.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import com.toggl.environment.services.calendar.CalendarEvent
+import com.toggl.models.domain.TimeEntry
+import org.threeten.bp.Duration
+import org.threeten.bp.OffsetDateTime
+
+typealias TimeEntryItem = TimeEntry
+typealias CalendarEventItem = CalendarEvent
+
+sealed class CalendarItem {
```
semanticer, line 11: I just want to highlight the fact that you've put `CalendarItem` and `CalendarItemsSelector` into UI package whereas `TimeEntriesLogSelector` and `TimeEntryViewModel` are in Domain package. I don't have a strong opinion on where it should be but we should be consistent.

calendar/src/test/java/com/toggl/calendar/ui/CalendarItemsSelectorKtTest.kt
```diff
@@ -0,0 +1,349 @@
...
+                )
+            }
+
+            val selectedTimeEntries = calendarItemsSelector(
+                (expectedTimeEntries + notExpectedTimeEntries).associateBy { it.id },
+                emptyList(),
+                date
+            )
+
+            assertThat(selectedTimeEntries).size().isEqualTo(expectedTimeEntries.size)
```
semanticer, line 54: I thought that we are still using kotest matchers

calendar/src/main/java/com/toggl/calendar/ui/CalendarLayoutCalculator.kt
juanlaube: I agree with William, I haven't considered this in the issue, but we should merge the properties from the layout attribute class (i.e: total columns and column index) into the calendar item.

daividssilverio:
The only detail I don't like about having the selector as just a function is passing the `timeService` around through the functions... it looks & feels like pure noise.

calendar/src/test/java/com/toggl/calendar/ui/CalendarItemsSelectorKtTest.kt
daividssilverio: Did we make a decision about that? I've added `assertJ` a while back for db tests #196 and the helper functions are quite handy. Those apis are all very similar either way.

calendar/src/main/java/com/toggl/calendar/ui/CalendarItem.kt
daividssilverio: I don't know what's the best place to put it.
It makes sense for the calendar item to be in the UI, but it also makes sense to leave it in the domain because I think we'll be using this class in the reducers (maybe not now that we'll be merging the `CalendarLayoutAttributes` and the `CalendarItem`).

calendar/src/main/java/com/toggl/calendar/ui/CalendarLayoutCalculator.kt
daividssilverio: Not sure if the added performance cost & effort is worth it. This is code that will be used during UI updates... 

But, no actual tests were made to see if it would make much of a difference. I'm just making assumptions here. If you really want that, I can rewrite it and create an issue to check the calendar performance during layout updates.

calendar/src/main/java/com/toggl/calendar/ui/CalendarLayoutCalculator.kt
```diff
@@ -0,0 +1,175 @@
...
+
+typealias CalendarItemGroup = MutableList<Pair<Int, CalendarItem>>
+typealias CalendarItemGroups = MutableList<CalendarItemGroup>
+
+class CalendarLayoutCalculator @Inject constructor(private val timeService: TimeService) {
+
+    private val offsetFromNow: Duration = Duration.ofMinutes(7)
+    private val minimumDurationForUIPurposes: Duration = Duration.ofMinutes(15)
+
+    fun calculateLayoutAttributes(calendarItems: List<CalendarItem>): List<CalendarItemLayoutAttributes> {
```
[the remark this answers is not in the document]
daividssilverio: These docs were added just to keep us sane trying to understand the algorithm used to calculate the calendar attributes. We don't have a policy about docs AFAIK.

calendar/src/main/java/com/toggl/calendar/ui/CalendarLayoutCalculator.kt
semanticer: Even though I'm pretty sure the performance difference would be almost no-existent, we can keep it as it is, It's a very minor thing.

calendar/src/test/java/com/toggl/calendar/ui/CalendarItemsSelectorKtTest.kt
semanticer: I actually have no idea if we agreed on anything :-D I was just assuming this based on our latest tests. I like kotest matchers a bit more but since we have no plans to use kotest in the near future, it maybe makes more sense to get rid of the dependency completely.

daividssilverio: I don't mind keeping the dependency tbh. For assertion helpers, as long as it's readable, reliable, fast and they do what we need, I don't see why not keeping both.

calendar/src/main/java/com/toggl/calendar/ui/CalendarLayoutCalculator.kt
daividssilverio: Good suggestion. I'll apply everywhere else.

semanticer:
I would actually also prefer selector as full-blown class with dependencies like `timeService` injected in constructor and then provide a selector method

calendar/src/main/java/com/toggl/calendar/ui/CalendarItem.kt
semanticer: I now lean stightly towards our domain package as this really is domain logic imo

calendar/src/main/java/com/toggl/calendar/ui/CalendarLayoutCalculator.kt
daividssilverio: I understood the point of merging and keeping only the `CalendarItem` [as discussed internally](https://toggl.slack.com/archives/CD40QCV25/p1590148813116500).

I've left the calculator on its own class because it's really its own thing really.

calendar/src/main/java/com/toggl/calendar/ui/CalendarItem.kt
daividssilverio: I've moved everything to the domain package to keep the consistency.

daividssilverio:
@juanlaube @semanticer @heytherewill 

One thing that this PR isn't handling and should be added to the selector added here is:
- Handling the layout changes during edition (when the user drags to edit the calendar items). I think we should handle it in another issue for 🐎

calendar/src/main/java/com/toggl/calendar/domain/CalendarItemsSelector.kt
```diff
@@ -0,0 +1,26 @@
...
+        it.startTime.toLocalDate() == localDate && (endTime == null || endTime.toLocalDate() == localDate)
+    }
+    val filteredEvents = calendarEvents.filter {
+        val endTime = it.startTime + it.duration
+        it.startTime.toLocalDate() == localDate && (endTime == null || endTime.toLocalDate() == localDate)
+    }
+    return filteredTimeEntries
+        .values.map { CalendarItem.ofTimeEntry(it) }
+        .plus(filteredEvents.map { CalendarItem.ofCalendarEvent(it) })
+        .run { calendarLayoutCalculator.calculateLayoutAttributes(this) }
```
semanticer, line 25: What do you think about using method references here?
```suggestion
    return filteredTimeEntries.values
        .map(CalendarItem::ofTimeEntry)
        .plus(filteredEvents.map(CalendarItem::ofCalendarEvent))
        .let(calendarLayoutCalculator::calculateLayoutAttributes)
```

calendar/src/main/java/com/toggl/calendar/domain/CalendarItemsSelector.kt
```diff
@@ -0,0 +1,26 @@
...
+): List<CalendarItem> {
+    val localDate = date.toLocalDate()
+    val filteredTimeEntries = timeEntries.filterValues {
+        val endTime = it.startTime + it.duration
+        it.startTime.toLocalDate() == localDate && (endTime == null || endTime.toLocalDate() == localDate)
+    }
+    val filteredEvents = calendarEvents.filter {
+        val endTime = it.startTime + it.duration
+        it.startTime.toLocalDate() == localDate && (endTime == null || endTime.toLocalDate() == localDate)
+    }
```
semanticer, line 21: This better hides the unimportant details so the reader can focus on higher-level concepts and only dive deeper when necessary. Not sure about the `isOnDate` name though.
```suggestion
    val localDate = date.toLocalDate()
    fun isOnDate(startTime: OffsetDateTime, endTime: OffsetDateTime?) =
        startTime.toLocalDate() == localDate && (endTime == null || endTime.toLocalDate() == localDate)

    val filteredTimeEntries = timeEntries.filterValues { isOnDate(it.startTime, it.startTime + it.duration) }
    val filteredEvents = calendarEvents.filter { isOnDate(it.startTime, it.startTime + it.duration) }
```

calendar/src/main/java/com/toggl/calendar/domain/CalendarItem.kt
```diff
@@ -0,0 +1,49 @@
...
+        override val totalColumns: Int = 0
+    ) : CalendarItem()
+
+    companion object {
+        fun ofTimeEntry(timeEntry: TimeEntryItem): CalendarItem =
+            TimeEntry(timeEntry)
+
+        fun ofCalendarEvent(calendarEvent: CalendarEventItem): CalendarItem =
+            CalendarEvent(calendarEvent)
+    }
```
heytherewill, line 33: Why is this even needed? The syntax will be pretty much the same

calendar/src/main/java/com/toggl/calendar/domain/CalendarItemsSelector.kt
heytherewill: `startsOnDate` is a better name, since it better describes the logic used to include events/time entries in the calendar

calendar/src/main/java/com/toggl/calendar/domain/CalendarItemsSelector.kt
heytherewill: Yes, please

heytherewill: Here's another example of why we don't need the companion method. This will not compile since you need to `CalendarItem.Companion::ofTimeEntry` while using the sealed class itself you'd simply `CalendarItem::TimeEntry`

calendar/src/main/java/com/toggl/calendar/domain/CalendarLayoutCalculator.kt
```diff
@@ -0,0 +1,187 @@
...
+    fun calculateLayoutAttributes(calendarItems: List<CalendarItem>): List<CalendarItem> {
+        if (calendarItems.isEmpty())
+            return emptyList()
+
+        return calendarItems
+            .mapIndexed { index, calendarItem -> index to calendarItem }
+            .sortedBy { it.calendarItem.startTime() }
+            .fold(mutableListOf(), ::groupOverlappingItems)
+            .flatMap(::calculateLayoutAttributesFor)
+            .sortedBy { (index, _) -> index }
```
heytherewill, line 27: I don't get this line. Why do we sort by start time and then go back to the pre-sorting order? 🤔

calendar/src/main/java/com/toggl/calendar/domain/CalendarItemsSelector.kt
daividssilverio: Actually, `CalendarItem::TimeEntry` doesn't work, unless I'm missing something (at least not on the version of kotlin we are using).

We can either keep the companion object method and use `map(CalendarItem::ofTimeEntry)`or use `map { CalendarItem.TimeEntry(it) } `.

calendar/src/main/java/com/toggl/calendar/domain/CalendarItemsSelector.kt
daividssilverio: `startsOnDate` does not describe de actual behaviour here. We are filtering the time entries that **Start And End** on this date.

`isOnDate` or maybe `isOnTheSameDay` would describe it better.

calendar/src/main/java/com/toggl/calendar/domain/CalendarLayoutCalculator.kt
juanlaube: That’s not needed anymore probably, it was to preserve the same order from the data because the calendar items and their layout were in two different arrays, but now it's all in the same class.

daividssilverio: Thanks for clarifying @juanlaube.

daividssilverio: This changes things a lot. Oh boy.

PR #332, opened 2020-05-25
✨ Add ItemTapped action

### LINKED ISSUE #270: Add the ItemTapped action
## Description

Add the ItemTapped action and handle it in the reducer. This should simply set the selected item in the state.

## Definition of done
- [ ] The `selectedItem` is set in the state with an editable time entry
- [ ] Tests are created

semanticer:
### Summary
This PR implements ItemTapped action as described in #270 

### Technical considerations
Nothing really interesting

`EditableTimeEntry` and `EditableProject` were moved to `com.toggl.models.domain` because I needed to use them in the Calendar Module. Let me know if you find a better place for these two orphans.

Instead of using `Either<TimeEntry, CalendarEvent>` for selectedItem, I created dedicated sealed type `SelectedCalendarItem`

### Relationships

Closes #270

### Pull Request checklist

### Pointers
It's fairly simple PR but going commit by commit will make the reviewing more pleasant.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it


calendar/src/main/java/com/toggl/calendar/domain/CalendarState.kt
```diff
@@ -1,10 +1,12 @@
 package com.toggl.calendar.domain
 
 import arrow.optics.optics
+import com.toggl.models.domain.Task
 
 @optics
 data class CalendarState(
-    val toBeDeleted: String
+    val tasks: Map<Long, Task>,
```
juanlaube, line 8: Why do we need tasks here? 🤔

semanticer: I was probably in a very wierd state of mind when I wrote this 🤔

timer/src/main/java/com/toggl/timer/startedit/domain/ProjectTagChipSelector.kt
```diff
@@ -15,11 +15,11 @@ fun projectTagChipSelector(
...
-        val project = projects[editableTimeEntry.projectId] ?: throw ProjectDoesNotExistException()
-        yield(ChipViewModel.Project(project))
-    } else {
-        yield(ChipViewModel.AddProject(addProjectLabel))
+    editableTimeEntry.projectId.let { projectId ->
+        if (projectId != null) {
+            val project = projects[projectId] ?: throw ProjectDoesNotExistException()
+            yield(ChipViewModel.Project(project))
+        } else yield(ChipViewModel.AddProject(addProjectLabel))
     }
```
daividssilverio, line 23:
```suggestion
    editableTimeEntry.projectId?.let { projectId ->
        val project = projects[projectId] ?: throw ProjectDoesNotExistException()
        yield(ChipViewModel.Project(project))
    } ?: yield(ChipViewModel.AddProject(addProjectLabel))
```

The elvis operator looks alright for this case. Doesn't provide much of a readability improvement though.

timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
```diff
@@ -318,7 +318,11 @@ class StartEditReducer @Inject constructor(
 
     private fun StartEditState.modifyWithCreateProjectSuggestion(autocompleteSuggestion: AutocompleteSuggestion.CreateProject): StartEditState =
         StartEditState.editableTimeEntry.modify(this) {
-            it.copy(editableProject = EditableProject(name = autocompleteSuggestion.name, workspaceId = 1))
+            it.copy(editableProject = EditableProject(
+                name = autocompleteSuggestion.name,
+                workspaceId = 1
+            )
+            )
```
daividssilverio, line 325: 👀

timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
```diff
@@ -462,11 +466,13 @@ class StartEditReducer @Inject constructor(
...
-
-        val relativeDuration = duration ?: startTime.absoluteDurationBetween(now)
-        return startTime + relativeDuration
+        return startTime.let { startTime ->
+            if (startTime == null) now
+            else {
+                val relativeDuration = duration ?: startTime.absoluteDurationBetween(now)
+                startTime + relativeDuration
+            }
+        }
```
daividssilverio, line 475:
```suggestion
        return startTime?.let { startTime ->
                val relativeDuration = duration ?: startTime.absoluteDurationBetween(now)
                startTime + relativeDuration
            } ?: now
        }
```

Another opportunity 🤔 What is our standing on the elvis operator again? 🤔 Maybe we should discuss when we should use it 🤔

semanticer: I'm not against, especially when it helps reduce nesting.

timer/src/main/java/com/toggl/timer/startedit/domain/ProjectTagChipSelector.kt
```diff
@@ -15,17 +15,15 @@ fun projectTagChipSelector(
...
 
-    if (editableTimeEntry.projectId != null) {
-        val project = projects[editableTimeEntry.projectId] ?: throw ProjectDoesNotExistException()
+    editableTimeEntry.projectId?.let { projectId ->
+        val project = projects[projectId] ?: throw ProjectDoesNotExistException()
         yield(ChipViewModel.Project(project))
-    } else {
-        yield(ChipViewModel.AddProject(addProjectLabel))
-    }
+    } ?: yield(ChipViewModel.AddProject(addProjectLabel))
```
heytherewill, line 21: This change is completely unrelated to this PR. Also, using a null-coalescence PR for a side effect (yielding) reads really odd

semanticer: It is related. `EditableTimeEntry` and `EditableProject` were moved to a different module and that broke the smart cast magic here. 
Side effect yield is indeed suboptimal but my original alternative was maybe even worse 🙈

PR #333, opened 2020-05-25
Make selectors classes

heytherewill:
### Summary
This refactors our selectors to be proper classes

### Relationships
Follows-up #325

### Pull Request checklist

### Pointers
Go at it commit by commit

### Review template

- [ ] The code considers i18n

## Merge permissions
Only myself


heytherewill:
For now, this strategy of using raw strings to inject is worse because stringly typed things are never good. Let's keep an eye on Dagger and see if we can figure out something like what ButterKnife does where we can do `@Inject(R.string.add_tags)`

PR #353, opened 2020-05-27
🖼 Reducer decorators

### LINKED ISSUE #330: Reducer decorators
## Description
Add a way to decorate reducers with another reducer so they gain their behavior. Base reducers should include the needed state and an action to redirect decorator actions.

## Definition of done
- [ ] Add instance method on reducer to decorate them
- [ ] Create core time entries reducer
- [ ] Decorate all reducers that handle time entries with the time entries reducer
- [ ] Write tests for time entry reducer
- [ ] Fix and update all old tests
- [ ] Edit documentation

semanticer:
### Summary
This PR implements #330 it's roughly the equivalent of https://github.com/toggl/ios/pull/259/files (iOS) although it's 3 times as big for some reason 🤔 
 
### Technical considerations
The new TimeEntryReducer/Action/State and a bunch of effects were extracted up to the feature module so that they can be used across all of our features. I used this opportunity to make our TimeEntryActions a little more explicit than before with matched effects as well. We are now also using more `DTO`'s to make our repository requests less vague. 

Let me know if there is a better way

### Relationships

Closes #330

### Pull Request checklist

### Pointers
~Review commit by commit.~ There are probably too many commits now to go commit by commit comfortably. Instead, start with the `com.toggl.common.feature.timeentry` package. Then you can follow the `decorateWith` extension method and see where is the TimeEntryReducer behavior used and what kind of work is delegated there. 
 
Pay special attention to new and old tests. Functional tests for all time entry related actions would be also appreciated. 

### Review template

- [ ] The code considers i18n

## Merge permissions
Me when it's ready


repository/src/main/java/com/toggl/repository/Repository.kt
```diff
@@ -97,11 +99,8 @@ class Repository(
 
     override suspend fun loadTasks(): List<Task> = taskDao.getAll().map(DatabaseTask::toModel)
 
-    override suspend fun startTimeEntry(
-        workspaceId: Long,
-        description: String
-    ): StartTimeEntryResult {
-        return timeEntryDao.startTimeEntry(workspaceId, description, timeService.now())
+    override suspend fun startTimeEntry(startTimeEntryDTO: StartTimeEntryDTO): StartTimeEntryResult {
+        return timeEntryDao.startTimeEntry(startTimeEntryDTO.workspaceId, startTimeEntryDTO.description, timeService.now())
```
heytherewill, line 103: We need to start considering the start time of the running time entry

repository/src/main/java/com/toggl/repository/extensions/DomainToDtoMappingExtensions.kt
```diff
@@ -0,0 +1,38 @@
...
+// fun TimeEntry.toCreateDto() = CreateTimeEntryDTO(
+//     description = description,
+//     startTime = startTime,
+//     billable = billable,
+//     duration = duration,
+//     workspaceId = workspaceId,
+//     projectId = projectId,
+//     taskId = taskId,
+//     tagIds = tagIds
+// )
```
heytherewill, line 29: 👀

timer/src/main/java/com/toggl/timer/di/TimerModule.kt
```diff
@@ -91,26 +94,42 @@ class TimerModule {
...
         startEditReducer: StartEditReducer,
         runningTimeEntryReducer: RunningTimeEntryReducer,
         projectReducer: ProjectReducer
     ): TimerReducer {
 
         return combine(
-            timeEntriesLogReducer.pullback(
+            timeEntriesLogReducer.decorateWith(timeEntryReducer,
+                mapToLocalState = { TimeEntryState(it.timeEntries) },
+                mapToLocalAction = { if (it !is TimeEntriesLogAction.TimeEntryHandling) null else it.timeEntryAction },
```
daividssilverio, line 107: it reads a tiny better like this imo:
```suggestion
                mapToLocalAction = { (it as? TimeEntriesLogAction.TimeEntryHandling)?.timeEntryAction },
```

timer/src/main/java/com/toggl/timer/di/TimerModule.kt
```diff
@@ -91,26 +95,41 @@ class TimerModule {
...
         runningTimeEntryReducer: RunningTimeEntryReducer,
         projectReducer: ProjectReducer
     ): TimerReducer {
 
         return combine(
-            timeEntriesLogReducer.pullback(
+            timeEntriesLogReducer.decorateWith(timeEntryReducer,
+                mapToLocalState = { TimeEntryState(it.timeEntries) },
+                mapToLocalAction = { TimeEntryAction.fromTimeEntryActionHolder(it) },
+                mapToGlobalState = { globalState, localState -> globalState.copy(timeEntries = localState.timeEntries) }
```
heytherewill, line 109: Do you think properly named methods here would be helpful at all? Are we ever going to reuse them?
```suggestion
                mapToLocalState = TimeEntriesLogState.Companion::toTimeEntriesState,
                mapToLocalAction =TimeEntriesLogAction.Companion::toTimeEntriesAction,
                mapToGlobalState = TimeEntriesLogState.Companion::fromTimeEntriesState }
```

semanticer: I don't think this would make much difference :-(

timer/src/main/java/com/toggl/timer/startedit/domain/StartEditReducer.kt
```diff
@@ -99,7 +99,7 @@ class StartEditReducer @Inject constructor(
                 if (editableTimeEntry.shouldStart()) {
                     effect(
                         StartEditAction.TimeEntryHandling(
-                            StartTimeEntry(editableTimeEntry.toStartDto())
+                            StartTimeEntry(editableTimeEntry.copy(startTime = timeService.now()).toStartDto())
```
heytherewill, line 102: Shouldn't you check before overriding the start date? 🤔

semanticer: My license to write code should have been taken from me like 3 hours ago

semanticer: @heytherewill check out the current version without dumb runtime errors and overides.

PR #356, opened 2020-05-29
Update project to use AS4 & Enable desugaring & Remove 310

### LINKED ISSUE #195: Enable desugaring and remove ThreeTenABP
## Description
Read [here](https://jakewharton.com/d8-library-desugaring/) and [here](https://blog.stylingandroid.com/d8-desugaring/) to understand why, but the tl;dr is that ThreeTenABP is not needed because of desugaring.

## Dependencies
- [ ] Desugaring is enabled
- [ ] The ThreeTenABP library is removed

daividssilverio:
### Summary
This updates our project to work with AS4.

### Technical considerations
- The android gradle plugin was updated to 4.0.0
- The gradle version updated to 6.1.1
- The `buildSrc` folder had it `include` instruction removed from `settings.gradle` because we don't need to that anymore (and the project breaks otherwise).
The 
- The gradle script to replace our `google-services.json.template` secrets was updated because it also broke... Some behaviour changed in the newer gradle version. I actually had to fix it by trial and error as I didn't find any instructions or anything on Stackoverflow about the problem that was happening.
- The `BuildConfig` for the `com.toggl.common` project started to get duplicated after the update, which was fixed by changing the packages of the `common-android` and `common-feature` modules to `com.toggl.common.android` and `com.toggl.common.feature` respectively.

That's what I could to make it work.

### Relationships
Closes #195 

### Pull Request checklist
- [ ] The code does not hinder the app's performance (I'm not really sure, please help testing)
- [ ] The code does not introduce regressions (Gotta test it a bit more to be sure)

### Review template
Clean your project. Invalidate and Refresh the cache if necessary.
Run the app and see if everything still works :)
Try to see if there's any difference in build times.

## Merge permissions
🦑 it unless it doesn't work for someone in the team.
Ideally everyone on @toggl/androids tests it.


environment/src/main/java/com/toggl/environment/services/time/ThreeTenTimeService.kt
```diff
@@ -1,6 +1,6 @@
 package com.toggl.environment.services.time
 
-import org.threeten.bp.OffsetDateTime
+import java.time.OffsetDateTime
 
 class ThreeTenTimeService : TimeService {
```
heytherewill, line 5: This class name is wrong now

common-android/build.gradle
```diff
@@ -38,6 +39,7 @@ android {
 }
 
 dependencies {
+    coreLibraryDesugaring 'com.android.tools:desugar_jdk_libs:1.0.5'
```
heytherewill, line 42: Why not?
```suggestion
    coreLibraryDesugaring Libs.Android.desugarJdkLibs
```

semanticer: I have the same question @daividssilverio

daividssilverio: I forgot to do it 🙈

PR #377, opened 2020-05-29
Add the EmptyPositionLongPressed action

### LINKED ISSUE #271: Add the EmptyPositionLongPressed action
## Description

Add the EmptyPositionLongPressed action and handle it in the reducer. This should simply set the selected item in the state with an empty editable time entry.

## Definition of done
- [ ] The `selectedItem` is set in the state with an empty editable time entry that starts at the given date
- [ ] Tests are created

daividssilverio:
### Summary
This adds and handles `EmptyPositionLongPressed` to the calendar day reducer.

### Technical considerations
I'm just using a dummy way to get the `workspaceId`. Should we start to use the user default workspace id? iOS started doing that. Check https://github.com/toggl/ios/pull/313

Because it's simple enough. Feel free to make your suggestions.

### Relationships
Closes #271

### Pull Request checklist

### Pointers
Read the tests. Check the iOS version.

### Review template

## Merge permissions
🦑


daividssilverio:
@juanlaube We need an issue to calculate what should be actual duration set when tapping an empty space. It should fill in the gaps between time entries/events. I'm guessing that logic could be handled in this action's reducer logic.

PR #393, opened 2020-06-01
✨ Calendar dragging

### LINKED ISSUE #272: Add the StartTimeDragged action
## Description

Add the StartTimeDragged action and handle it in the reducer. This action should update the selected item with the new start time.

## Definition of done
- [ ] The start time in the selected item is updated when it's a time entry.
- [ ] Tests are created

### LINKED ISSUE #273: Add the StopTimeDragged action
## Description

Add the StopTimeDragged action and handle it in the reducer. This action should update the selected item with the new stop time.

## Definition of done
- [ ] The stop time in the selected item is updated when it's a time entry.
- [ ] Tests are created

### LINKED ISSUE #274: Add the TimeEntryDragged action
## Description

Add the TimeEntryDragged action and handle it in the reducer. This action should update the selected item's start time while keeping the same duration.

## Definition of done
- [ ] The start time in the selected item is updated when it's a time entry.
- [ ] Tests are created

semanticer:
### Summary
This PR adds three drag actions as described in three related issues.

### Technical considerations
This is mostly boilerplate but I also modified some common code, let me explain what and why:

I extracted `toEditableTimeEntry()` and `toCalendarEvent()` functionality for broather availibility outside of the `ContextualMenuReducer`

I added `fun <T> T.withoutEffects() = noEffect()` because I wanted to keep `mutateEditableTimeEntry` 'effect agnostic' for possible future use in different contexts and `.withoutEffects()` looks much better here than `} returnEffect noEffect()`. Overall I think it's a nice companion to our `mutateWithoutEffects` and `returnEffect`.

I added `val EditableTimeEntry.endTime`. This is an extension property that calculates end time based on the start time and duration properties. I believe this will be useful in many places where it's more convenient to think in terms of `start - end` rather than `start - duration`

Let me know

### Relationships

Closes #273 
Closes #274 
Closes #272 

### Pull Request checklist

### Pointers
Commit by commit works here, maybe start by looking at the tests though.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


calendar/src/main/java/com/toggl/calendar/common/domain/CalendarItem.kt
```diff
@@ -1,5 +1,8 @@
 package com.toggl.calendar.common.domain
 
+import com.toggl.calendar.exception.SelectedItemShouldBeAATimeEntryException
```
daividssilverio, line 3: Oh, there's a typo in this exception's name `SelectedItemShouldBeAATimeEntryException`. Would you fix it?

calendar/src/main/java/com/toggl/calendar/calendarday/domain/CalendarDayReducer.kt
```diff
@@ -36,6 +47,21 @@ class CalendarDayReducer @Inject constructor(private val calendarService: Calend
                 val workspaceId = this.defaultWorkspaceId()
                 copy(selectedItem = createEmptyTimeEntry(workspaceId, action.startTime))
             }
+            is CalendarDayAction.TimeEntryDragged -> state.mutateEditableTimeEntry {
+                it.copy(startTime = action.startTime)
+            }.withoutEffects()
+            is CalendarDayAction.StartTimeDragged -> state.mutateEditableTimeEntry {
+                val prevStartTime = it.startTime!!
+                val newStartTime = action.startTime
+                if (!newStartTime.isBefore(it.endTime())) throw TimeEntryShouldNotStartAfterEndTime()
```
daividssilverio: 🤔 Not sure if we should throw here. The view just needs to ignore invalid values. The user can try to drag the start time after the endTime, I think we should just ignore it. Besides that, why not use `isAfter`? inverted conditionals are a tiny worse to read in general.

calendar/src/main/java/com/toggl/calendar/calendarday/domain/CalendarDayReducer.kt
```diff
@@ -36,6 +47,21 @@ class CalendarDayReducer @Inject constructor(private val calendarService: Calend
...
+                val durationDiff = Duration.between(prevStartTime, newStartTime)
+                it.copy(startTime = newStartTime, duration = it.duration!!.minus(durationDiff))
+            }.withoutEffects()
+            is CalendarDayAction.StopTimeDragged -> state.mutateEditableTimeEntry {
+                if (action.stopTime.isBefore(it.startTime)) throw TimeEntryShouldNotEndBeforeStartTime()
```
daividssilverio, line 61: Same point about throwing here. Also ping @juanlaube about this 🤔

calendar/src/main/java/com/toggl/calendar/calendarday/domain/CalendarDayReducer.kt
```diff
@@ -48,5 +74,16 @@ class CalendarDayReducer @Inject constructor(private val calendarService: Calend
         )
     }
 
+    private fun MutableValue<CalendarDayState>.mutateEditableTimeEntry(modifyEditableTimeEntry: (EditableTimeEntry) -> EditableTimeEntry) {
+        val editableTimeEntry = this().selectedItem.toEditableTimeEntry()
+        editableTimeEntry.throwIfNew()
+        editableTimeEntry.throwIfRunning()
```
daividssilverio, line 80: We can drag a running time entry's start time, I don't think we should throw here.

calendar/src/test/java/com/toggl/calendar/calendarday/domain/StartTimeDraggedActionTests.kt
```diff
@@ -0,0 +1,144 @@
...
+        val initialState = createInitialState(selectedItem = SelectedCalendarItem.SelectedTimeEntry(timeEntry))
+        reducer.testReduceException(
+            initialState,
+            CalendarDayAction.StartTimeDragged(OffsetDateTime.MAX),
+            TimeEntryShouldNotBeNewException::class.java
+        )
+    }
+
+    @Test
+    fun `throws if executed on a running time entry`() = runBlockingTest {
```
daividssilverio, line 75: This should be totally fine.

daividssilverio: I mean, it shouldn't throw.

calendar/src/main/java/com/toggl/calendar/calendarday/domain/CalendarDayReducer.kt
juanlaube: I agree with Daivid, it shouldn't update the values instead of throwing.

calendar/src/main/java/com/toggl/calendar/calendarday/domain/CalendarDayReducer.kt
juanlaube: Similar to the previous comment, instead of throwing it should not update the values.

PR #400, opened 2020-06-03
Fix StartEditReducer done button logic

### LINKED ISSUE #399: Fix time entry updates
## Description
Currently our StartEditReducer doesn't handle time entry updates properly.
We are currently updating only the `description` and `billable` fields of the time entries during edition. 

## Definition of done
- [x] The `StartEditReducer` can start manual time entries, currently the `TimeEntryReducer` is able to handle that, but we only forward actions from the `StartEditReducer` to start time entries or edit existing ones
- [x] The `StartEditReducer` maps correctly all the fields from the `EditableTimeEntry` to a `TimeEntry` when handling `DoneButtonTapped`. Currently it only updates the `description` and the `billable` flag.
- [ ] Tests were added to verify the behaviour described above

heytherewill:
### Summary
This is a nice catch found by Des. Our creation logic was all wrong

### Relationships
Closes #399 

### Pull Request checklist

### Pointers
Review commit by commit

### Review template

## Merge permissions
Only myself


common-feature/src/main/java/com/toggl/common/feature/timeentry/extensions/EditableTimeEntryExtensions.kt
```diff
@@ -9,6 +9,8 @@ fun EditableTimeEntry.isNew() = this.startTime == null
 fun EditableTimeEntry.isRunning() = this.duration == null
 fun EditableTimeEntry.isRunningOrNew() = isRunning() || isNew()
 fun EditableTimeEntry.isStopped() = this.startTime != null && this.duration != null
+fun EditableTimeEntry.isRepresentingGroup() = this.ids.size > 1
+fun EditableTimeEntry.needsPersisting() = this.ids.isEmpty()
```
daividssilverio, line 13: `needsToBeCreated`, `wasNotPersisted`,  or something that gives the idea of "creation" could make more sense.
I say that because time entries that are being updated also need to be persisted. It read weird to me.

heytherewill: Fair point

PR #410, opened 2020-06-04
🎨 Add color value picker

### LINKED ISSUE #383: Create the Value control
## Description
Implement the Value control as a `SeekBar` subclass and connect it to the state. Designs are here: https://toggl.invisionapp.com/d/main#/console/20014282/419193923/preview, while implementation specs can be found here: https://www.notion.so/toggl/Color-Picker-Unification-fe79a56649d74175a1b004936c7ebe10#1a1dc1f21c7f44f89adc95b666248458
TLDR: Values go from 30 to 100

## Dependencies
Depends on #381

## Definition of done
- [ ] The Value control is implemented as a `SeekBar` subclass
- [ ] The Value control parses the V values of the currently set custom colour and sets its thumb to the appropriate coordinates
- [ ] The control's thumb is coloured into the currently set custom colour
- [ ] The control's max value tint gets the HS value of the currently set custom colour
- [ ] Moving the control's thumb emits a `.colorPicked` action with a new colour with the new V value

semanticer:
### Summary
This PR implements #383 or at least part of it. I'm now 100% sure about the scope here. There is more work to be done in the reducer and state but I have to double-check if it's not a part of some other issue. 

### Technical considerations
Nothing really interesting

Let me know if you know about something simpler

### Relationships

Closes #383

### Pull Request checklist

### Pointers
Does the color math make sense? 

### Review template

- [ ] The code considers i18n

## Merge permissions
Me


timer/src/main/java/com/toggl/timer/project/ui/ColorHueSaturationPicker.kt
```diff
@@ -0,0 +1,130 @@
...
+
+        onHueSaturationChangedListener?.invoke(hue, saturation)
+
+        invalidate()
+    }
+
+    override fun draw(canvas: Canvas) {
+        super.draw(canvas)
+
+        fun Canvas.drawGradientBackground() {
```
daividssilverio, line 104: why nest it?

timer/src/main/java/com/toggl/timer/project/ui/ColorHueSaturationPicker.kt
```diff
@@ -0,0 +1,130 @@
...
+        hue = pointX / width
+        saturation = complement(pointY / height)
+
+        onHueSaturationChangedListener?.invoke(hue, saturation)
+
+        invalidate()
+    }
+
+    override fun draw(canvas: Canvas) {
+        super.draw(canvas)
```
daividssilverio, line 102: Why `draw` and not `onDraw`? Usually we should override `onDraw` when creating custom views 🤔 At least that's what the [docs](https://developer.android.com/reference/android/view/View#draw(android.graphics.Canvas)) says

timer/src/main/java/com/toggl/timer/project/ui/ColorHueSaturationPicker.kt
```diff
@@ -0,0 +1,130 @@
...
+
+        fun Canvas.drawGradientBackground() {
+            gradientBackgroundPaint.shader =
+                LinearGradient(0f, 0f, width.toFloat(), 0f, colorGradientColors, null, Shader.TileMode.CLAMP)
```
daividssilverio, line 106: Do you need to create a new `LinearGradient` for every draw call?

timer/src/main/java/com/toggl/timer/project/ui/ColorHueSaturationPicker.kt
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    override fun onTouchEvent(e: MotionEvent): Boolean =
+        when (e.action) {
+            MotionEvent.ACTION_MOVE,
+            MotionEvent.ACTION_DOWN -> {
+                updateLocationTo(e.x, e.y)
+                true
+            }
+            else -> super.onTouchEvent(e)
```
daividssilverio, line 85:
```suggestion
            MotionEvent.ACTION_DOWN -> {
                parent.requestDisallowInterceptTouchEvent(true)
                updateLocationTo(e.x, e.y)
                true
            }
            MotionEvent.ACTION_UP -> {
                parent.requestDisallowInterceptTouchEvent(false)
                true
            }
            else -> super.onTouchEvent(e)
```

Seems to be enough to prevent the bottomsheet parent from consuming touches while this view is handling them.
I think we can use this idea on the wheel as well 🤔

You might need to handle `ACTION_CANCEL` as well to give the control back to the parent.

semanticer: Looks good works well 👍

timer/src/main/java/com/toggl/timer/project/ui/ColorHueSaturationPicker.kt
semanticer: Not sure how this happened but it should be onDraw for sure 👍

timer/src/main/java/com/toggl/timer/project/ui/ColorHueSaturationPicker.kt
semanticer: The idea here was to provide two different levels of complexity so that you can first look at the higher level:
```kotlin
with(canvas) {
    clipPath(roundedCornerClipPath)
    drawGradientBackground()
    drawThumbCircle()
}
```
And then choose to go into the detail of some particular method if necessary.

timer/src/main/java/com/toggl/timer/project/domain/ProjectAction.kt
```diff
@@ -13,6 +13,9 @@ sealed class ProjectAction {
     object PrivateProjectSwitchTapped : ProjectAction()
     object CloseButtonTapped : ProjectAction()
     object DialogDismissed : ProjectAction()
+    data class ColorValueChanged(val value: Float) : ProjectAction()
+    data class ColorHueSaturationChanged(val hue: Float, val saturation: Float) : ProjectAction()
+    object CustomColorPicked : ProjectAction()
```
heytherewill, line 18: Is this even needed? 🤔

timer/src/main/java/com/toggl/timer/project/domain/ProjectReducer.kt
```diff
@@ -52,6 +53,13 @@ class ProjectReducer @Inject constructor(
                         it.copy(isPrivate = !it.isPrivate)
                     }
                 }
+            is ProjectAction.ColorValueChanged -> state.mutateWithoutEffects {
+                copy(customColor = customColor.copy(value = action.value))
+            }
+            is ProjectAction.ColorHueSaturationChanged -> state.mutateWithoutEffects {
+                copy(customColor = customColor.copy(hue = action.hue, saturation = action.saturation))
```
heytherewill, line 60: Shouldn't this be picked in real time. E.g.: This changes the custom color and then emit it? 🤔

timer/src/main/java/com/toggl/timer/project/ui/ProjectDialogFragment.kt
```diff
@@ -105,6 +104,7 @@ class ProjectDialogFragment : BottomSheetDialogFragment() {
 
         color_done_button.setOnClickListener {
             coloPickerVisibilityRequestFlow.value = false
+            store.dispatch(ProjectAction.CustomColorPicked)
```
heytherewill, line 107: Is this correct? I thought this was meant to simply end edition, not select the custom colors. Imagine the following:

- I have a basic color
- I open the picker
- I tap done

My color has now changed. Also, right now the done button is visible even in the basic picker, so this yields an illegal state. Please clear things up with Milad

timer/src/main/java/com/toggl/timer/project/ui/ProjectDialogFragment.kt
```diff
@@ -197,8 +212,7 @@ class ProjectDialogFragment : BottomSheetDialogFragment() {
                 coloPickerVisibilityRequestFlow.value = false
             }
             is ColorViewModel.CustomColor -> {
-                // TODO Dispatch colorPicked using the values of the hs and v picker controls
-                store.dispatch(ProjectAction.ColorPicked(Constants.DefaultCustomColor.hex))
+                store.dispatch(ProjectAction.CustomColorPicked)
```
heytherewill, line 215: This is a good use of `CustomColorPicker`, though arguably using the hex yields the same results (after all, we can check for non-basic colors in the `ColorPicked` event to see what users are picking

timer/src/main/java/com/toggl/timer/project/ui/ColorHueSaturationPicker.kt
```diff
@@ -0,0 +1,151 @@
...
+            shader = LinearGradient(0f, 0f, 0f, height.toFloat(), intArrayOf(Color.TRANSPARENT, Color.WHITE), null, Shader.TileMode.CLAMP)
+        }
+    }
+
+    private val gradientBackgroundPaint by lazy {
+        Paint().apply {
+            style = Paint.Style.FILL
+            flags = Paint.ANTI_ALIAS_FLAG
+            shader = LinearGradient(0f, 0f, width.toFloat(), 0f, intArrayOf(
+                Color.RED, Color.YELLOW, Color.GREEN, Color.CYAN, Color.BLUE, Color.MAGENTA, Color.RED
```
heytherewill, line 52: I'm also using this array in the list of colors, so maybe adding it as a constant called `HueRange` or something is a good idea

timer/src/main/java/com/toggl/timer/project/ui/ColorHueSaturationPicker.kt
daividssilverio: I see. I'm not opposed to that. It looks ok here.
But I do find it harder to get the gist of what's going on a callback when there's functions defined before what's actually run. Similar to our `TimeEntriesLogSelector`, I think it is mildly uncomfortable to read imho.

timer/src/main/java/com/toggl/timer/project/ui/ColorHueSaturationPicker.kt
```diff
@@ -0,0 +1,157 @@
...
+            ), null, Shader.TileMode.CLAMP)
+        }
+    }
+
+    private val roundedCornerClipPath by lazy {
+        Path().apply {
+            val radius = 30.0f
```
daividssilverio, line 62: Yeah, I think this should be a themed attribute, or at least based on `dps`

daividssilverio:
The color palette drawables are all broken on android 5. I'd just use an `ImageView` and set the tint. This is probably happening in other places in the app as well 🤔 
image

timer/src/main/java/com/toggl/timer/project/ui/ColorHueSaturationPicker.kt
semanticer: I couldn't extract this to constants as they are not android module. So I used our color extensions file instead for now.

PR #411, opened 2020-06-04
Draw calendar shapes

### LINKED ISSUE #358: Draw the calendar shape
## Description
This needs to add the functions to draw a calendar item shape

## Dependencies
Depends on #357

## Definition of done
- [x] it takes in a rectF and the calendarItem to draw the shape
- [x] stopped time entry
- [x] running time entry (dashed background, dashed border after "now")
- [x] calendar time entry (calendar icon)
- [ ] edition handles (top & bottom on stopped, top on running, none on calendar)
- [x] draw one item on BaseCalendarWidgetView for now

daividssilverio:
### Summary
This updates the `CalendarItemDrawingDelegate` to actually draw calendar shapes.
This includes everything described on #358

The only thing missing at the moment are the editing handles.

### Technical considerations
This is mostly a straight port from the old implementation.
I'm accepting suggestions to improve it and make it more kotlin-y.

Let's not make early performance improvements though. But I've thought about some... like pre-calculating the possible calendar events widths with different columns numbers.

### Relationships
Closes #358

### Pull Request checklist

### Pointers
You may manually set calendar items in the list of calendars to see how they look.
Change the `scrollOffset` manually to scroll the view.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it after the draft is ready


calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarItemDrawingDelegate.kt
```diff
@@ -1,21 +1,337 @@
...
+            if (calendarItem is CalendarItem.CalendarEvent) calendarIconSize - calendarIconRightInsetMargin else 0f
+        val eventHeight = drawingRect.height()
+        val eventWidth = drawingRect.width() - textLeftPadding
+        val fontSize =
+            if (eventHeight <= shortCalendarItemHeight) shortCalendarItemFontSize else regularCalendarItemFontSize
+        var textVerticalPadding =
+            if (eventHeight <= shortCalendarItemHeight) shortCalendarItemVerticalPadding else regularCalendarItemVerticalPadding
+        textVerticalPadding = kotlin.math.min(eventHeight - fontSize, textVerticalPadding)
+        val textHorizontalPadding =
+            if (eventHeight <= shortCalendarItemHeight) shortCalendarItemHorizontalPadding else regularCalendarItemHorizontalPadding
```
semanticer, line 95:
```suggestion
        val isShort = eventHeight <= shortCalendarItemHeight
        val fontSize = if (isShort) shortCalendarItemFontSize else regularCalendarItemFontSize
        var textVerticalPadding = if (isShort) shortCalendarItemVerticalPadding else regularCalendarItemVerticalPadding
        textVerticalPadding = kotlin.math.min(eventHeight - fontSize, textVerticalPadding)
        val textHorizontalPadding = if (isShort) shortCalendarItemHorizontalPadding else regularCalendarItemHorizontalPadding
```

Just a small suggestion in the name of DRY. But maybe you could come up with some data class that holds all of these params and you could construct it with just one if statement 🤔

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarItemDrawingDelegate.kt
```diff
@@ -1,21 +1,337 @@
...
+        var totalLineHeight = 0
+        for (i in 0 until eventTextLayout.lineCount) {
+            val lineBottom = eventTextLayout.getLineBottom(i)
+            if (lineBottom <= eventHeight) {
+                totalLineHeight = lineBottom
+            }
+        }
+
+        return totalLineHeight
+    }
```
semanticer, line 124:
```suggestion
    private fun calculateLineHeight(eventHeight: Float, eventTextLayout: StaticLayout): Int = 
        (0 until eventTextLayout.lineCount)
            .map(eventTextLayout::getLineBottom)
            .filter { it <= eventHeight }
            .max() ?: 0
```

This should be a var-less equivalent right?

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarItemDrawingDelegate.kt
```diff
@@ -1,21 +1,337 @@
...
+
+    private fun getCalendarItemTextLayout(
+        item: CalendarItem,
+        eventWidth: Float,
+        fontSize: Float,
+        isRunning: Boolean
+    ): StaticLayout {
+        val eventTextLayout = textLayouts[item.description()]
+        if (eventTextLayout != null && !(abs(eventTextLayout.width - eventWidth) > 0.1) && eventTextLayout.text == item.description())
```
semanticer, line 133:
```suggestion
        if (eventTextLayout != null && abs(eventTextLayout.width - eventWidth) <= 0.1 && eventTextLayout.text == item.description())
```
Without negation

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarItemDrawingDelegate.kt
```diff
@@ -1,21 +1,337 @@
...
+        if (primaryTextColorContrast >= minimumTextContrast) return primaryTextColor
+
+        val whiteContrast =
+            ColorUtils.calculateContrast(Color.WHITE, lastCalendarItemBackgroundColor)
+        return if (whiteContrast >= minimumTextContrast) Color.WHITE else Color.BLACK
+    }
+
+    private var lastCalendarItemBackgroundColor: Int = 0
+
+    private fun drawCalendarEventItem(canvas: Canvas, calendarItem: CalendarItem.CalendarEvent) {
```
semanticer, line 181: How about writing all of these `draw*` methods as calendar extension functions? Then we could call these as:  `canvas.drawCalendarEventItem(calendarItem)`

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarItemDrawingDelegate.kt
```diff
@@ -1,21 +1,337 @@
...
+
+    private fun drawShapeBaseBackgroundFilling(canvas: Canvas, color: Int) {
+        itemPaint.color = color
+        itemPaint.style = Paint.Style.FILL_AND_STROKE
+        canvas.drawRoundRect(drawingRect, cornerRadius, cornerRadius, itemPaint)
+    }
+
+    private fun drawShapeBackgroundStripes(canvas: Canvas, color: Int) {
+        canvas.save()
+        canvas.clipRect(drawingRect)
```
semanticer, line 235: It looks like every time you use save & restore there is some ktx extension to help you eliminate these. In this case, I think 
`canvas.withClip(stripeRect) { }` should be what you are looking for.

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarItemDrawingDelegate.kt
```diff
@@ -1,21 +1,337 @@
...
+
+    private fun drawSolidBorder(canvas: Canvas, color: Int) {
+        itemPaint.style = Paint.Style.STROKE
+        itemPaint.color = color
+        itemPaint.strokeWidth = runningTimeEntryBorderStrokeWidth
+        canvas.drawRoundRect(drawingRect, cornerRadius, cornerRadius, itemPaint)
+    }
+
+    private fun drawBottomDashedBorder(canvas: Canvas, color: Int) {
+        canvas.save()
```
semanticer, line 267: Another opportunity to kill save & restore with ktx

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarItemDrawingDelegate.kt
```diff
@@ -1,21 +1,337 @@
...
+
+    private fun calculateItemRect(calendarItem: CalendarItem, calendarBounds: RectF) {
+        val totalItemSpacing = (calendarItem.totalColumns - 1) * itemSpacing
+        val availableWidth = calendarBounds.width() - leftMargin
+        val eventWidth = (availableWidth - leftPadding - rightPadding - totalItemSpacing) / calendarItem.totalColumns
+        val left = leftMargin + leftPadding + eventWidth * calendarItem.columnIndex + calendarItem.columnIndex * itemSpacing
+        val startTime = calendarItem.startTime()
+        val totalHours = startTime.hour + startTime.minute.toFloat() / 60f
+        val durationHours = duration(calendarItem).toMinutes().toFloat() / 60f
+        val durationInHeight = kotlin.math.max(minHourHeight, durationHours * currentHourHeight)
```
semanticer, line 319: Let's import max and min functions

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarItemDrawingDelegate.kt
```diff
@@ -1,21 +1,337 @@
...
+    private val rightPadding: Float,
+    private val cornerRadius: Float,
+    private val calendarEventBottomLineHeight: Float,
+    private val normalCalendarIconBitmap: Bitmap,
+    private val smallCalendarIconBitMap: Bitmap,
+    private val calendarBackgroundColor: Int,
+    private val runningTimeEntryThinStripeWidth: Float,
+    private val runningTimeEntryStripesSpacing: Float,
+    private val runningTimeEntryBorderStrokeWidth: Float,
+    private val runningTimeEntryDashedHourTopPadding: Float,
```
semanticer, line 42: It looks like some of these parameters are closely related. Maybe we could decrease the number of parameters by using data classes that would hold these params together. This could also improve readability if done carefully. On the other hand, this could impact performance so I guess I will just leave this for your consideration.

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarWidgetView.kt
```diff
@@ -273,6 +294,41 @@ class CalendarWidgetView @JvmOverloads constructor(
...
+            shortCalendarItemHeight = shortCalendarItemHeight,
+            shortCalendarItemHorizontalPadding = shortCalendarItemHorizontalPadding,
+            shortCalendarItemVerticalPadding = shortCalendarItemVerticalPadding
+        ).apply {
+            currentHourHeight = calendarBaseHourHeight
+        }
+        backgroundDrawingDelegate = CalendarBackgroundDrawingDelegate()
     }
 
     override fun onDraw(canvas: Canvas) {
```
semanticer, line 334: Eliminate save and restore with KTX and consider changing drawCalendarBackground and drawCalendarItems to canvas extensions methods

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarItemDrawingDelegate.kt
```diff
@@ -1,21 +1,337 @@
...
+        val totalLineHeight = calculateLineHeight(eventHeight, eventTextLayout).toFloat()
+
+        canvas.save()
+        canvas.translate(
+            drawingRect.left + textHorizontalPadding + textLeftPadding,
+            drawingRect.top + textVerticalPadding
+        )
+        canvas.clipRect(0f, 0f, eventWidth - textHorizontalPadding, totalLineHeight)
+        eventTextLayout.draw(canvas)
+        canvas.restore()
```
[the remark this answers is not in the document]
daividssilverio: I saw these extensions 🤔 But since I'm clipping as well, I thought it wasn't worth the extra nest. Maybe I can add my own 🤔 `withTranslationAndClip`?

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarItemDrawingDelegate.kt
daividssilverio: Yeah, we can make it shorter with `maxBy` as well. I'll add this.

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarItemDrawingDelegate.kt
daividssilverio: Those are configuration parameters and are set only once. I think adding a data class for it wouldn't hurt.
I'll try to clean it up before finishing this PR.

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarItemDrawingDelegate.kt
daividssilverio: I'm not opposed to that 👍 It would read nicely :)

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarItemDrawingDelegate.kt
daividssilverio: Yeah, there's a bunch. I don't like the extra nesting TBH, but it is less error prone 🤔 .

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarItemDrawingDelegate.kt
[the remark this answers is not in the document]
daividssilverio: Maybe:
```kotlin
private inline fun Canvas.with(
        translationX: Float = 0.0f,
        translationY: Float = 0.0f,
        clipLeft: Float,
        clipTop: Float,
        clipRight: Float,
        clipBottom: Float,
        block: Canvas.() -> Unit
    ) {
        val checkPoint = save()
        translate(translationX, translationY)
        clipRect(clipLeft, clipTop, clipRight, clipBottom)
        try {
            block()
        } finally {
            restoreToCount(checkPoint)
        }
    }
```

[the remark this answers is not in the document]
daividssilverio: it would look like:
```kotlin
canvas.with(
            translationX = drawingRect.left + textHorizontalPadding + textLeftPadding,
            translationY = drawingRect.top + textVerticalPadding,
            clipLeft = 0f,
            clipTop = 0f,
            clipRight = eventWidth - textHorizontalPadding,
            clipBottom = totalLineHeight
        ) {
            eventTextLayout.draw(canvas)
        }
```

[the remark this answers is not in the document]
semanticer: Yeah, maybe it is. I wouldn't go with the custom implementation. Feel free to leave it as it was then.

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarItemDrawingDelegate.kt
semanticer: I think a one level nesting is definitely worth it, it's much nicer when you don't need to think about save & restore.

daividssilverio: I've made the change for sanity's sake.

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarItemDrawingDelegate.kt
daividssilverio: Hey @semanticer I didn't go through with creating data classes to simplify this simply because I rather get it working first. I'll create an issue to come back at this and handle this technical dept.

daividssilverio: #414

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarWidgetViewDrawingData.kt
```diff
@@ -0,0 +1,29 @@
+package com.toggl.calendar.calendarday.ui.views
+
+import com.toggl.calendar.common.domain.CalendarItem
+import com.toggl.calendar.common.domain.SelectedCalendarItem
+
+data class CalendarWidgetViewDrawingData(
```
daividssilverio, line 6: @semanticer what do you think about this addition? 
I've added this to keep changes made to anything the widget uses to draw atomic.
We had a bit of trouble with this when we were doing the `Wheel`, so I'm trying to avoid the same trouble from before.

@juanlaube (how are you doing this on iOS?)

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarWidgetView.kt
```diff
@@ -111,10 +98,12 @@ class CalendarWidgetView @JvmOverloads constructor(
             val defaultEditingHoursLabelPaintTextSize = getDimension(R.dimen.default_calendar_editingHoursLabelPaintTextSize)
             val defaultEditingHandlesHorizontalMargins = getDimension(R.dimen.default_calendar_editingHandlesHorizontalMargins)
             val defaultEditingHandlesRadius = getDimension(R.dimen.default_calendar_editingHandlesRadius)
+            val defaultEditingHandlesStrokeWidth = getDimension(R.dimen.default_calendar_editingHandlesStrokeWidth)
             val defaultCalendarEventsStartMargin = getDimension(R.dimen.default_calendar_calendarEventsStartMargin)
             val defaultCalendarEventsLeftPadding = getDimension(R.dimen.default_calendar_calendarEventsLeftPadding)
             val defaultCalendarEventsRightPadding = getDimension(R.dimen.default_calendar_calendarEventsRightPadding)
             val defaultCalendarEventsItemsSpacing = getDimension(R.dimen.default_calendar_calendarEventsItemsSpacing)
+            val defaultCalendarEventsCornerRadius = getDimension(R.dimen.default_calendar_calendarEventsCornerRadius)
```
semanticer, line 106: This is really up to you but if you want to make the code a bit shorter and you don't care that much about `CalendarWidgetView`'s resilience, you can get rid off all these defaults here and use throw variants of TypedArray getters like [getDimensionOrThrow](https://developer.android.com/reference/kotlin/androidx/core/content/res/package-summary#(android.content.res.TypedArray).getDimensionOrThrow(kotlin.Int) )
and we would simply make sure to add all of these parameters in XML.

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarWidgetViewDrawingData.kt
semanticer: It does make sense. The only thing I don't love is that you provide two ways to query selectedItem, either via `selectedCalendarItem` property or `selectedItemIndex` and each can give you different info in some cases. This can be improved to explicitly expose only one way to query selectedItem and also making it easier by providing instance itself rather than index only. 

```kotlin
data class CalendarWidgetViewDrawingData(
    val items: List<CalendarItem> = emptyList(),
    private val itemToBeSelected: SelectedCalendarItem? = null
) {
    val selectedItem = items.find { it.isSelectedItem(itemToBeSelected) }
}
```

The `itemToBeSelected` name is not perfect though 🤔

daividssilverio: The detail is that I was using that index first draw the calendar items that are not in "edit mode" and then the selected calendar item. Although I currently don't see a reason why not simply split the list here anyway 🤔

```kotlin
data class CalendarWidgetViewDrawingData(
    sourceItems: List<CalendarItem> = emptyList(),
    itemToBeSelected: SelectedCalendarItem? = null
) {
    val selectedItem = sourceItems.find { it.isSelectedItem(itemToBeSelected) }
    val calendarItems = items.filter { not selectedItem } 
}
```

What do you think?

calendar/src/main/res/values/colors.xml
```diff
@@ -6,4 +6,5 @@
     <color name="default_calendar_calendarHourTextColor">#C2C2C2</color>
     <color name="default_calendar_editingHourLabelTextColor">#FF328FFF</color>
     <color name="default_calendar_primaryTextColor">@color/text_on_background</color>
+    <color name="default_calendar_defaultCalendarItemColor">#C2C2C2</color>
```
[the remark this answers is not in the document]
daividssilverio: I was leaving that like that to simplify the port on my side :)
I can change it to 🐍 🔤 now to maintain the consistency.

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarWidgetView.kt
```diff
@@ -53,34 +52,22 @@ class CalendarWidgetView @JvmOverloads constructor(
...
-                description = "Test Event",
-                color = "#c2c2c2",
-                calendarId = "1"
-            ),
-            0,
-            1
-        )
-    )
-    var selectedItem: SelectedCalendarItem? = null
+    var drawingData: CalendarWidgetViewDrawingData = CalendarWidgetViewDrawingData()
```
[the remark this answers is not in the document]
daividssilverio: We'll eventually have to set this, not sure about the interface yet. I'll make it private for now.

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarWidgetView.kt
```diff
@@ -268,35 +279,80 @@ class CalendarWidgetView @JvmOverloads constructor(
                     )
                     calendarIconId = getResourceId(R.styleable.CalendarWidgetView_calendarIcon, defaultCalendarIconId)
                     primaryTextColor = getColor(R.styleable.CalendarWidgetView_primaryTextColor, defaultPrimaryTextColor)
+                    defaultCalendarItemColor = getColor(R.styleable.CalendarWidgetView_defaultCalendarItemColor, defaultDefaultCalendarItemColor)
                 } finally {
                     recycle()
                 }
```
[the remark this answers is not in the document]
daividssilverio: Cool o/

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarWidgetView.kt
daividssilverio: Unless you have a strong opinion about leaving it resilient, I rather keep it the way it is right now.

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarWidgetView.kt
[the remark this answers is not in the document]
daividssilverio: 🤔 Using the `use` doesn't provide the contract guarantees the `apply` extension does. Just using `use` makes the compiler think we didn't initialize the themed properties.
I would need to run something like:

```kotlin
typedArray.use { 
  it.apply  {
    ...
  } 
} 

Or right our own contract 🤔 

```kotlin
@ExperimentalContracts
inline fun <T> TypedArray.applyAndRecycle(block: TypedArray.() -> T): T {
    contract {
        callsInPlace(block, InvocationKind.EXACTLY_ONCE)
    }
    return block(this).also {
        recycle()
    }
}
```

Or simply making everything `lateinit`, which is ok as well 🤔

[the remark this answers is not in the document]
daividssilverio: except `lateinit` doesn't work with primitives.

PR #418, opened 2020-06-08
Improve autocomplete layout

### LINKED ISSUE #249: Ensure that auto complete suggestions popups are animated smoothly
## Description
This issue shouldn't be much of a problem, but let's make sure that we deliver a delightful experience when the user is receiving suggestions in the StartEdit or any of its sub views (Projects, Tags, Clients, Workspaces).

## Dependencies
This issue is being created very early, but it basically needs all the auto complete suggestion lists to be implemented already.

Ideally, this issue will be closed without any extra work after the autocomplete suggestions are done.

## Definition of done
- [ ] Auto complete suggestions pops are transitioned into the screen (they don't simply flash into existence). The todoist popups are a good reference of how it could be. (This could be only thing that needs some actual work in this issue).
- [ ] Collection changes are properly animated for all of the auto complete suggestions.

### LINKED ISSUE #405: Calculate and fix the initial position of the autocomplete suggestions popup
## Description
#335 couldn't finish up some details about the suggestions popup.
One of them is that popup isn't always positioned correctly when it's first shown.

Sometimes it will appear from the bottom of the description edit text instead of appearing from above (of course, this only happens when the `StartEdit` bottom sheet is not fully expanded).

## Dependencies
Needs #335 to be merged

## Definition of done
- [ ] The popup always appears from above the description field in the start edit view when it's not expanded (small start edit view)

### LINKED ISSUE #406: Improve the transitions of the auto complete suggestions popup
## Description
The suggestions popup introduced on #335 isn't animating as well as it could.
Currently as you type in the description and the number of suggestions starts to change, the auto complete suggestions popup size also changes, and that's ok, but it doesn't look great.

The wonky behaviour is more noticeable when the number of suggestions increases or decreases from 1 to 4 items. Try having 4 tags and one time entry in the log. Then type `name of the time entry` and then type and delete `#` quickly to see the problem.

## Dependencies
Depends on #335 getting merged

## Definition of done
- [ ] The popup transitions are fast and smooth

heytherewill:
### Summary
This implements the proper card layout for our autocomplete suggestions and also improves the display logic

### Technical Considerations
There's a known bug where apparently the popup is not being correctly collected when you close the start/edit fragment. To reproduce:

- Open start/edit
- Get a suggestion
- Dismiss start/edit by tapping outside the view
- Reopen start/edit

The suggestions will appear behind the fragment, but only until you manually trigger a dismiss. I have no idea why this is happening. Even weirder is the fact that I'm always manually dismissing the popup whenever the view closes.

### Relationships
Closes #249 
Closes #405 
Closes #406 

### Pull Request checklist

### Pointers
Play around and break the current implementation. I need corner cases and situations where this can be improved to make this as foolproof as possible

### Review template
- [ ] The code considers i18n

## Merge permissions
Only myself


daividssilverio:
@heytherewill 
> The suggestions will appear behind the fragment, but only until you manually trigger a dismiss. I have no idea why this is happening. Even weirder is the fact that I'm always manually dismissing the popup whenever the view closes.

This seems to be happening because the anchor view isn't attached yet when you try to show the popup for the first time.

if you run the logic to show the popup wrapped on a `View.doOnAttach`, it appears on the right layer of the screen.
Not really sure if is necessary, but I've added back the window type setting to test it as well: 
```
PopupWindowCompat.setWindowLayoutType(this, WindowManager.LayoutParams.TYPE_APPLICATION_SUB_PANEL)
```

That doesn't fix the size & y position calculation though.

heytherewill:
What is wrong with the size and Y calculation? 🤔

daividssilverio:
Sometimes it appears way too high (this should've been closer to the sheet).


Sometimes the max size & the position aren't right. (In this example, it should be all the way to the top and a bit shorter, right?)


Sometimes it appears bellow the text input. (It shouldn't do that on the sam

heytherewill:
The layout has been completely changed; are you testing on the right branch? 🤔

daividssilverio:
@heytherewill yes, `will/autocomplete-layout` and I can even see the rounded corners (at least half of the time)

timer/src/main/java/com/toggl/timer/startedit/ui/suggestions/AutocompleteSuggestionsPopup.kt
```diff
@@ -0,0 +1,75 @@
...
+import com.toggl.timer.R
+import com.toggl.timer.startedit.ui.autocomplete.AutocompleteSuggestionAdapter
+import com.toggl.timer.startedit.ui.autocomplete.AutocompleteSuggestionViewHolder
+import java.lang.ref.WeakReference
+import kotlin.math.min
+
+class AutocompleteSuggestionsPopup(
+    context: Context,
+    private val anchorTextView: WeakReference<TextView>,
+    onSuggestionTapped: (AutocompleteSuggestion) -> Unit = {}
```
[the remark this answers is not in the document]
heytherewill: Can we do that when we are migrating to Hilt? We could just inject this guy using the nice new fragment assisted injection thingy

semanticer:
@heytherewill Sorry, I reacted on slack https://toggl.slack.com/archives/CD40QCV25/p1591862644060000?thread_ts=1591814769.054300&cid=CD40QCV25 with a video if you missed it.

timer/src/test/java/com/toggl/timer/startedit/domain/AddProjectChipTappedTests.kt
```diff
@@ -15,21 +15,23 @@ internal class AddProjectChipTappedTests {
...
+        reducer.testReduceEffects(initialState, StartEditAction.AddProjectChipTapped) { effects ->
+            val action = effects.single().execute() as StartEditAction.DescriptionEntered
+            action.description shouldBe projectToken.toString()
+            action.cursorPosition shouldBe 1
         }
     }
 
     @Test
-    fun `should append @ to description and return no effects`() = runBlockingTest {
+    fun `should append the project token to description and return no effects`() = runBlockingTest {
```
daividssilverio, line 27: The test name says "return no effects", but it looks like an effect is returned 🤔

timer/src/test/java/com/toggl/timer/startedit/domain/AddTagChipTappedTests.kt
```diff
@@ -15,21 +15,23 @@ internal class AddTagChipTappedTests {
...
+        reducer.testReduceEffects(initialState, StartEditAction.AddTagChipTapped) { effects ->
+            val action = effects.single().execute() as StartEditAction.DescriptionEntered
+            action.description shouldBe tagToken.toString()
+            action.cursorPosition shouldBe 1
         }
     }
 
     @Test
-    fun `should append # to description and return no effects`() = runBlockingTest {
+    fun `should append the tag token to description and return no effects`() = runBlockingTest {
```
daividssilverio, line 27: Same here.

timer/src/test/java/com/toggl/timer/startedit/domain/ProjectButtonTappedActionTests.kt
```diff
@@ -15,21 +15,23 @@ internal class ProjectButtonTappedActionTests {
...
+        reducer.testReduceEffects(initialState, StartEditAction.ProjectButtonTapped) { effects ->
+            val action = effects.single().execute() as StartEditAction.DescriptionEntered
+            action.description shouldBe projectToken.toString()
+            action.cursorPosition shouldBe 1
         }
     }
 
     @Test
-    fun `should append @ to description and return no effects`() = runBlockingTest {
+    fun `should append the project token to description and return no effects`() = runBlockingTest {
```
daividssilverio, line 27: Same here. Am I getting something wrong here?

timer/src/test/java/com/toggl/timer/startedit/domain/TagButtonTappedActionTests.kt
```diff
@@ -15,21 +15,23 @@ internal class TagButtonTappedActionTests {
...
+        reducer.testReduceEffects(initialState, StartEditAction.TagButtonTapped) { effects ->
+            val action = effects.single().execute() as StartEditAction.DescriptionEntered
+            action.description shouldBe tagToken.toString()
+            action.cursorPosition shouldBe 1
         }
     }
 
     @Test
-    fun `should append # to description and return no effects`() = runBlockingTest {
+    fun `should append the tag token to description and return no effects`() = runBlockingTest {
```
daividssilverio, line 27: Same here.

timer/src/main/java/com/toggl/timer/startedit/ui/suggestions/AutocompleteSuggestionsPopup.kt
```diff
@@ -0,0 +1,75 @@
...
+
+            val locationOnScreen = intArrayOf(0, 0)
+            it.getLocationOnScreen(locationOnScreen)
+            val yPositionOfTextView = locationOnScreen[1]
+            val popupHeight = min(numberOfSuggestions * itemHeight, popupMaxHeight)
+
+            val positionAboveTextView = yPositionOfTextView - popupHeight
+            val positionBelowTextView = yPositionOfTextView + it.height
+            val shouldShowBelowTextView = positionAboveTextView < 0
+            val popupYPosition = if (shouldShowBelowTextView) positionBelowTextView else positionAboveTextView
```
daividssilverio, line 61: I still think we should force showing the popup above or bellow the text input layout based on the bottom sheet expanded state. That is, show the popup above the input when the view is not fully expanded (small view) and bellow the input in the expanded (big view).

timer/src/main/java/com/toggl/timer/startedit/ui/suggestions/AutocompleteSuggestionsPopup.kt
```diff
@@ -0,0 +1,75 @@
...
+        anchorTextView.get()?.doOnLayout {
+            if (numberOfSuggestions <= 0) {
+                dismiss()
+                return@doOnLayout
+            }
+
+            val locationOnScreen = intArrayOf(0, 0)
+            it.getLocationOnScreen(locationOnScreen)
+            val yPositionOfTextView = locationOnScreen[1]
+            val popupHeight = min(numberOfSuggestions * itemHeight, popupMaxHeight)
```
daividssilverio, line 56: To calculate the actual maxHeight considering the space between the input and the keyboard/top of the window, you might wanna check the initial idea here: https://github.com/toggl/android/blob/e718be986f09efce3753c6506395eff8f9cc1760/common-android/src/main/java/com/toggl/common/Popup.kt#L139-L182

daividssilverio: It looks like changing it to something like this fixes the problem with suggestions to long to fit the screen:
```kotlin
            val actualMaxContentHeight = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.N)
                popup.getMaxAvailableHeight(it, 0, false)
            else popup.getMaxAvailableHeight(it, 0)

            val actualHeight = min(popupHeight, actualMaxContentHeight)
            val positionAboveTextView = yPositionOfTextView - actualHeight
            val positionBelowTextView = yPositionOfTextView + it.height
            val shouldShowBelowTextView = positionAboveTextView < 0
            val popupYPosition = if (shouldShowBelowTextView) positionBelowTextView else positionAboveTextView

            popup.show(
                x = popupHorizontalPadding,
                y = popupYPosition,
                width = screenWidth - (popupHorizontalPadding * 2),
                height = actualHeight
```

heytherewill:
> After you pick a project suggestion and type @ again, no suggestions appear
This is so minor I'll create an issue and fix when we have time

> If you press the Add project, Add tags chips, they always add @ or # to the description, even if the token is already there

What behavior do you suggest here?

daividssilverio:
> >If you press the Add project, Add tags chips, they always add @ or # to the description, even if the token is already there

> What behavior do you suggest here?

Don't re-add a tag/project token if it's already there. But this is also very minor, can be fixed (or not) somewhere else.

heytherewill:
Follow up issue to fix the keyboard: https://github.com/toggl/android/issues/470
I'll start a discussion with designers to see what the behavior of pressing multiple times should be

PR #434, opened 2020-06-12
💭 Suggestions load

### LINKED ISSUE #388: Add the LoadSuggestions Action
## Description
Add the LoadSuggestions Action. Use the 2 suggestionProviders to generate suggestions. Old logic is [here](https://github.com/toggl/mobileapp/blob/develop/Toggl.Core/Interactors/Suggestions/GetSuggestionsInteractor.cs)
## Dependencies
#385
## Definition of done
- [ ] Add the LoadSuggestions Action
- [ ] Add tests

### LINKED ISSUE #390: Trigger LoadSuggestions action
## Description
Trigger LoadSuggestions action when:
- app resumes from background
- user calendar preferences change
- sync finishes/completes when in foreground
If any of the above is not possible to implement, make a new issue for that.
## Dependencies
#385
#387 
#386 
## Definition of done
- [ ] Trigger LoadSuggestions action based on the 3 points above
- [ ] Add tests
- [ ] make new issues for each point the trigger situations that can not be implemented at this time

semanticer:
### Summary
This PR implements 

- LoadSuggestions
- SuggestionsLoaded
- LoadSuggestionEffect
- Completes suggestion DI boilerplate
- Dispatches LoadSuggestions on MainActivity resume

### Technical considerations
It's mostly just boilerplate, the only thing that I'm not so sure about is the placement of LoadSuggestions dispatch. I would much rather see suggestions somehow subscribed rather than relying on our correct refresh logic. 

Let me know If you have any ideas

### Relationships

Closes #390
Closes #388 

### Pull Request checklist

### Pointers
Commit by commit

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


heytherewill:
Please create the followup issues for calendar and sync interaction as described in #390

app/src/main/java/com/toggl/ui/MainActivity.kt
```diff
@@ -51,6 +60,11 @@ class MainActivity : AppCompatActivity(R.layout.main_activity), BottomNavigation
         return true
     }
 
+    override fun onResume() {
+        super.onResume()
+        suggestionsStore.dispatch(SuggestionsAction.LoadSuggestions)
```
daividssilverio, line 65: Yeah... this feels weird.

daividssilverio: It would be nice to see we somehow utilizing the lifecycle components to avoid wasting resources on the effects of this call. As it is, if the user comes and goes from the background a lot, this will dispatch and run a multiple times... If loading the suggestions becomes heavy, this can be a problem.

Do you mind creating an issue to revisit this? I mean, revisit where firing `load suggestions` or `how to load suggestions` should happen.

semanticer: https://github.com/toggl/android/issues/466

PR #463, opened 2020-06-16
Handle calendar auto scrolls

### LINKED ISSUE #367: Handle auto scrolls
## Description
This needs to add support for dragging items outside of the visible area and the scrollview has to scroll smoothly.

## Dependencies
Depends on #361, #362 , #363, #364, #365 , #366

## Definition of done
- [x] The scrollview scrolls smoothly when dragging outside the visible area

daividssilverio:
### Summary
This PR adds auto scrolls to the calendar widget.

### Technical considerations
There was a bunch of clutter in the old implementation and there might still be some here.
I've tried to trim it as much as possible.
The animator used to handle the auto scrolls now is indefinite and doesn't count on having its restart manually handled like in the old implementation.

There are some minor moments (while dragging) while the drawing rect doesn't really reflect the state, but that's there to provide smoothness for the view.

Because it seemed simple enough. Feel free to add your point of view.

### Relationships
Closes #367 

### Pull Request checklist

### Pointers
Create some time entries and move them around the edges of the screen.
They should auto scroll smoothly when doing so.
They should also auto scroll into frame when you lift your finger.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it


calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarWidgetView.kt
```diff
@@ -543,30 +573,79 @@ class CalendarWidgetView @JvmOverloads constructor(
                 EditAction.ChangeStart -> updateItemInEditModeStartTime()
                 EditAction.ChangeEnd -> updateItemInEditModeEndTime()
                 EditAction.ChangeOffset -> updateItemInEditModeOffset()
-                EditAction.None -> {
-                }
+                else -> Unit
```
heytherewill, line 576: Why not do an exhaustive match?

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarWidgetView.kt
```diff
@@ -543,30 +573,79 @@ class CalendarWidgetView @JvmOverloads constructor(
...
+            }
+
+            when {
+                shouldAutoScrollUp() -> draggingDirection = -1
+                shouldAutoScrollDown() -> draggingDirection = 1
```
heytherewill, line 581: Wouldn't it be better to use enums rather than using magic numbers?

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarWidgetView.kt
daividssilverio: It is exhaustive, just not using the `EditAction.None`. I can add it back, I don't even know why I replaced it with else 🤔

heytherewill:
Actually, now that I was actually holding the device, we might be overdoing the vibration a bit 😂 
Maybe let's vibrate once per hour?

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarWidgetView.kt
daividssilverio: 🤔 it would be more readable and the compiler should make it fast 🤔 I think we can replace it with enums 🤔

daividssilverio:
> Actually, now that I was actually holding the device, we might be overdoing the vibration a bit 😂
Maybe let's vibrate once per hour?

That's how we did in the old implementation 🤔 .

Do you mean like vibrating when we move close to an hour line? Or maybe when it's snapping to some other start/end time? Either way, let's maybe change this into another issue?

PR #469, opened 2020-06-16
💭 Add suggestions UI

### LINKED ISSUE #391: Add suggestions UI
## Description
1.Create a new Cell for suggestions, similar to the TE cell but no group support, no tag, no billable and no duration symbols. Check [designs](https://www.notion.so/toggl/Suggestions-c3e223d7feb942c88c6c9a20aa5606c2)
2. Add a new section in the Main log to show the suggestions. The section should only be visible when suggestions are available.
## Dependencies
#385
## Definition of done
- [ ] Create a new Cell for suggestions
- [ ] Add a new section in the Main log
- [ ] Hook the SuggestionTapped action to the UI
- [ ] Make sure everything works

semanticer:
### Summary
This implements #391 and a few other things that are not mentioned there like the necessary suggestion selectors or the main log titles. 

### Technical considerations

- I'm using merge adapter, it works great. 
- Some of the layout values still preliminary since we don't have the final design for Android.

🤷 

### Relationships

Closes  #391

### Pull Request checklist

### Pointers
You can go commit by commit, it should be easy to understand.
To see this in action, you have to actually add some suggestions, either in Suggestions Selector or higher in suggestion providers. Then you just need to give `SuggestionsState.fromTimerState` a fake user or make sure it's provided by AppState. 

@heytherewill Please make sure that the copy for 🇵🇹  is ok (I used google translator so it's probably not) 
@daividssilverio Please do the same for 🇯🇵   
Not sure who can do the 🇫🇷 

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


common-android/src/main/res/values-ja/strings.xml
```diff
@@ -22,6 +22,8 @@
     <plurals name="entriesDeleted" tools:ignore="UnusedQuantity">
         <item quantity="other">%d個の記録が削除されました</item>
     </plurals>
+    <string name="log_suggestions_title">提案</string>
+    <string name="log_time_entries_title">あなたの活動</string>
```
daividssilverio, line 26: 👍

PR #490, opened 2020-06-19
Route Navigation

### LINKED ISSUE #489: Investigate Routes
## Description
Our current implementation of Navigation is not the best and now that we are going to implement some other view that don't require editing, it might make sense to have proper routes.

## Definition of done
- [ ] We implement some kind of routing system that takes care of things so we don't have to implement `editableX` for all views, even those that don't require editing anything

heytherewill:
### Summary
This adds a routing mechanism. It has a couple of benefits:

- Adding new routes is a lot simpler
- All navigation code is centralized in a single place rather than scattered in many fragment files
- Changes to navigation behavior are easy
- The navigation routes encode all information needed, making the state cleaner

### Technical considerations
I think it's worth mentioning that this is possibly the first time in my software developer career where something I wrote worked on the first run. Goes to show that it's simple 😂 

I'm open to ideas!

### Relationships
Closes #489 

### Pull Request checklist
- [ ] The code in this PR is tested thoroughly enough that changing the behavior to an incorrect one breaks tests *I STILL NEED TO UPDATE TESTS. I want a pair of eyes on this before*

### Pointers
Go at it commit by commit

### Review template

## Merge permissions
Only myself


common-feature/src/main/java/com/toggl/common/feature/navigation/BackStackExtensions.kt
```diff
@@ -0,0 +1,40 @@
...
+
+typealias BackStack = List<Route>
+
+fun BackStack.push(route: Route) =
+    this + route
+
+fun BackStack.pop() =
+    this.dropLast(1)
+
+fun BackStack.getEditableTimeEntryIfAny() =
```
semanticer, line 15: I'm not the biggest fan of these non-generic functions for every single property. Take a look at `tomas/navigation-niceties-experiments` and see if you want to borrow something from me.

common-feature/src/main/java/com/toggl/common/feature/navigation/Router.kt
```diff
@@ -17,18 +16,34 @@ class Router @Inject constructor(private val context: Context) {
...
+
+                val newRoute = newBackStack[index]
+
+                if (newRoute == oldRoute) popOperations
+                else listOf(BackStackOperation.Pop)
+            }
+
+            yieldAll(neededPopOperations)
+
+            val numberOfValidOperations = oldBackStack.size - neededPopOperations.size
```
daividssilverio, line 37: What makes an operation valid?

daividssilverio:
Please don't merge this without adding tests for the router itself.

common-feature/src/test/java/com/toggl/common/feature/navigation/RouterTests.kt
```diff
@@ -0,0 +1,136 @@
...
+    fun `does not call the navController when the routes are the same but the parameters are different`() = runBlockingTest {
+
+        val oldStack = listOf<Route>(
+            Route.StartEdit(EditableTimeEntry.empty(1)),
+            Route.Project(EditableProject.empty(1))
+        )
+        val newStack = listOf<Route>(
+            Route.StartEdit(EditableTimeEntry.empty(1)),
+            Route.StartEdit(EditableTimeEntry.empty(12))
+        )
```
daividssilverio, line 86: Just to check if I got this right.
Those two route processing statements would be equivalent to the following actions in the app:
1. User clicks on the description to start creating a time entry
2. User clicks on a suggestion to create a project
3. User goes back (either by creating a project or giving up)
4. User goes back and tries to start another time entry.

calendar/src/main/java/com/toggl/calendar/contextualmenu/domain/ContextualMenuReducer.kt
```diff
@@ -123,7 +126,7 @@ class ContextualMenuReducer @Inject constructor(
     private fun MutableValue<ContextualMenuState>.mapToCalendarEvent() =
         mapState { selectedItem.toCalendarEvent() }
 
-    private fun MutableValue<ContextualMenuState>.mapToEditableTimeEntry() =
+    private fun MutableValue<ContextualMenuState>. mapToEditableTimeEntry() =
```
semanticer, line 129: 👀

common-feature/src/test/java/com/toggl/common/feature/navigation/RouterTests.kt
daividssilverio: @heytherewill please notice me senpai.

daividssilverio:
The close action isn't working anymore for contextual menu.

common-feature/src/test/java/com/toggl/common/feature/navigation/RouterTests.kt
heytherewill: This is not a realistic app route, it's merely meant to test the expected behavior

daividssilverio: Oh, ok. Thanks senpai.

daividssilverio:
Hey, the calendar contextual menu still doesn't work properly. It's ignoring changes to the selected item.
But if you rather fix that on another issue to get this moving and prevent further merge conflicts, I'm totally fine with that.

PR #503, opened 2020-06-24
🏗 Make calendar events part of the AppState

### LINKED ISSUE #473: Make calendar events part of the global app state
## Description

We need to make the calendar events part of the global app state.
This is not exactly part of the calendar project, but will be used by suggestions.

## Definition of done
- [x] Calendar events are part of the global app state and can be used by other features.
- [x] The calendar suggestions provider is updated to use the events from the state rather than fetching them again

daividssilverio:
### Summary
This does what's requested here: #473 (The calendar events are now part of the AppState)

### Technical considerations
I don't think there's any 🤔 to worry about... Do you see something wrong?
We just need to actually fetch & sync calendar events somewhere (and store it in the state).

🤷‍♀️

### Relationships
Closes #473

### Pull Request checklist

### Pointers
Review by commit.

### Review template

## Merge permissions
🦑 it


timer/src/main/java/com/toggl/timer/suggestions/domain/CalendarSuggestionProvider.kt
```diff
@@ -14,23 +13,26 @@ import kotlin.math.abs
...
-            .getCalendarEvents(startOfRange, endOfRange, calendarService.getAvailableCalendars())
-            .filter { it.description.isNotBlank() }
-            .sortedBy { abs(now.minutesUntil(it.startTime)) }
-            .map { Suggestion.Calendar(it, suggestionsState.user.defaultWorkspaceId) }
-            .take(maxNumberOfSuggestions)
-    }
+            calendarEvents.values
+                .asSequence()
+                .filter { it.description.isNotBlank() }
+                .filter { it.startTime >= startOfRange && it.startTime <= endOfRange }
```
semanticer, line 32: Let's use a range here
```suggestion
                .filter { it.startTime in startOfRange..endOfRange }
```

PR #507, opened 2020-06-25
📆🔨 Implement the calendar service

### LINKED ISSUE #475: Fetch user calendars
## Description

We need to complete the `getAvailableCalendars()` method. This should access the calendar's content resolver and fetch the available calendars.
For a reference of the implementation, check the old app.

## Dependencies
Depends on #474 

## Definition of done
- [ ] If permissions are granted, fetch all the existing calendars from the resolver and transform them into `CalendarSource`
- [ ] If permissions are not granted, return an empty array

### LINKED ISSUE #476: Fetch calendar events
## Description

We need to complete the `getCalendarEvents()` method. This should fetch all the calendar events between the given dates that are in the selected calendars.
The method needs to filter all events between the given start and end dates, that are not all day events.
For a reference of the implementation, check the old app.

## Dependencies
Depends on #474 

## Definition of done
- [ ] If permissions are granted, fetch all the existing events from the resolver that match the predicate and transform them into `CalendarEvent`
- [ ] If permissions are not granted, return an empty array

daividssilverio:
### Summary
This implements the calendar service.

### Technical considerations
I could not figure out how to use the content provider api properly to do something like a `calendar_id IN (?, ?, ?, ?)` to filter the calendar events by calendar id directly. So... I'm filtering after querying. It's a bit wasteful :/

Because I didn't manage to use the content provider api properly, and I didn't find docs on how to use it the way I wanted.
If you know the way, please enlighten me.

### Relationships
Closes #475
Closes #476 

### Pull Request checklist
- [x] The code does not hinder the app's performance (unless some user has like a lot of user calendars imported and a lot of calendar events every day... everything should be fine)

### Pointers
Review by commit.
If you want to test the fetching, you can add the following line to the`CalendarDayFragment:onViewCreated`:
```kotlin
        store.dispatch(CalendarDayAction.CalendarViewAppeared)
```

modify the `FetchCalendarEventsEffect` like this:
```kotlin
class FetchCalendarEventsEffect(
    private val calendarService: CalendarService,
    private val fromStartDate: OffsetDateTime,
    private val toEndDate: OffsetDateTime,
    private val fromCalendars: List<Calendar>,
) : Effect<CalendarDayAction.CalendarEventsFetched> {
    override suspend fun execute(): CalendarDayAction.CalendarEventsFetched? = withContext(dispatcherProvider.io) {
        val userCalendars = calendarService.getAvailableCalendars()
        calendarService.getCalendarEvents(fromStartDate, toEndDate, userCalendars)
            .let(CalendarDayAction::CalendarEventsFetched)
    }
}
```

and of course, grant the calendar permission during the startup or from the app settings.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it


environment/src/main/java/com/toggl/environment/services/calendar/PermissionAwareCalendarService.kt
```diff
@@ -0,0 +1,120 @@
...
+            fromStartDate.toEpochSecond() * 1000,
+            toEndDate.toEpochSecond() * 1000
+        )
+
+        val calendarIds = fromCalendars.map { it.id }.toSet()
+
+        return sequence {
+            cursor?.use {
+                if (it.count <= 0)
+                    return@sequence
```
semanticer, line 86: Same here
```suggestion
            cursor?.takeIf { it.count > 0 }?.use {
```

environment/src/main/java/com/toggl/environment/services/calendar/PermissionAwareCalendarService.kt
```diff
@@ -0,0 +1,120 @@
...
+
+                while (it.moveToNext()) {
+                    val isAllDay = it.getInt(eventIsAllDayIndex) == 1
+                    val calendarId = it.getString(eventCalendarIdIndex)
+                    if (isAllDay || !calendarIds.contains(calendarId))
+                        continue
+
+                    yield(it.nextCalendarItem())
```
semanticer, line 94: I always prefer to eliminate `continue`/`break` when you're not secrificing redebility. This is one of those cases IMO.  
```suggestion
                    if (!isAllDay && calendarIds.contains(calendarId))
                        yield(it.nextCalendarItem())
```

environment/src/main/java/com/toggl/environment/services/calendar/PermissionAwareCalendarService.kt
```diff
@@ -0,0 +1,51 @@
+package com.toggl.environment.services.calendar
+
+import android.content.Context
+import android.provider.CalendarContract
+import java.time.OffsetDateTime
+import javax.inject.Inject
+
+class PermissionAwareCalendarService @Inject constructor(val context: Context) : CalendarService {
```
heytherewill, line 12: I don't think we need to call it this way. We did it in the Xamarin App because of shared layer reasons

environment/src/main/java/com/toggl/environment/services/calendar/PermissionAwareCalendarService.kt
heytherewill: I agree with Tomas, but I'd wrap this in `{ }` because it's not a return statement

environment/src/main/java/com/toggl/environment/services/calendar/PermissionAwareCalendarService.kt
```diff
@@ -0,0 +1,120 @@
...
+            CalendarContract.Calendars.CONTENT_URI,
+            calendarProjection,
+            null,
+            null,
+            null
+        )
+        return sequence {
+            cursor?.use {
+                if (it.count <= 0)
+                    return@sequence
```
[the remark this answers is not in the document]
daividssilverio: I would have to add another line to close the cursor if did that, wouldn't I?

environment/src/main/java/com/toggl/environment/services/calendar/PermissionAwareCalendarService.kt
daividssilverio: Same point? Is it worth the extra line?

environment/src/main/java/com/toggl/environment/services/calendar/PermissionAwareCalendarService.kt
daividssilverio: This name was the easy way... The interface is already named `CalendarService`...
What do you suggest? `CursorCalendarService`, `ContentProviderCalendarService`, `CalendarServiceImpl`?

heytherewill: `ContentProviderCalendarService` is a good one. Impl is cursed

environment/src/main/java/com/toggl/environment/services/calendar/PermissionAwareCalendarService.kt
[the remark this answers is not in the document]
semanticer: Good catch I didn't realize that. This would be the full alternative:
```kotlin
cursor?.takeIf { it.count > 0 }?.use {
      // ... //
} ?: cursor?.close()
```

Not sure if it's better than the original, it's up to you 🤷‍♂️

PR #509, opened 2020-06-26
A hilt makes you not cut yourself

### LINKED ISSUE #429: Migrate to Hilt dependency injection
## Description
Transition to a new Hilt library which is built on top of Dagger specifically for android usage
https://developer.android.com/training/dependency-injection/hilt-android
https://dagger.dev/hilt/

## Dependencies
none

## Definition of done
- [ ] Add Hilt dependency to Gradle
- [ ] Refactor our DI code to use Hilt
- [ ] Use proper scopes

heytherewill:
### Summary
Imagine holding a dagger by the blade. Not nice, eh?

### Technical considerations
Maybe we should review our scoping a bit? I'd like another pair of eyes on that
Also, we need to update our templates :)

### Relationships
Closes #429 

### Pull Request checklist

### Pointers
Go at it commit by commit and pay attention to scoping to see if I have messed something up

### Review template

- [ ] The code considers i18n

## Merge permissions
Only myself


semanticer:
The crash is really weird, I couldn't find solution so far. Take a look at https://github.com/chrisbanes/tivi/pull/640/files for inspiration if you haven't already. 

Also, modules can now be objects instead of classes

daividssilverio:
Really liking all the 🔥 🔥 🔥
I'm reading the docs to see about the scoping. Will test the app to see if everything still works alright o/.

buildSrc/src/main/java/com/toggl/buildsrc/dependencies.kt
```diff
@@ -67,6 +67,12 @@ object Libs {
         const val browser = "androidx.browser:browser:1.0.0"
         const val fragment = "androidx.fragment:fragment:1.0.0"
 
+        object Hilt {
+            private const val version = "1.0.0-alpha01"
+            const val viewModel = "androidx.hilt:hilt-lifecycle-viewmodel:$version"
+            const val compiler = "com.google.dagger:hilt-android-compiler:$version"
```
daividssilverio, line 73:
```suggestion
            const val compiler = "androidx.hilt:hilt-compiler:$version"
```

Seems to fix the view models injections crashes. IWOMM.

daividssilverio:
One tiny detail, I can't create time entries in this branch of some reason.
I can't navigate to the  `StartEdit` view as well.
The logger tells me I can create time entries by tapping the play/stop button, but they won't appear in the log.

buildSrc/src/main/java/com/toggl/buildsrc/dependencies.kt
heytherewill: Copy pasta strikes again

PR #534, opened 2020-07-01
🔨 Handle calendar permission related settings actions

### LINKED ISSUE #447: Add AllowCalendarAccessToggled action
## Description
Add `AllowCalendarAccessToggled` action. It should update the `UserPrefs` object in the state. Add the missing property if needed.

Also include the `RequestCalendarPermission` effect
`CalendarPermissionRecieved` action. (If possible check in the cal module how this part is done)

## Dependencies
#435
## Definition of done
- [ ] Add AllowCalendarAccessToggled action
- [ ] Add tests

daividssilverio:
### Summary
This PR adds the reducer logic to handle enabling the calendar integration in the settings module.
It also changes the permission service to use some new cool ktx extensions o/

### Technical considerations
I'm not injecting the `PermissionRequesterService` and using it directly in the `SettingsReducer` because it depends on the activity lifecycle scope to work... and our reducers live in the `ApplicationScope`... We could in theory change all our di graph to live in the `ActivityScope`, but that can become a problem when we need to inject work managers, services and other activities (if ever needed).

With that, the idea is to use the `shouldRequestCalendarPermission` value wherever we are listening to the state to then handle requesting the calendar permission from the UI and emit `CalendarPermissionReceived` with the result.

Feel free to make a suggestion. I'm all ears.

### Relationships
Closes #447

### Pull Request checklist

### Pointers
You can review by commit.
Check [some of the internal discussion](https://toggl.slack.com/archives/C02APMYTM/p1593554097174500) here if it helps.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it


semanticer:
I'm gonna 🦑 this and require @heytherewill review on my PR(which will be build on top of this) instead to speed things up.

PR #533, opened 2020-06-30
⚙️ More settings actions

semanticer:
### Summary
This PR implements two more user preference modifying actions and FeedbackEntered action. I've also changed some names to make it more consistent and extracted default UserPreferences so they can be used in multiple places.

### Technical considerations
Nothing new or exciting here really. I've made the feedback message part of the `SettingsState` even though according to 

### Why is it done like this and not some other way?
Feel free to suggest anything

### Relationships

Closes #448 
Closes #453 
Closes #449 

### Pull Request checklist

- [x] The code does not hinder the app's performance
- [x] The code does not introduce regressions
- [x] The code in this PR is tested thoroughly enough that changing the behavior to an incorrect one breaks tests
- [x] One of our [live or file custom templates](https://github.com/toggl/android/tree/master/.idea/templates) was used in this PR

### Pointers
It should be fairly simple to review everything at once.

### Review template

- [ ] The code is thoroughly tested in well written and resilient manner
- [ ] The code is performant and leads to an optimal user experience
- [ ] The code is testable and does not cut corners for the sake of delivering
- [ ] The code considers older platforms well and will work in older devices just like it will in new ones

- [ ] The code considers light and dark theme
- [ ] The code considers RTL and does not mention left/right
- [ ] The code considers i18n
  - [ ] There are no hardcoded strings in the UI
  - [ ] All strings are included in every language we support OR there's an issue requesting new translations
- [ ] There are no hard-coded numbers referring to dimensions or font sizes
- [ ] The code considers UI elements that are similar and extracts their similarities into styles to avoid duplication

## Merge permissions
🦑 it

PR #535, opened 2020-07-02
🔨 Handle send feedback tapped action

### LINKED ISSUE #450: Add SendFeedbackTapped action
## Description
Add SendFeedbackTapped action. It should update the `UserPrefs` object in the state. Delete the message from the state. Add the missing property if needed.

Also add the `SendFeedback` effect and `FeedbackSent` action. Mock what can't be implemented like the actual network call and make an issues for that to be fixed later on

## Dependencies
#435
## Definition of done
- [ ] Add SendFeedbackTapped action
- [ ] Add SendFeedback effect
- [ ] Add FeedbackSent action
- [ ] Mock and make issues for what is not possible to implement
- [ ] Add tests

daividssilverio:
### Summary
This handles the flow to send the feedback message from our app settings.

### Technical considerations
The flow to send the feedback message feels a bit cumbersome... 
It's gonna go like this:

1. The feedback message is filled with the `FeedbackEntered` action.
2. `SendFeedbackTapped` is sent, changing the state of the `sendFeedbackRequest` to `Loading` (so that we can show a progress indicator) and `SendFeedbackEffect` is fired.
3. In the `SendFeedbackEffect`, we make the feedback api request and fire either `FeedbackSent` or `SetSendFeedbackError`.
4. When reducing `FeedbackSent`, we clear the feedback message and set the `sendFeedbackRequest` to `Loaded` so that we can show a success indicator, a toast idk.
5. When reducing `SetSendFeedbackError`, we set the `sendFeedbackRequest` to `Error`  so that we can show some friendly error message.
6. When reducing `SendFeedbackResultSeen` (which can be sent from the ui right after the success indicator has been seen, or if the user taps the error message, idk) 

What do you think?

🤷‍♂️ Feel free to suggest something different
Maybe we don't need that `Request` sealed class I've added, maybe we can just use `Unit`?

### Relationships
Depends on #533 to finish get the feedback message
Closes #450 

### Pull Request checklist

### Pointers
Reviewing by commit might help, but there are not a lot of additions.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it after we agree if that's the way to go & the tests were written


environment/src/main/java/com/toggl/environment/di/EnvironmentModule.kt
```diff
@@ -54,6 +59,31 @@ object EnvironmentModule {
...
             }
         }
+
+    @Provides
+    fun platformInfo(@ApplicationContext context: Context): PlatformInfo {
+        val language = Locale.getDefault().displayLanguage
+        val timezone = TimeZone.getDefault()
+        val packageInfo = context.packageManager.getPackageInfo(context.packageName, 0)
+        val versionName = packageInfo.versionName
+        @Suppress("DEPRECATION")
```
heytherewill, line 69: What api is google suggesting us to use if this is deprecated?

api/src/main/java/com/toggl/api/feedback/MockFeedbackApi.kt
```diff
@@ -0,0 +1,16 @@
...
+import com.toggl.models.validation.Email
+import kotlinx.coroutines.delay
+
+class MockFeedbackApi : FeedbackApi {
+
+    @Suppress("UNUSED_PARAMETER")
+    override suspend fun sendFeedback(email: Email, message: String, platformInfo: PlatformInfo): Request {
+        delay(1000)
+
+        return Request.Success
```
heytherewill, line 14: Any reason why you're mocking this? The endpoint is pretty straight forward

api/src/main/java/com/toggl/api/Request.kt
```diff
@@ -0,0 +1,6 @@
+package com.toggl.api
+
+sealed class Request {
+    object Success : Request()
+    data class Failure(val throwable: Throwable) : Request()
+}
```
heytherewill, line 6: I think such generic types should be contained in the api layer (IOW, made internal) and the APIs should return meaninful types instead.

heytherewill: TL;DR: Let's abstract networking from our API layer

api/src/main/java/com/toggl/api/feedback/FeedbackApi.kt
```diff
@@ -0,0 +1,9 @@
+package com.toggl.api.feedback
+
+import com.toggl.api.Request
+import com.toggl.models.domain.PlatformInfo
+import com.toggl.models.validation.Email
+
+interface FeedbackApi {
+    suspend fun sendFeedback(email: Email, message: String, platformInfo: PlatformInfo): Request
```
heytherewill, line 8: I would just inject the platform info in the feedback api concrete implementation and request only what the user can actually change

heytherewill: But I'm fine with doing things this way as well

app/src/main/java/com/toggl/domain/AppState.kt
```diff
@@ -43,6 +43,7 @@ data class AppState(
     val backStack: BackStack = emptyList(),
     val calendarPermissionWasGranted: Boolean = false,
     val shouldRequestCalendarPermission: Boolean = false,
+    val sendFeedbackRequest: Loadable<Unit> = Loadable.Uninitialized,
```
heytherewill, line 46: How's this relevant to anyone but the settings state?

models/src/main/java/com/toggl/models/domain/PlatformInfo.kt
```diff
@@ -0,0 +1,14 @@
+package com.toggl.models.domain
+
+data class PlatformInfo(
+    val currentNativeLanguageCode: String,
+    val timezoneIdentifier: String,
+    val version: String,
+    val buildNumber: String,
+    val phoneModel: String,
+    val installLocation: String
```
heytherewill, line 9:
```suggestion
    val installLocation: InstallLocation
```

models/src/main/java/com/toggl/models/domain/User.kt
```diff
@@ -1,8 +1,10 @@
 package com.toggl.models.domain
 
 import com.toggl.models.validation.ApiToken
+import com.toggl.models.validation.Email
 
 data class User(
     val apiToken: ApiToken,
-    val defaultWorkspaceId: Long
+    val defaultWorkspaceId: Long,
+    val email: Email
```
heytherewill, line 9: Should this be `Email.Valid`?

settings/src/main/java/com/toggl/settings/di/SettingsModule.kt
```diff
@@ -1,9 +1,17 @@
...
+import javax.inject.Singleton
 
 @Module
 @InstallIn(ApplicationComponent::class)
-object SettingsModule
\ No newline at end of file
+object SettingsModule {
+    @Provides
+    @Singleton
+    fun feedbackApi(): FeedbackApi = MockFeedbackApi()
```
heytherewill, line 16: This should be provided by the API
And again, I don't think we should mock this, given how simple it is

settings/src/main/java/com/toggl/settings/domain/SettingsReducer.kt
```diff
@@ -53,6 +61,24 @@ class SettingsReducer @Inject constructor(
                 else
                     effects
             }
+            is SettingsAction.SendFeedbackTapped -> {
+                val user = state().user() ?: throw IllegalStateException("No user found")
```
heytherewill, line 65: You can make the settings an optional state that depends on the user existing

api/src/main/java/com/toggl/api/feedback/MockFeedbackApi.kt
daividssilverio: Because we don't have proper auth setup yet, nor network plumbing.

app/src/main/java/com/toggl/domain/AppState.kt
daividssilverio: We still need to add that mapping right? Otherwise we'd lose that state during the `AppState <-> SettingsState` mapping. Did I get something wrong?

environment/src/main/java/com/toggl/environment/di/EnvironmentModule.kt
daividssilverio: I'm using it down below: `longVersionCode` for Pie and above.

settings/src/main/java/com/toggl/settings/domain/SettingsReducer.kt
daividssilverio: Oh, I was actually thinking about doing that but I wasn't sure about that. Will do.

app/src/main/java/com/toggl/domain/AppState.kt
heytherewill: We need the mapping yes, but hid this behind a `SettingsLocalState` rather than exposing it to all other modules

models/src/main/java/com/toggl/models/domain/User.kt
daividssilverio: It makes sense to be, I'll change it.

daividssilverio:
Ok... the last three commits add retrofit, set it up and implement the feedback api.
The authorization part wasn't done because I'm not sure yet how to provide the api token to the authentication interceptor. 

We could maybe always pass the user and use a `@Header` annotation for every single authenticated api call. That would work 🤔

environment/src/main/java/com/toggl/environment/di/EnvironmentModule.kt
```diff
@@ -84,6 +94,31 @@ object EnvironmentModule {
...
+        return Base64.encodeToString(authStringBytes, Base64.DEFAULT)
+    }
+
+    @Provides
+    fun authInterceptor(@ApiAuthCredentials authString: String?): AuthInterceptor = AuthInterceptor(authString)
+
+    @Provides
+    @BaseApiUrl
+    fun baseEndpointUrl(): String =
+        if (BuildConfig.BUILD_TYPE.contentEquals("release")) "https://mobile.toggl.com"
```
heytherewill, line 113: Let's create our own wrapper that does not give me gag reflex when we read it 😂 

```suggestion
        if (BuildConfig.buildTypeIsRelease) "https://mobile.toggl.com"
```

Or something like that

api/src/main/java/com/toggl/api/feedback/FeedbackApi.kt
```diff
@@ -2,7 +2,9 @@ package com.toggl.api.feedback
 
 import com.toggl.models.domain.PlatformInfo
 import com.toggl.models.validation.Email
+import retrofit2.http.POST
 
 interface FeedbackApi {
+    @POST("mobile/feedback")
     suspend fun sendFeedback(email: Email.Valid, message: String, platformInfo: PlatformInfo)
```
heytherewill, line 9: I doubt Retrofit knows how to serialize all that into a sensible post request. And, like I requested in the other review, I'd rather not exposed these retrofit clients at all, instead relying on exposing services that can receive any kind input, convert that into the actual API and only then call these retrofit interfaces, which should be internal

daividssilverio: Cool. Were you thinking about having a big facade like we have for the `Repository` and our daos?

daividssilverio:
Issue to implement the remaining feedback data #579

PR #536, opened 2020-07-03
📈 Add analytics event GroupTimeEntriesSettingsChanged

### LINKED ISSUE #456: Add GroupTimeEntriesSettingsChanged event
## Description
Add 'GroupTimeEntriesSettingsChanged' event.

Check [notion](https://www.notion.so/toggl/Settings) for details regarding parameters

## Dependencies
#435 #444
## Definition of done
- [ ] Add GroupTimeEntriesSettingsChanged event
- [ ] Add tests

semanticer:
### Summary
Implements GroupTimeEntriesSettingsChanged event

### Technical considerations
I had to do the string conversion manually to capitalize True/False so it's consistent with how C# reported this event.

🤷 

### Relationships

Closes #456

### Pull Request checklist

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑


environment/src/main/java/com/toggl/environment/services/analytics/Event.kt
```diff
@@ -55,5 +55,8 @@ class Event private constructor(val name: String, val parameters: Map<String, St
                 "DistinctWorkspaceCount" to workspaceCount.toString()
             ) + providerCounts
         )
+
+        fun groupTimeEntriesSettingsChanged(state: Boolean) =
+            Event("GroupTimeEntriesSettingsChanged", mapOf("State" to if (state) "True" else "False"))
```
pdrj, line 60: I'm totally not in Aurora frame of mind yet, just wanted to offer a tiny suggestion

```suggestion
            Event("GroupTimeEntriesSettingsChanged", mapOf("State" to state.toString().capitalize()))
```

Not sure if it actually reads better or not 😄

semanticer: I'm not sure either but it's shorter so I guess I'll take it.

PR #537, opened 2020-07-03
👞 Declare sign out actions and effect

### LINKED ISSUE #454: Add SignOutTapped action
## Description
Add SignOutTapped action, `SignOut` effect and `SignOutCompleted` action. 
SignOut effect and the SignOutCompleted action are the most interesting part here since it should change the state in a way that it takes the user to the login view and triggers a clean up of the data like TEs etc. and api tokes and what not. Whatever is not possible to implement at this point, leave a comment and make an issue so that it can be handled later on.

As for the correct name `SignOut` vs `Log Out` I leave it to the dev working on this to talk with product to find out what is best.

## Dependencies
#435
## Definition of done
- [ ] Add SignOutTapped action
- [ ] Add SignOut effect
- [ ] Add SignOutCompleted action
- [ ] Add tests

### LINKED ISSUE #455: Add SignOut event
## Description
Add SignOut event.

## Dependencies
#435 #454 
## Definition of done
- [ ] Add SignOut event
- [ ] Add tests

semanticer:
### Summary
This PR implements SignOut magic and adds event to track it. #547

### Technical considerations
I needed some global place to handle signout so I created a new high order reducer **AuthReducer** to handle AppState cleaning. We still need to handle unsynced entities etc but this can be done only after we have some kind of sync in place - here is the new issue for it #547

Please let me know in case you have any suggestions. 

### Relationships

Closes #454 
Closes #455

### Pull Request checklist

### Pointers
Probably review commit by commit

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 anyone


app/src/main/java/com/toggl/domain/reducers/AuthReducer.kt
```diff
@@ -0,0 +1,27 @@
...
+import com.toggl.architecture.core.Effect
+import com.toggl.architecture.core.HigherOrderReducer
+import com.toggl.architecture.core.MutableValue
+import com.toggl.architecture.core.Reducer
+import com.toggl.architecture.core.isOrWraps
+import com.toggl.domain.AppAction
+import com.toggl.domain.AppState
+import com.toggl.settings.domain.SettingsAction
+
+class AuthReducer(override val innerReducer: Reducer<AppState, AppAction>)
```
daividssilverio, line 12: Could add tests for the auth reducer as well? It would be nice to see a tests asserting that the inner reducer also handles the the action (that's what it doing, right?).

app/src/main/java/com/toggl/domain/reducers/AuthReducer.kt
```diff
@@ -0,0 +1,27 @@
...
+    : HigherOrderReducer<AppState, AppAction> {
+    override fun reduce(
+        state: MutableValue<AppState>,
+        action: AppAction
+    ): List<Effect<AppAction>> {
+        if (action.isOrWraps<SettingsAction.SignOutCompleted>()) {
+            state.mutate {
+                AppState(
+                    calendarPermissionWasGranted = calendarPermissionWasGranted
```
heytherewill, line 21: I'm not immediately sure why this is needed 🤔

app/src/main/java/com/toggl/domain/reducers/AuthReducer.kt
heytherewill, line 12: Why is the name auth?

semanticer: Yeah, it's a bad name, I guess I was thinking it could handle more things in the future but it doesn't really make much sense.

app/src/main/java/com/toggl/domain/reducers/AuthReducer.kt
semanticer: My idea was that we grant permissions on the app installation basis rather than per user, so I wanted to preserve this information in our state. But I might be completely wrong about this...

semanticer:
> Would you mind explaining why you need to handle logging out "in a global place"?

@heytherewill Because this signout cleanup can be triggered by other things than just settings, like token expiration.

simonrozsival:
I must say that I'm not very happy that there are multiple reducers managing the same part of the state, but I guess we can't do better with the current architecture (without any middleware). What might be possible though is to add a `CleanUp` action to every single sub-store and let every reducer handle the action itself. That could solve also the thing with the calendar permissions in a more elegant way (it's really not obvious why you're dealing with calendar permissions in such a "top level" reducer, also there might be even more exceptions to the "clean up" in the future (maybe onboarding?)).

app/src/main/java/com/toggl/domain/reducers/AuthReducer.kt
daividssilverio: We ask for permissions on the app installation basis, yes. But we do that because we can't invalidade given permissions after they were already given. They should be in the AppState basically because we need to know the permission state in the calendar and settings modules.

PR #538, opened 2020-07-06
Use user preferences selected calendars

### LINKED ISSUE #477: Use selected calendars
## Description

We need to store the selected calendars in user defaults and use that when fetching events.

## Dependencies
Depends on #474 

## Definition of done
- [ ] Selected calendars are stored in user defaults
- [ ] All calls to `getCalendarEvents` use the selected calendars only

daividssilverio:
### Summary
This makes the calendar day view use the selected user calendars from the user preferences to load calendars events.

### Technical considerations
None. I've added some missing tests and I've updated the calendar service to be expose suspend functions.

🤷‍♂️ This seems simple enough.

### Relationships
Closes #477

### Pull Request checklist

### Pointers
Check the tests.

### Review template

## Merge permissions
🦑 it


calendar/src/main/java/com/toggl/calendar/calendarday/domain/FetchCalendarEventsEffect.kt
```diff
@@ -2,19 +2,21 @@ package com.toggl.calendar.calendarday.domain
...
     private val calendarService: CalendarService,
+    private val settingsRepository: SettingsRepository,
     private val fromStartDate: OffsetDateTime,
     private val toEndDate: OffsetDateTime,
-    private val fromCalendars: List<Calendar>,
     private val dispatcherProvider: DispatcherProvider
 ) : Effect<CalendarDayAction.CalendarEventsFetched> {
     override suspend fun execute(): CalendarDayAction.CalendarEventsFetched? = withContext(dispatcherProvider.io) {
+        val selectedCalendarIds = settingsRepository.loadUserPreferences().calendarIds
+        val fromCalendars = calendarService.getAvailableCalendars().filter { selectedCalendarIds.contains(it.id) }
```
heytherewill, line 19: Why not have two methods in the calendar service: A `getAllAvailableCalendars` and a `getUserSelectedCalendars`? I guess that will make the API a bit more foolproof than expecting people to use a filter and import the settings repository

daividssilverio: 🤔 We could have another service for that I assume. Some `UserSettingsAwareCalendarService` of sorts, and provide it in the calendar module.

If we don't want to do that, we'll create a dependency to the settings repository in the calendar service and we'll need to either provide the new calendar service from the calendar module, or add a dependency to the `repository` module in the `environment` module.

Which way should we go?

daividssilverio:
Regarding the [internal discussion](https://toggl.slack.com/archives/C02APMYTM/p1594065236390400) about when the calendars should be reloaded into the state. Here's an issue to handle loading/refreshing the calendars & user calendars into the state: #580

daividssilverio:
@pdrj good point about where we'll be using this, actually, we'll eventually need to load the calendars in a background service when the device is restarted or periodically to update the smart alerts.

PR #546, opened 2020-07-07
Handle client picking

### LINKED ISSUE #487: Implement client picking in the project view
## Description
We need to implement the client-picker-thingy in the project view. For this we should reuse the nice PopupWindow that we are using for the autocomplete suggestions. The input should be the list of all clients in the current workspace, filtered according to whatever is written is in the edit text

The chip should morph into an edit text. The first suggestions should be a suggestion to create a client if there's no client with the exact same name in the workspace. There should always be a suggestion to select no client. Implement client creation and the needed actions as well.

## Definition of done
- [ ] The chip morphs into an EditText when tapped
- [ ] The user can pick a client from the list of type to filter clients
- [ ] Only clients in the current workspace appear
- [ ] The user can create a new client if the query typed doesn't match a client name exactly
- [ ] The user can select no client any time
- [ ] When the workspace changes, the client should be set to null
- [ ] When the Workspace is picked, we emit the appropriate effect
- [ ] Docs are updated to reflect the changes done in this PR

daividssilverio:
### Summary
This PR handles picking clients in the project view.

### Technical considerations
~The logic to filter the clients, show "no client" and "create client" is being handled in the ui side... It made sense when implementing the auto completion for the workspaces... but I'm not so sure anymore if that's a good idea. I mean, it works... but should we do that?~

I've updated the view and the reducer so that the filtering and which suggestions should appear are handled directly by the reducer.

~I was following the idea I wrote when doing the workspace picker.~
Now I'm not sure anymore. What do you think? Should we move the logic back to the reducer and try to keep the view leaner?
I've updated the view and the reducer so that the filtering and which suggestions should appear are handled directly by the reducer.

### Relationships
Closes #487 

### Pull Request checklist

### Pointers
Test the view, it works.
~Review by commit, it might help.~ Don't bother trying to review by commit... I can rearrange them if you really want.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it after the draft is done. The layouts are not done yet and we might want to implement it in a different way.


timer/src/main/java/com/toggl/timer/project/domain/ProjectAction.kt
```diff
@@ -18,6 +18,9 @@ sealed class ProjectAction {
     data class ColorPicked(val color: String) : ProjectAction()
     data class WorkspacePicked(val workspace: Workspace) : ProjectAction()
     data class ClientPicked(val client: Client) : ProjectAction()
+    data class CreateClientSuggestionTapped(val name: String): ProjectAction()
+    data class ClientCreated(val client: Client) : ProjectAction()
+    object NoClientPicked: ProjectAction()
```
heytherewill, line 23: Why not change `ClientPicked` to `ClientPicked(val client: Client?)`

timer/src/main/java/com/toggl/timer/project/ui/ProjectDialogFragment.kt
```diff
@@ -190,12 +236,46 @@ class ProjectDialogFragment : BottomSheetDialogFragment() {
...
+
+        selectedWorkspaceClients
+            .combine(clientAutoCompleteDescription) { clients, text ->
+                val createClientSuggestion = if (!text.isNullOrBlank() && clients.count { it.name == text } == 0) {
+                    listOf<ClientSuggestion>(ClientSuggestion.CreateClient(text))
+                } else emptyList()
+                createClientSuggestion + noClientSuggestion + clients
+            }
+            .onEach { clientSuggestionSuggestionsPopup.updateAutocompleteSuggestions(it) }
+            .launchIn(lifecycleScope)
```
heytherewill, line 265: If you move the search string to the state you could probably extract this into a nice selector, couldn't you?

timer/src/main/java/com/toggl/timer/project/domain/ProjectAction.kt
daividssilverio: That works too, even less tests to write.

timer/src/main/java/com/toggl/timer/project/ui/ProjectDialogFragment.kt
daividssilverio: That's what I was wondering about doing or not.
I could make both auto complete suggestions popups work to only show whatever comes from the state and not have much done in the UI.

That would imply adding:
- Adding the client/workspace text query to the local state
- Adding the clients/workspaces suggestions to the local state
- Adding the actions/effects to handle it all, pretty much like the `StartEdit` does for the auto complete suggestions

Would you do it some other way?

timer/src/main/java/com/toggl/timer/project/ui/ProjectDialogFragment.kt
```diff
@@ -151,16 +152,17 @@ class ProjectDialogFragment : BottomSheetDialogFragment() {
...
 
         client_chip.setOnClickListener {
             autocompleteRequestFlow.value = AutoCompleteTarget.Clients
-            project_name_edit_text.setSafeTextPreventingActionDispatching("")
+            client_workspace_edit_text.setText("")
         }
 
         cancel_pick.setOnClickListener {
             autocompleteRequestFlow.value = AutoCompleteTarget.None
+            client_workspace_edit_text.setText("")
```
heytherewill, line 162: Not sure I like this, why not ensure that the query is cleared from the reducer? 🤔

PR #548, opened 2020-07-07
👨🏿‍🎤 Add UpdateEmail and UpdateName actions

### LINKED ISSUE #510: Settings: Add `updateEmail` action
## Description
It should update the user email with the new value

## Dependencies
None
## Definition of done
- [ ] Add action
- [ ] Add tests

### LINKED ISSUE #511: Settings: Add `updateName` action
## Description
It should update the user name with the new value

## Dependencies
None
## Definition of done
- [ ] Add action
- [ ] Add tests

semanticer:
### Summary
I'm trying to implement `UpdateEmail` and `UpdateName` actions in this PR. Since I'm not 100% sure how this should be implemented I'm giving you a draft version first. 

### Technical considerations
So far, I'm simply changing User's name or email locally however we should probably call some service, or is this going to be handled during the main sync? I'm not exactly sure what's the plan here @sofokli1 ?

Let me know

### Relationships
Closes #510 
Closes #511

### Pull Request checklist

### Pointers
Not yet ready for reviews

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 when it's ready


heytherewill:
You are right, this is handled by the sync. In fact, the default workspace is also a property of user and I think it doesn't need to be included in our shared pref settings

app/src/main/java/com/toggl/domain/mappings/Settings.kt
```diff
@@ -1,23 +1,31 @@
...
+import com.toggl.models.domain.User
 import com.toggl.settings.domain.SettingsAction
 import com.toggl.settings.domain.SettingsState
 
-fun mapAppStateToSettingsState(appState: AppState): SettingsState =
-    SettingsState(
+fun mapAppStateToSettingsState(appState: AppState): SettingsState? {
+    val user = (appState.user as? Loadable.Loaded<User>)?.value ?: return null
+    return SettingsState(
+        user,
```
heytherewill, line 13: I think this would read much better without the parenthesis

```suggestion
    val user = appState.user as? Loadable.Loaded<User> ?: return null
    return SettingsState(
        user.value,
```

sofokli1:
@semanticer since this is something new, I do not know what is the best corse of action. For me what makes sense is to update the user locally (in memory and in the repo) and leave it to the sync logic to push to the server.

semanticer:
Ok, so I guess this is the best we can do at this moment and let's see what needs to be done in the future.

PR #598, opened 2020-07-13
Implement login

### LINKED ISSUE #527: Settings: Implement Sign Out
## Description
Same as the old app. It should remove all data and direct the user to the login view.

## Dependencies
## Definition of done
- [ ] Implement the logic
- [ ] implement the UI

### LINKED ISSUE #547: Handle sign out in the repository
## Description
We need to handle unsynced data appropriately during sign out the same way we did in the old app.

## Dependencies
#454 
Probably some future sync related tasks

## Definition of done
- [ ] `Repository.signOut()` method should handle unsynced data properly.

### LINKED ISSUE #585: Email/password API interaction
## Description
Implement the login API call. You can look at the Xamarin app to see how the request should be serialized.

## Definition of done
- Login API can be called with email/password
- The response is handled
- Tests are added

heytherewill:
### Summary
This adds the login interaction to our app

### Technical considerations
The loading is a bit wonky. Subscriptions will fix it!
Also, we need to figure out what we are going to do with persisting the rest of the user object, but for now we only need the api token.

### Relationships
Closes #527
Closes #547
Closes #585 

### Pull Request checklist

### Pointers
Please look at it commit by commit

### Review template

## Merge permissions
Only myself


api/src/main/java/com/toggl/api/di/ApiModule.kt
```diff
@@ -0,0 +1,90 @@
...
+    fun baseApiUrl(@BaseUrl baseUrl: String): String = "$baseUrl/api/v9/"
+
+    @Provides
+    @Singleton
+    @BaseReportsUrl
+    fun baseReportsUrl(@BaseUrl baseUrl: String): String = "$baseUrl/reports/api/v3/"
+
+    @Provides
+    @Singleton
+    fun okHttpClient(authInterceptor: Interceptor) = OkHttpClient.Builder()
```
daividssilverio, line 48: Why isn't this specific already?

api/src/main/java/com/toggl/api/serializers/UserDeserializer.kt
```diff
@@ -0,0 +1,27 @@
...
+import com.google.gson.JsonDeserializationContext
+import com.google.gson.JsonDeserializer
+import com.google.gson.JsonElement
+import com.toggl.models.domain.User
+import com.toggl.models.validation.ApiToken
+import com.toggl.models.validation.Email
+import java.lang.IllegalStateException
+import java.lang.reflect.Type
+
+class UserDeserializer : JsonDeserializer<User> {
```
daividssilverio, line 12: Please test this class.

api/src/main/java/com/toggl/api/di/ApiModule.kt
heytherewill: I don't get what you mean

daividssilverio: Why not use the concrete implementation `AuthInterceptor`. If you use `Interceptor`, it will clash with other interceptors if we ever add them.

daividssilverio:
From my functional tests:
- Login works!
- The app is broken after ogin:
  - The bottom bar doesn't appear back
  - if you try to start a time entry, you get a weird [broken start edit view](https://user-images.githubusercontent.com/2173493/87327862-a3fb8e00-c502-11ea-82eb-e2f5de11fc12.png); You can't do anything with it.
  - The app just freezes after that.
  - I'm getting this crash when I go to the start edit:
  ```
  java.lang.IllegalStateException: View androidx.constraintlayout.widget.ConstraintLayout{733220 V.E...... ......ID 0,0-1080,1609 #7f090186 app:id/start_time_entry_layout aid=1073741858} does not have a NavController set
  ```
- For some reason, the app was quite slow when testing on emulators api 29 and 30; on my note 10, it run fine.
- I managed to see the main log briefly when I was logged out and then was navigated to the login screen. Not a big deal right now, but we should handle that at some point.
- ⚠️  **It seems like there's an infinite loop going on.** The entities never stop loading (they do load, then restart). I guess that explains why everything is so slow for now.
Leaks:
- There's some leaks going on the TimeEntriesLogFragment, something about the views refs not being cleared: [views.txt](https://github.com/toggl/android/files/4913658/views.txt)
- There a leak in the nav host fragment: [nav_host.txt](https://github.com/toggl/android/files/4913663/nav_host.txt)

app/src/main/java/com/toggl/domain/loading/LoadingSubscriber.kt
```diff
@@ -0,0 +1,24 @@
...
+
+@Singleton
+class LoadingSubscriber @Inject constructor(
+    private val store: Store<LoadingState, LoadingAction>
+) {
+    private var currentBackStack: BackStack = listOf()
+
+    fun processNewBackStack(newBackStack: BackStack) {
+        when {
+            currentBackStack.isEmpty() -> LoadingAction.StartLoading
```
daividssilverio, line 17: It looks like this is the source of the infinite loop.

daividssilverio: What's the idea of this loading subscriber? Could add some tests to describe what it should be doing?

api/src/main/java/com/toggl/api/di/ApiModule.kt
heytherewill: Oh, because `AuthInterceptor` doesn't exist in `ApiModule`. If we ever introduce other interceptors, we can use attributes to annotate this further, but for now that's fine (and I don't see us adding other interceptors)

heytherewill:
The crash when dismissing the StartEdit dialog is in develop and is related to the Router. It seems easy to solve, but solving it is unrelated to this PR.

app/src/main/java/com/toggl/ui/MainActivity.kt
```diff
@@ -32,32 +30,33 @@ import kotlinx.coroutines.flow.onEach
...
+        bottom_navigation.isVisible = when (destination.id) {
+            R.id.time_entries_log,
+            R.id.reports,
+            R.id.calendar,
+            R.id.start_time_entry_dialog -> true
+            else -> false
+        }
+
+        if (destination.id == R.id.time_entries_log) {
+            store.dispatch(AppAction.Loading(LoadingAction.StartLoading))
```
daividssilverio, line 48: Do we really want to trigger loading all the data every time we navigate to the main log?

daividssilverio:
The infinite loop has been fixed 👍 Nice.
The nav controller leak is still there.
From the new quick round of functional tests: If you press back in the login screen (without doing anything else) and re-open the app, the app is opened in the main log, but it's not interactive.

heytherewill:
@pdrj The change to handle errors is quite big, I'd rather merge the happy path (so we can extend on top of it) and figure this out later :)

heytherewill:
@pdrj and @daividssilverio I've implemented logging out in this PR because it was a nightmare to test the app otherwise. 

I won't be fixing the non-happy navigation path in this PR because the login navigation will change a lot (we don't have a welcome screen etc), so we might end up with two different nav-graphs. We should study some google apps to see how they solve it, because there's no default way of doing it in the navigation library.

Please review again with these things in mind 🙇

app/src/main/java/com/toggl/ui/MainActivity.kt
daividssilverio: @heytherewill ?

repository/src/main/java/com/toggl/repository/di/RepositoryModule.kt
```diff
@@ -1,20 +1,25 @@
...
 
 import android.content.SharedPreferences
+import com.toggl.api.ApiTokenProvider
 import com.toggl.database.dao.ClientDao
 import com.toggl.database.dao.ProjectDao
 import com.toggl.database.dao.TagDao
 import com.toggl.database.dao.TaskDao
 import com.toggl.database.dao.TimeEntryDao
+import com.toggl.database.dao.UserDao
+// import com.toggl.database.dao.UserDao
```
daividssilverio, line 11: 👀

daividssilverio:
Note: If necessary, go ahead and dismiss my review to get this merged ASAP if @pdrj approves this PR.

daividssilverio:
By the way... being the user essential to use basically every module other than `onboarding`, doesn't it make weird/troublesome to have all the other stores optional views? It's just a thought, but that bothered me a little.

app/src/main/java/com/toggl/ui/MainActivity.kt
heytherewill: Subscriptions will save us here

heytherewill:
Leaks: https://github.com/toggl/android/issues/604
Nav Graph bug: https://github.com/toggl/android/issues/605

The error handling bit should be handled by Piotr as he creates issues for the login

PR #599, opened 2020-07-14
Calendar swipe actions

### LINKED ISSUE #483: Add the week swiped action
## Description
Add the calendar week swiped action to the date picker module. This action receives a swipe direction, either left or right.

## Dependencies
Depends on #480

## Definition of done
- [ ] When swiping right, the selected date is updated to the same day in the next week in the reducer
- [ ] When swiping left, the selected date is updated to the same day in the previous week in the reducer
- [ ] Doesn't update the selected date if the new values are not in the available dates

simonrozsival:
### Summary
This is a straightforward implementation of an issue: I added a parametrized action + updated the reducer to handle this action.

While I was preparing this PR, I actually noticed that this PR might not be "RTL safe"... The issue explicitly says "swipe right" and "swipe left" and these actions are mapped to "shift one week into future" and "shift one week into the past", but that might not be how the app would work in a RTL language 🤔 I have no idea what's the "direction of time" in RTL languages and I guess that's not something that we consider at this stage of the app's development and it might have to be revisited in the future once we add proper support for RTL and we know more about the subject.

### Relationships
Closes #483

### Pull Request checklist

### Pointers
Let me know if you think there should be more tests or if the tests should be different.

### Review template

## Merge permissions
Anyone can merge this PR.


calendar/src/main/java/com/toggl/calendar/datepicker/domain/CalendarDatePickerAction.kt
```diff
@@ -8,10 +9,12 @@ sealed class CalendarDatePickerAction {
         val availableDates: List<OffsetDateTime>,
         val visibleDates: List<OffsetDateTime>
     ) : CalendarDatePickerAction()
+    data class Swiped(val direction: SwipeDirection) : CalendarDatePickerAction()
```
heytherewill, line 12: There's no action whose name is just `Swiped` in the [module feature definition](https://www.notion.so/toggl/CalendarDatePicker-19438314293746ae8eeeeead387fcc22)

calendar/src/main/java/com/toggl/calendar/datepicker/domain/CalendarDatePickerReducer.kt
```diff
@@ -33,5 +35,24 @@ class CalendarDatePickerReducer @Inject constructor(
...
 }
+
+private fun MutableValue<CalendarDatePickerState>.shiftSelectedDateByWeek(
+    direction: SwipeDirection
+): List<Effect<CalendarDatePickerAction>> {
+    return mutateWithoutEffects {
+        val shiftBy = when (direction) {
+            SwipeDirection.Left -> Constants.Calendar.datePickerRowItemsCount
+            SwipeDirection.Right -> -Constants.Calendar.datePickerRowItemsCount
+        }.toLong()
```
heytherewill, line 49: Why aren't the constant longs if they're only used here?

simonrozsival: Maybe they should. I'll think about it a bit.

daividssilverio:
There was one detail left in the description of the issue. When swiping left/right, we should set the selected date to the edge of the available dates when the swipe would end up out of bounds.

PR #602, opened 2020-07-15
📪 Implement time entry subscription

### LINKED ISSUE #329: Investigate different ways how to implement subscriptions and implement one of them
## Description
The goal is to explore multiple ways we can implement subscriptions into our current architecture. Proposed solutions should though out enough to identify the pros and cons as well as the difficulty of each individual approach.  

This could be partly done in related discussion https://github.com/toggl/mobile-discussions/issues/66

## Definition of done
- [ ] At least one but preferably multiple subscription implementations are proposed
- [ ] Pros and Cons are documented

### LINKED ISSUE #610: Implement time entry subscription
## Description
Implement subscription that will constantly update time entry list whenever there is a change in the database. 

## Dependencies
#329 

## Definition of done
- [ ] Refactor dao and repository to provide Flow 
- [ ] Implement subscription that will provide continuous time entry updates
- [ ] 🔥 Old code that was responsible for time entry updates

### LINKED ISSUE #611: Implement projects subsription
## Description
Implement a subscription that will constantly update projects list whenever there is a change in the database. 

## Dependencies
#329 

## Definition of done
- [ ] Refactor dao and repository to provide Flow 
- [ ] Implement subscription that will provide continuous project list updates
- [ ] 🔥 Old code that was responsible for project updates if possible

semanticer:
### Summary
This PR introduces the new subscription infrastructure and implements the first two subscriptions for **TimeEntries** and **Projects**.

Subscription is a very simple interface:
```kotlin
interface Subscription<State, Action : Any> {
    fun subscribe(state: Flow<State>): Flow<Action>
}
```

You can plug your suggestion into store: 
```kotlin
return FlowStore.create(
    initialState = AppState(),
    reducer = reducer,
    subscription = subscription,  // NEW!
    dispatcherProvider = dispatcherProvider,
    storeScopeProvider = application as StoreScopeProvider
)
```

### Technical considerations
Note that `LoadProjectsSubscriptions` couldn’t simplify `ProjectAction.ProjectCreated` handling the same way we simplified TimeEntry updates thanks to `LoadTimeEntriesSubscription` because if we won’t update the state immediately with the new project in `ProjectAction.ProjectCreated` handling we trigger [this throw in our Selector](https://github.com/toggl/android/blob/7289f54b6b76e5141a0f095998d415aecdc5cba0/timer/src/main/java/com/toggl/timer/startedit/domain/ProjectTagChipSelector.kt#L24). 
We can either leave it as it is or apply this small little change https://github.com/toggl/android/commit/4093f556f13ec1547c96b31a00fce7645af50b6f to prevent app from the immediate crash and let it update with the new project right after.
This kind of issue is caused by our “manual” pairing of state entities.

**Update** https://github.com/toggl/android/commit/4093f556f13ec1547c96b31a00fce7645af50b6f has been applied. 

I could have just made our effect return flow (similarly to what @Odrakir [did on iOS](https://github.com/toggl/ios/pull/595#issuecomment-660890111)) and we'd have a similar thing done within the bounds of our existing architecture. However, I decided to separate those two concepts completely to keep the existing effects clearly single-emit with suspended function magic in place. I believe it's easier to reason about the code when we keep those two concepts separate.

### Relationships

Closes #329
Closes #610  
Closes #611

### Pull Request checklist

### Review template

- [ ] The code considers i18n

## Merge permissions
Me when it's ready


app/src/main/java/com/toggl/di/SubscriptionModule.kt
```diff
@@ -0,0 +1,45 @@
...
+@ExperimentalCoroutinesApi
+@Module
+@InstallIn(ApplicationComponent::class)
+object SubscriptionModule {
+
+    @Provides
+    @ProvideAppSubscription
+    fun compositeSubscription(
+        @ProvideLoadTimeEntriesSubscription loadTimeEntriesSubscription: Subscription<AppState, AppAction>,
+        @ProvideProjectsSubscription loadProjectsSubscription: Subscription<AppState, AppAction>
```
heytherewill, line 25: Can't you do a `BindsIntoSet` to get all subscriptions here and then get a list here so you can inject in the composite?

daividssilverio:
This looks really nice! I think you can ignore my comment about triggering the `StartLoading` action on #598 @heytherewill if we decide to move forward with this.

app/src/main/java/com/toggl/domain/loading/LoadTimeEntriesSubscription.kt
```diff
@@ -0,0 +1,19 @@
...
+import kotlinx.coroutines.flow.flowOf
+import kotlinx.coroutines.flow.map
+import javax.inject.Inject
+
+@ExperimentalCoroutinesApi
+class LoadTimeEntriesSubscription @Inject constructor(
+    private val timeEntryRepository: TimeEntryRepository
+) : BaseLoadingSubscription() {
+    override fun subscribe(isUserLoggedIn: Boolean): Flow<LoadingAction> {
+        val timeEntries = if (isUserLoggedIn) timeEntryRepository.loadTimeEntries()
```
pdrj, line 15: We no longer use the IO dispatcher to load the entries. Is that fine?

app/src/main/java/com/toggl/di/SubscriptionModule.kt
```diff
@@ -19,25 +20,20 @@ import kotlinx.coroutines.ExperimentalCoroutinesApi
...
+    ): CompositeSubscription<AppState, AppAction> =
+        CompositeSubscription(subscriptions)
 
     @Provides
-    @ProvideLoadTimeEntriesSubscription
+    @IntoSet
     fun loadTimeEntriesSubscription(
         repository: Repository
     ): Subscription<AppState, AppAction> =
         LoadTimeEntriesSubscription(repository)
```
heytherewill: Given that this only has one dependency, can't you annotate the class itself with `@Binds @IntoSet` instead of having a provides method? 🤔

semanticer:
@heytherewill to be honest, I haven't even considered the idea of manipulating state directly inside of subscriptions. I wanted to leave the privilege of state change to reducers only. Subscriptions are closer to effect in my mind, their only responsibility should be the translation from state changes to actions, this is also more flexible since you can have the action handling in one place(reducer) but trigger the same action from various places (subscriptions, effects, UI). This is also how it's implemented in Elm - my main source of inspiration for this feature.

app/src/main/java/com/toggl/di/SubscriptionModule.kt
semanticer: Both `@Binds` and `@IntoSet` are not applicable to classes :-(

app/src/test/java/com/toggl/domain/loading/LoadProjectSubscriptionTests.kt
```diff
@@ -0,0 +1,144 @@
...
+            expectedOutput = listOf(
+                listOf(p1),
+                listOf(p2),
+                listOf(p1, p2)
+            )
+        )
+    }
+
+    private suspend fun LoadProjectsSubscriptions.testSubscribe(
+        inputTimeEntriesFlow: Flow<List<Project>>,
```
daividssilverio, line 132: ©️ 🍝 should've been projects, right?

app/src/test/java/com/toggl/domain/loading/LoadTimeEntriesSubscriptionTests.kt
```diff
@@ -0,0 +1,144 @@
...
+import org.junit.jupiter.api.Test
+
+@ExperimentalCoroutinesApi
+@DisplayName("The LoadTimeEntriesSubscription")
+class LoadTimeEntriesSubscriptionTests : CoroutineTest() {
+    private val repository = mockk<TimeEntryRepository>()
+    private val subscription = LoadTimeEntriesSubscription(repository, dispatcherProvider)
+
+    private val uninitializedUser = Loadable.Uninitialized
+    private val loadingUser = Loadable.Loading<User>()
```
daividssilverio, line 31:
```suggestion
    private val loadingUser = Loadable.Loading
```

app/src/test/java/com/toggl/domain/loading/LoadTimeEntriesSubscriptionTests.kt
```diff
@@ -0,0 +1,144 @@
...
+    private val errorUser = Loadable.Error<User>(Failure(IllegalAccessError(), ""))
+    private val loadedUser = Loadable.Loaded(User(
+        1,
+        ApiToken.from("token"),
```
daividssilverio, line 35:
```suggestion
        ApiToken.from("token") as ApiToken.Valid,
```

PR #603, opened 2020-07-16
✨ Add single-choice settings

### LINKED ISSUE #517: Settings: Implement select from list sub-setting
## Description
This will be used for situation like `duration format selection`, `First Day Of The Week selection`, `Date Format selection` and `Workspace selection`.

So this component should accept:
- a list of `SelecteableISetting`s that implement the protocol `SelecteableISetting`. (Feel free to change the name)
- a single `SelecteableISetting` for the selected value (for comparison purposes)
- The component should also accept a closure that will be called every time a new selection it is tapped. The closer should accept a parameter of type same as the `SelecteableISetting` `original item` so that it dispatches it to the store using the appropriate action.

The protocol should have a computed var for the `name` of the option (string) (it should also takes care of possible needed translation) and the `original item` that it is supposed to send as parameter to the action.

Overall this should be a table view with cells that have a string on one side for the name of the option and a checkmark on the other side to show the selected option.

*Keep in mind that this is just a suggestion and it is up to you (the person working on this issues) to make final decision.*

This component will be presented with a push transition (same as old app)

## Dependencies
## Definition of done
- [ ] Implement the logic
- [ ] implement the UI

### LINKED ISSUE #518: Settings: Implement First Day Of The Week selection
## Description
Present the options using what was implemented in `Settings: Implement select from list sub-setting`. Check dependencies for links.

## Dependencies
`Settings: Implement select from list sub-setting`: #517
## Definition of done
- [ ] Implement the logic
- [ ] implement the UI

### LINKED ISSUE #519: Settings: Implement Duration Format selection
## Description
Present the options using what was implemented in `Settings: Implement select from list sub-setting`. Check dependencies for links.

## Dependencies
`Settings: Implement select from list sub-setting`: #517
## Definition of done
- [ ] Implement the logic
- [ ] implement the UI

### LINKED ISSUE #520: Settings: Implement Date Format selection
## Description
Present the options using what was implemented in `Settings: Implement select from list sub-setting`. Check dependencies for links.

## Dependencies
`Settings: Implement select from list sub-setting`: #517
## Definition of done
- [ ] Implement the logic
- [ ] implement the UI

### LINKED ISSUE #521: Settings: Implement Workspace selection
## Description
Present the options using what was implemented in `Settings: Implement select from list sub-setting`. Check dependencies for links.

## Dependencies
`Settings: Implement select from list sub-setting`: #517
## Definition of done
- [ ] Implement the logic
- [ ] implement the UI

pdrj:
### Summary
This adds a component that can display a single-choice list of items and dispatch actions upon selection. It also uses the component for date format, duration format and first day of the week settings.

### Technical considerations
It's been a while since I've done aurora, so please tell me where I'm failing to do stuff the right way. Tests are not added yet (for the reducer) - they are very simple and I'll add them tomorrow.

We also have the same bug about NavController when navigating back from the choice dialog fragment as the one we have with the start/edit view.

The dialog fragment seems to be overlaid with some grayish opacity - I don't know what's up with that yet.

The preview doesn't work (only shows the header) - maybe it's because of the flow and it doesn't deal with that well?

I'm open to suggestions.

### Relationships

Closes #517
Closes #518
Closes #519 
Closes #520
Closes #521 

### Pull Request checklist

### Pointers
Go commit by commit.
Use the 3 implemented settings and see how the settings screen is updated in the background.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


settings/src/main/java/com/toggl/settings/ui/EditSettingsFragment.kt
```diff
@@ -4,13 +4,112 @@ import android.os.Bundle
...
+                    }
+            }
+            SettingsType.Workspace -> {
+                settingHeader = getString(R.string.default_workspace)
+                settingChoiceListItems = listOf()
+            }
+            else -> {
+                settingHeader = "UNKNOWN"
+                settingChoiceListItems = listOf()
+            }
```
heytherewill, line 103: Please move these away from the fragment

settings/src/main/java/com/toggl/settings/ui/EditSettingsFragment.kt
```diff
@@ -4,13 +4,112 @@ import android.os.Bundle
...
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        store.state
+            .mapNotNull {
+                if (it.backStack.last() is Route.SettingsEdit) {
+                    return@mapNotNull Pair(it.userPreferences, (it.backStack.last() as Route.SettingsEdit).parameter)
+                } else {
+                    null
+                }
```
heytherewill, line 52: There's an extension for getting a parameter from the backStack

daividssilverio:
> The preview doesn't work (only shows the header) - maybe it's because of the flow and it doesn't deal with that well?

That seems to be the case. I managed to get it working by separating the observable part of the composable from the ui part.

settings/src/main/java/com/toggl/settings/ui/composables/ChoiceListWithHeader.kt
```diff
@@ -0,0 +1,78 @@
...
+    ThemedPreview {
+        ChoiceListWithHeader(listOf(previewItems).asFlow(), previewHeader)
+    }
+}
+
+@Preview("ChoiceListWithHeader dark theme")
+@Composable
+fun PreviewChoiceListWithHeaderDark() {
+    ThemedPreview(darkTheme = true) {
+        ChoiceListWithHeader(listOf(previewItems).asFlow(), previewHeader)
```
daividssilverio, line 76:
```suggestion
    ChoiceListWithHeaderWrapper(items = observableItems, header = header, dispatcher = dispatcher)
}

@Composable
fun ChoiceListWithHeaderWrapper(
    items: List<ChoiceListItem>,
    header: String,
    dispatcher: (SettingsAction) -> Unit = {}
) {
    Column(modifier = Modifier.padding(grid_2)) {
        Text(
            text = header,
            style = MaterialTheme.typography.h5
        )
        Spacer(Modifier.preferredHeight(grid_2))
        RadioGroup {
            items.forEach { item ->
                RadioGroupItem(selected = item.isSelected, onSelect = { item.dispatchSelected(dispatcher) }) {
                    Row {
                        RadioButton(selected = item.isSelected, onSelect = { item.dispatchSelected(dispatcher) })
                        Spacer(modifier = Modifier.preferredWidth(grid_1))
                        Text(text = item.label, style = MaterialTheme.typography.body1)
                    }
                    Spacer(modifier = Modifier.preferredHeight(grid_1))
                }
            }
        }
    }
}

internal val previewItems = (1..5).map {
    if (it == 1) ChoiceListItem(
        "Choice selected",
        isSelected = true
    ) else ChoiceListItem("Choice ${Random.nextInt() % 100}")
}
internal val previewHeader = "First day of week"

@Preview("ChoiceListWithHeader light theme")
@Composable
fun PreviewChoiceListWithHeaderLight() {
    ThemedPreview {
        ChoiceListWithHeaderWrapper(previewItems, previewHeader)
    }
}

@Preview("ChoiceListWithHeader dark theme")
@Composable
fun PreviewChoiceListWithHeaderDark() {
    ThemedPreview(darkTheme = true) {
        ChoiceListWithHeaderWrapper(previewItems, previewHeader)
```

daividssilverio: With ☝️ the previews work and I think the parts of the composable are more well separated (ui logic, update logic)

settings/src/main/java/com/toggl/settings/ui/EditSettingsFragment.kt
```diff
@@ -4,13 +4,111 @@ import android.os.Bundle
...
+                settingHeader = "UNKNOWN"
+                settingChoiceListItems = listOf()
+            }
+        }
+
+        (view as ViewGroup).setContent(Recomposer.current()) {
+            ChoiceListWithHeader(
+                items = listOf(settingChoiceListItems).asFlow(),
+                header = settingHeader,
+                dispatcher = store::dispatch
```
daividssilverio, line 110: Shouldn't we dismiss the dialog when picking an option?

settings/src/main/java/com/toggl/settings/ui/composables/ChoiceListWithHeader.kt
```diff
@@ -0,0 +1,78 @@
...
+
+@Composable
+internal fun ChoiceListWithHeader(
+    items: Flow<List<ChoiceListItem>>,
+    header: String,
+    dispatcher: (SettingsAction) -> Unit = {}
+) {
+    val observableItems by items.collectAsState(initial = listOf())
+
+    Column(modifier = Modifier.padding(grid_2)) {
```
daividssilverio, line 35: If we add a modifier `Modifier.fillMaxWidth()` to this column, the dialog already looks much nicer.

daividssilverio:
🤔 I wonder what's the right way (or what would look/feel nicer) to do things with Compose.

We could have used a dialog compose without the navigation graph for this case 🤔 

We could change the `SettingsFragment.kt` to handle showing the dialog directly 🤔 

```kotlin
@ExperimentalCoroutinesApi
    override fun onCreateView(
        inflater: LayoutInflater,
        container: ViewGroup?,
        savedInstanceState: Bundle?
    ): View? = FrameLayout(requireContext()).apply {
        layoutParams = ViewGroup.LayoutParams(ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.MATCH_PARENT)

        (this as ViewGroup).setContent(Recomposer.current()) {
            val selectedState = store.state
                .select(settingsSelector!!)
            SettingsPage(selectedState, getString(R.string.settings), store::dispatch)

            // We could add or remove the dialog from the composition based on the state flow here
            Dialog(onCloseRequest = {}) { 
                Surface {
                    ChoiceListWithHeaderWrapper(previewItems, previewHeader)
                }
            }
        }
    }
```

settings/src/main/java/com/toggl/settings/ui/composables/ChoiceListWithHeader.kt
pdrj: Hm I'm not sure about that. Content is not in the center then 
Screenshot 2020-07-17 at 10 59 54

Is this what you had in mind?

pdrj:
> We could have used a dialog compose without the navigation graph for this case 🤔

Hm, that indeed might be nicer. I'll do some experimenting.

pdrj:
@daividssilverio I actually started working on your suggestion but now I'm not sure how this will play with navigation. Do we want to keep in state the ID of the single-select list we should be presenting? Right now we get that ID value from route param, but we won't have access to that anymore.

pdrj:
@daividssilverio or whoever else is interested. I did a quick POC of the dialog render inside the settings screen. It's on `peter/list-dialog-like-react` branch and only works for the 'Date format' setting. Let me know what you think. If this is something we want to do, I will polish it up.

pdrj:
That is the relevant commit: https://github.com/toggl/android/commit/81d09748272ad60e913669b4015966d9b205f6be

settings/src/main/java/com/toggl/settings/ui/composables/ChoiceListWithHeader.kt
daividssilverio: Yeah, that's actually how the current implementation looks like.
We could maybe add a min width if you think it looks like too much wasted space.
When it's wrapping content, it's just too narrow.

daividssilverio:
> @daividssilverio I actually started working on your suggestion but now I'm not sure how this will play with navigation. Do we want to keep in state the ID of the single-select list we should be presenting? Right now we get that ID value from route param, but we won't have access to that anymore.

Didn't check your POC branch before replying this, but we could maybe just use the backstack from the state without the nav framework. Or maybe create a custom navigator that would consider that we'd be handling that sort of navigation using compose.

daividssilverio:
Oh, I liked the idea of the selector in your branch as well! (Didn't like the problems injecting them though... I wish the compiler would be more helpful pointing out what's the problem there...)

daividssilverio:
@semanticer @heytherewill what do you think about the dialog idea and @pdrj 's POC?

semanticer:
Yep, I'd go with `peter/list-dialog-like-react` at least for "small" dialogs like these. Not just because it makes more sense in Compose but I would also argue that these dialogs are very tightly part of settings screen more than for example starEditDialog is part of the main log - it could theoretically be entered from some other place, therefore, it deserves more separation than settings dialogs.

pdrj:
Thanks @semanticer. Do you think it's ok to use the local state to track which dialog should be shown? https://github.com/toggl/android/commit/81d09748272ad60e913669b4015966d9b205f6be#diff-272bd07ac95810d31a268e47e0b7aa4fR20

Or do you think we need something more elegant like @daividssilverio suggested? That would be a bigger change though, since our routing is coupled with deep links from what I can tell.

pdrj:
BTW I also added the workspace selector setting to this PR, so it's now handling all the single-choice cases.

semanticer:
@pdrj I think it's completely fine to do this using the local state at least for now. It could be considered an internal state of settings view IMO. Imagine we would use a dropdown list instead of dialogs - we wouldn't use navigation infrastructure for that even though it's pretty much the same thing.

pdrj:
I've applied the discussed changed and made the dialog a bit prettier (I hope 🤔). This is ready for final reviews now.

Screenshot 2020-07-21 at 12 33 14

settings/src/main/java/com/toggl/settings/ui/composables/ChoiceListWithHeader.kt
```diff
@@ -0,0 +1,118 @@
...
+import com.toggl.settings.compose.theme.grid_2
+import com.toggl.settings.compose.theme.grid_5
+import com.toggl.settings.domain.ChoiceListItem
+import com.toggl.settings.domain.SettingsAction
+import com.toggl.settings.domain.SingleChoiceSettingViewModel
+import kotlinx.coroutines.flow.Flow
+import kotlin.random.Random
+
+@Composable
+internal fun ChoiceListWithHeader(
```
daividssilverio, line 33: I think this should be renamed to indicate that this is a Dialog.

daividssilverio: Or that it might show a dialog 🤔

pdrj:
> I'm not a fan of the narrow dialog, but I won't stop this PR from being merged because of that.

I made it conform to https://material.io/components/dialogs#specs now 🎨 
Screenshot 2020-07-22 at 10 44 25

> I left a small suggestion about the naming of the ChoiceListWithHeader, but I will leave it up to you.

Sure, done.

> One small change request: When tapping the current selected item, the dialog should also be closed. That's based on the old app implementation behaviour.

They do close - are you sure you're testing the latest changes?

daividssilverio:
Hey @pdrj, I just double checked. The calendar doesn't close when you tap on the already selected item. To verify this, just try the workspace selection dialog when you only have one workspace. Or have your first day of the week settings pre-set to `Monday` and try selecting `Monday` again.

daividssilverio:
Everything else looks good to me.

pdrj:
@daividssilverio I made it so the dialog disappears on re-selecting the already-selected option.

pdrj:
>  One tiny detail though: When I first open the app on a fresh install, the default workspace isn't selected 🤔 in the settings view. Not sure if that's a problem that should be fixed here or somewhere else.

Hm, I see that too. I'll create an issue.

pdrj:
#642

PR #600, opened 2020-07-15
Add the calendar date picker & calendar day pager

daividssilverio:
### Summary
This PR is doing the following:
- It adds the calendar date picker and handle emitting its actions
- It will add the calendar day pager and handle emitting the day swiped action
- It fixes some missing tests & behaviour on the date picker reducer

### Technical considerations
I've decided to work on both #492 and #494 at the same time because they will end up touching the same logic and it's all quite coupled.

I've tried using the `ViewPager2` without success :/ It won't animate properly :/ 
For some reason, the same logic I wrote in the `WeekViewHolder` works with the regular pager, but it won't work with the view pager 2. I'm guessing it's something related to lifecycle of the pager view holders.

### Why is it done like this and not some other way?
Feel free to make your suggestions.

### Relationships
Closes #494 
Closes #492 
Closes #582
Closes #601
Closes #493 

### Pull Request checklist
- [x] The code does not hinder the app's performance
- [x] The code does not introduce regressions
- [x] The code in this PR is tested thoroughly enough that changing the behavior to an incorrect one breaks tests
- [ ] One of our [live or file custom templates](https://github.com/toggl/android/tree/master/.idea/templates) was used in this PR

### Pointers
Try it. Compare with the old implementation.

### Review template

- [ ] The code is thoroughly tested in well written and resilient manner
- [ ] The code is performant and leads to an optimal user experience
- [ ] The code is testable and does not cut corners for the sake of delivering
- [ ] The code considers older platforms well and will work in older devices just like it will in new ones

- [ ] The code considers light and dark theme
- [ ] The code considers RTL and does not mention left/right
- [ ] The code considers i18n
  - [ ] There are no hardcoded strings in the UI
  - [ ] All strings are included in every language we support OR there's an issue requesting new translations
- [ ] There are no hard-coded numbers referring to dimensions or font sizes
- [ ] The code considers UI elements that are similar and extracts their similarities into styles to avoid duplication

## Merge permissions
🦑 it after I finish adding the pager for the calendar days.

Issue #497, opened 2020-06-22
Handle hiding the application bottom bar when showing the calendar contextual menu


daividssilverio:
## Description
This is considering we are still gonna do it.
We'll need to handle hiding the application bottom navigation bar while the calendar contextual menu is open.

In theory, this should be easy as observing the state for changes in the selected item while in the calendar tab.
Note that the calendar widget needs to react and have its bounds updated accordingly when the bottom bar changes.

## Dependencies
 #494 #496 

## Definition of done
- [ ] The bottom navigation bar is hidden while the calendar contextual menu is open and visible otherwise

PR #612, opened 2020-07-21
🚧 Add workarounds for StartEdit/Project/ContextualMenu navigation

daividssilverio:
### Summary
This PR adds some workarounds for our navigation solution.
The problems:
- Pressing back in the start edit dialog would crash the app
- Pressing back in the project dialog would crash the app
- Pressing back in the calendar contextual menu would navigate to the main log
- Dismissing the calendar contextual menu wouldn't unselect the calendar selected item

### Technical considerations
Where do I start... take a look at https://issuetracker.google.com/issues/112158843
Basically, our dialog fragments don't have a nav controller set by the time they are dismissed when pressing the back button.
I couldn't figure out if that's just a problem with the `BottomSheetDialogFragment` or if it's because of our router implementation popping the navigation stack in a way the messes things up.

I couldn't think of a better way. But I think our Router could know more about the nav controller navigation stack and take that into consideration when updating our internal `backstack`.
It seems like our problems are caused by having sync those two stacks (the router and the android nav controller's).

Feel free to make your suggestions. Feel free also to propose something else and close this PR if necessary.

### Relationships
Closes Nothing (sorry, I didn't create the issues for this).

### Pull Request checklist

### Pointers
Do a functional test. Don't approve without checking if everything works as expected.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it, ideally after #600 is merged though... 
And the tests are fixed ofc.


timer/src/main/java/com/toggl/timer/project/ui/ProjectDialogFragment.kt
```diff
@@ -72,6 +74,18 @@ class ProjectDialogFragment : BottomSheetDialogFragment() {
...
+    }
+
+    override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+        return super.onCreateDialog(savedInstanceState).also {
+            it.setOnKeyListener { _, _, event ->
+                if (event.action == KeyEvent.ACTION_UP && event.keyCode == KeyEvent.KEYCODE_BACK) {
+                    store.dispatch(ProjectAction.Close)
+                    true
+                } else false
+            }
```
pdrj, line 87: Can we extract it as an extension that just takes as an argument the closure to run on back presses and then use that extension here and in startedit?

calendar/src/main/java/com/toggl/calendar/contextualmenu/domain/ContextualMenuReducer.kt
```diff
@@ -41,7 +44,11 @@ class ContextualMenuReducer @Inject constructor(
         when (action) {
             ContextualMenuAction.DialogDismissed,
             ContextualMenuAction.DiscardButtonTapped,
-            ContextualMenuAction.CloseButtonTapped -> effectOf(ContextualMenuAction.Close)
+            ContextualMenuAction.CloseButtonTapped -> {
+                if (state().backStack.getRouteParam<SelectedCalendarItem>() != null) {
+                    effectOf(ContextualMenuAction.Close)
+                } else noEffect()
```
heytherewill, line 49: I hate these asymmetrical braces (the if has them and the else doesn't. Either remove all curly braces of add them to the else, pls

common-feature/src/main/java/com/toggl/common/feature/navigation/BottomSheetNavigator.kt
```diff
@@ -24,7 +24,10 @@ class BottomSheetNavigator @Inject constructor() : Navigator<BottomSheetNavigato
     }
 
     override fun popBackStack(): Boolean {
-        bottomSheetBehavior?.state = BottomSheetBehavior.STATE_HIDDEN
+        bottomSheetBehavior?.let {
+            if (it.state != BottomSheetBehavior.STATE_HIDDEN)
+                it.state = BottomSheetBehavior.STATE_HIDDEN
+        }
```
heytherewill, line 30: Why does this check even matter? If it's already hidden, nothing will happen 🤔 

Also, since you are not returning anything here, if you decide to keep this if statement surround it in braces

calendar/src/main/java/com/toggl/calendar/calendarday/ui/CalendarDayFragment.kt
```diff
@@ -76,10 +95,38 @@ class CalendarDayFragment : Fragment(R.layout.fragment_calendarday) {
...
+            }
+        }
+        bottomSheetDismissCallback = object : BottomSheetStateCallback() {
+            override fun onStateChanged(bottomSheet: View, newState: Int) {
+                if (newState == BottomSheetBehavior.STATE_HIDDEN) {
+                    behavior.removeBottomSheetCallback(this)
+                    contextualMenuStoreViewModel.dispatch(ContextualMenuAction.DialogDismissed)
+                }
+            }
+        }
```
heytherewill, line 113: Why is this being done here, though? Wouldn't it be possible to add such listeners to 

A) The ContextualMenu fragment
B) The custom navigator

?

timer/src/main/java/com/toggl/timer/project/ui/ProjectDialogFragment.kt
heytherewill: Yes, please. This is really hard on the eyes 😂 

Also, shouldn't you set this to null when clearing the dialog?

daividssilverio: > Can we extract it as an extension that just takes as an argument the closure to run on back presses and then use that extension here and in startedit?

Can do.

> Also, shouldn't you set this to null when clearing the dialog?

Not sure, didn't get any leaks so far.

common-feature/src/main/java/com/toggl/common/feature/navigation/BottomSheetNavigator.kt
daividssilverio: At least during my tests, setting the same value emits a change in the bottomsheet callback, but maybe I was tired. I'll double check.

calendar/src/main/java/com/toggl/calendar/calendarday/ui/CalendarDayFragment.kt
daividssilverio: A) The contextual menu fragment isn't a bottom sheet, it's just a fragment.

B) That I can try again with a fresher mind :) But the first time I tried, I run into feedback loop problems and the fact that the navigator would have to both react to changes to the navigation stack and indirectly trigger changes to it.

calendar/src/main/java/com/toggl/calendar/contextualmenu/ui/ContextualMenuFragment.kt
```diff
@@ -44,6 +53,39 @@ class ContextualMenuFragment : Fragment(R.layout.fragment_contextualmenu) {
...
+    }
+
+    private fun setupBackPressCallback(shouldHandleCallback: Boolean) {
+        onBackPressedCallback?.remove()
+        onBackPressedCallback = requireActivity().onBackPressedDispatcher.addCallback(viewLifecycleOwner, enabled = false) {
+            isEnabled = false
+            store.dispatch(ContextualMenuAction.DialogDismissed)
+        }.also {
+            it.isEnabled = shouldHandleCallback
+        }
```
pdrj, line 88: Why the need for the `also`? Can't we set `isEnabled` in line 83?

calendar/src/main/java/com/toggl/calendar/contextualmenu/domain/ContextualMenuReducer.kt
heytherewill: You still haven't fixed this :P

calendar/src/main/java/com/toggl/calendar/calendarday/ui/CalendarDayFragment.kt
```diff
@@ -26,11 +25,11 @@ import kotlin.contracts.ExperimentalContracts
 @AndroidEntryPoint
 class CalendarDayFragment : Fragment(R.layout.fragment_calendarday) {
 
-    @Inject lateinit var bottomSheetNavigator: BottomSheetNavigator
     @Inject lateinit var calendarItemsSelector: CalendarItemsSelector
     @Inject lateinit var permissionService: PermissionRequesterService
 
     private val store: CalendarDayStoreViewModel by viewModels()
+    private val contextualMenuStoreViewModel: ContextualMenuStoreViewModel by viewModels()
```
heytherewill, line 32: Why is this being imported at all?

daividssilverio: Leftovers

daividssilverio:
> I suppose it's unrelated to this PR, but when testing I noticed that first open of the contextual menu results in empty space where the bottom nav was:

@pdrj yes, it should be handled in #497

calendar/src/main/java/com/toggl/calendar/contextualmenu/ui/ContextualMenuFragment.kt
daividssilverio: Actually, yeah, we don't need the also.

heytherewill:
@pdrj All your concerns seem taken care of; can we merge this? :)

PR #623, opened 2020-07-22
Open browser to show Help in settings

### LINKED ISSUE #526: Settings: Implement Help
## Description
Same as the old app.

## Dependencies
## Definition of done
- [ ] Implement the logic
- [ ] implement the UI

simonrozsival:
### Summary
This PR adds the ability to open a browser at a specific URL.

### Technical considerations
I used the `backStack` state property. I'm not sure this is the best solution since we can't catch the action in some middleware (could hack it somehow with a "higher order reducer", but I'm not sure that's a good solution).

I used the `MainActivity` to open the browser when the top of the back stack points to a Browser URL. I had to handle returning back to the app to pop the browser route from the stack.

I had to deal with the back button issues, because otherwise the app behaved weirdly.

My knowledge of all possibilities of our architecture and Android APIs is still limited, so I'm really not sure if this is a good solution. I'm looking forward to your reviews and I suppose I'll rewrite the PR based on the feedback.

### Relationships
Closes #526 

### Pull Request checklist

### Review template

- [ ] The code considers i18n

## Merge permissions
🚫 not yet


settings/src/main/java/com/toggl/settings/domain/SettingsReducer.kt
```diff
@@ -114,4 +118,9 @@ class SettingsReducer @Inject constructor(
                 it.copy(sendFeedbackRequest = loadable)
             }
         }
+
+    private fun MutableValue<SettingsState>.navigateTo(route: Route) =
+        mutateWithoutEffects<SettingsState, SettingsAction> {
+            copy(backStack = backStack.push(route))
+        }
```
heytherewill, line 125: Nice extension 👌

simonrozsival:
@heytherewill I updated the PR and removed the `BackStackReducer` and the related code. I'm still not sure this is a good way to implement this feature, but at least it's quite simple. I'm not sure where to put the URL right now and it seemed overcomplicated. I'll revisit it tomorrow.

semanticer:
I'm taking over this issue. I don't think we need a separate fragment just to launch a browser. I would simply `startActivity(openURL)` from SettingsFragment and I wouldn't even define a special route for it. `navigateTo` extension looks dope tho, I would keep that for sure.

semanticer:
Closing this since there is a new PR implementing this https://github.com/toggl/android/issues/526

PR #624, opened 2020-07-22
Add calendar settings

### LINKED ISSUE #522: Settings: Implement Calendar Settings. Includes everything that can happen after tapping Calendar Settings
## Description
Check the old app and try to do the same. Talk with the calendar feature champion to coordinate technical decision regarding calendar service etc.

## Dependencies
## Definition of done
- [ ] Implement the logic
- [ ] implement the UI

heytherewill:
### Summary
This PR adds the calendar settings view

### Technical considerations
I've reused the components from the settings page. I've also made that view less ugly by removing separators (gross) and using proper padding so that the ripples take the entire row. It looks gorgeous

I'm on the point of considering ditching fragments altogether for view, since they are not doing much, and instead simply observing the top of the backstack and changing the drawing function 🤷 
Or maybe figuring out a way of having a compose navigator so that a new composable function is provided whenever a navigation occurs?
Dunno, ideas are welcome

### Relationships
Closes #522 

### Pull Request checklist

### Pointers
Go at it commit by commit and please do a functional test

### Review template
- [ ] The code considers i18n

## Merge permissions
Anyone, anytime


settings/src/main/java/com/toggl/settings/ui/composables/pages/CalendarSettingsPage.kt
```diff
@@ -0,0 +1,126 @@
...
+    calendarSettingsViewModels: List<CalendarSettingsViewModel>,
+    pageTitle: String,
+    dispatcher: (SettingsAction) -> Unit
+) {
+    Scaffold(
+        topBar = {
+            TopAppBar(
+                backgroundColor = MaterialTheme.colors.surface,
+                contentColor = MaterialTheme.colors.onSurface,
+                title = { Text(text = pageTitle) }
```
semanticer, line 58: You should add back arrow icon and back action. 
Check out https://github.com/toggl/android/pull/628/files#diff-dba56d9789764e4cdf1e2b8be8f5ad14

settings/src/main/java/com/toggl/settings/ui/composables/pages/CalendarSettingsPage.kt
```diff
@@ -0,0 +1,126 @@
...
+                            viewModel.accessGranted
+                        )
+                        Column {
+                            SettingsRow(setting, dispatcher)
+                            Text(
+                                text = context.getString(R.string.allow_calendar_message),
+                                style = MaterialTheme.typography.body2,
+                                modifier = Modifier.padding(grid_2)
+                            )
+                        }
```
semanticer, line 78: I would probably extract this just to keep this `when` statement more "high-level" without too much layout detail.

settings/src/main/java/com/toggl/settings/domain/CalendarSettingsSelector.kt
```diff
@@ -0,0 +1,39 @@
...
+class CalendarSettingsSelector @Inject constructor(
+    private val calendarService: CalendarService
+) : Selector<SettingsState, List<CalendarSettingsViewModel>> {
+    override suspend fun select(state: SettingsState): List<CalendarSettingsViewModel> {
+
+        val userCalendars = calendarService.getUserSelectedCalendars()
+        val availableCalendars = calendarService.getAvailableCalendars()
+
+        return sequence {
+            val accessGranted = !state.userPreferences.calendarIntegrationEnabled
```
daividssilverio, line 19: `accessGranted` isn't right. It should've been `calendarIntegrationEnabled`.

daividssilverio: Also, this needs to be merged with the actual system calendar permission granted status, shouldn't it?

daividssilverio: Also, why is this negated? The user prefs says the integration is disabled, the calendar permissions weren't given and yet the switch is on on the settings screen.

settings/src/main/java/com/toggl/settings/domain/SettingsViewModel.kt
```diff
@@ -13,3 +13,8 @@ sealed class SettingsViewModel {
 }
 
 data class SettingsSectionViewModel(val title: String, val settingsOptions: List<SettingsViewModel>)
+
+sealed class CalendarSettingsViewModel {
+    data class AccessGranted(val accessGranted: Boolean) : CalendarSettingsViewModel()
```
daividssilverio, line 18: 🤔 Should've been `IntegrationEnabled` or something like that. Not sure if we should include both the information about the integration enabled (user prefs) and the system calendar permission (access granted) or simply merge them into one name.

daividssilverio:
Some problems I found while testing:
- The app navigates to the main log when a permission is given/denied.
- ~There's still no "permission denied flow" handling. What I mean is, when the user denies the calendar permission when trying to access the calendar settings, we should show a dialog explaining the why we need this permission and that the app still works without it.~ Actually, I just double checked the current app, and it doesn't work that way there as well. The dialog only appears when the calendar permission is denied from the calendar page.
- The smart alerts section still appears when the calendar integration/system permissions are not enabled/granted.
- The composed fragments now have a weird top margin/padding above the top bar.

settings/src/main/java/com/toggl/settings/ui/composables/pages/SettingsPage.kt
```diff
@@ -1,71 +1,71 @@
...
     sectionsState: List<SettingsSectionViewModel>,
-    pageTitle: String,
+    statusBarHeight: Dp,
+    navigationBarHeight: Dp,
     dispatcher: (SettingsAction) -> Unit = {}
 ) {
     Scaffold(
         topBar = {
             TopAppBar(
+                modifier = Modifier.padding(top = statusBarHeight),
```
daividssilverio, line 58: I'm guessing we don't need this padding.

daividssilverio:
Alright, the weird top padding is fixed. ✅ 
Couple of details:
- The up navigation isn't working on the calendar settings page
- The calendars events aren't appearing in the calendar after you enable them in the settings page. But I'm guessing we can fix that somewhere else.

Issue #525, opened 2020-06-29
Settings: Implement About


sofokli1:
## Description
Should include the same options as the old app
- Privacy policy
- Terms of Service
- Licenses

[iOS Designs page.](https://toggl.invisionapp.com/d/main?origin=v7#/projects/prototypes/19800511) [Android Designs page.](https://toggl.invisionapp.com/d/main?origin=v7#/projects/prototypes/19932960) [Notion page.](https://www.notion.so/toggl/Settings)

## Dependencies
`Settings: Add settings view` #512
## Definition of done
- [ ] Implement the logic
- [ ] implement the UI

heytherewill:
Isn't this already done?

sofokli1:
so it seems. @semanticer are you doing anything here?

semanticer:
No, it should be done 🤷‍♂️

PR #627, opened 2020-07-23
Settings: About page

simonrozsival:
### Summary
This PR adds the _About_ page and it allows the user to navigate to that page (fragment) from the Settings menu.

### Technical considerations
I used Compose and it works nicely. The page is static so it doesn't need anything special, it just emits actions when any of the buttons is pressed. What isn't implemented is reaction to the button presses. Two of them will open the browser (that depends on #623) and the last one opens a separate Licenses view, ~which I'll add in a separate PR later (it's also gonna be a static list of things)~.

Currently when the back button is pressed, you'll get back to the _Settings_ page, but the `backStack` property of the state is broken. I'm not sure how to handle this, because I can't even detect the back btn in the fragment AFAIK and even if I did, there's no "back" action and the settings reducer is currently not a closable one.. and I don't know even if that would later work with two or three layers of nested fragments (Settings -> About -> Licenses). I suppose we should fix this _back buttton_ handling globally and not ad hoc in this PR.

What is still missing:
- [ ] opening browser (terms of service, privacy policy)
- [ ] an actual list of licenses for libraries we use
- [ ] correct system back button integration
- [ ] correct behavior of the title-bar back buttons

🤷 

### Relationships
Ref #525

### Pull Request checklist

### Review template

- [ ] The code considers i18n

## Merge permissions
🚫 this PR is just a draft for now, I'm still working on it.


semanticer:
I'll submit a new PR closing both, help and about issues

PR #628, opened 2020-07-23
🔙 Feedback Screen

### LINKED ISSUE #524: Settings: Implement Submit Feedback
## Description
This will be presented with a push navigation instead of sheet presentation. But instead of a back button there will be a `cancel` button. Also a `save` button on the other side. Check designs.

## Dependencies
## Definition of done
- [x] Implement the logic
- [x] implement the UI

semanticer:
### Summary
This PR implements feedback screen. 

### Technical considerations
As we already discussed on Slack I decided to implement this screen as a separate fragment with its own Route and navigation entry point however the UI is still build using Compose. 

I've added a new action `BackToMainScreenTapped` to handle returning from Feedback screen ant it will be useful for About screen as well. If you have a better idea of how to solve this, please let me know.

The UI is preliminary because we are still waiting for Android designs. 
Although I've managed to disable/enable Send button based on the loading state, I have no idea how to disable/enable TextField - probably something we will have to implement later with newer versions of Compose.

### Relationships

Closes #524

### Pull Request checklist

### Pointers
It's fairly small but you can go commit by commit to make it even easier. Change `SendFeedbackEffect` to send `SettingsAction.FeedbackSent` if you want to see what happens after successful feedback request finishes. 

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


settings/src/main/java/com/toggl/settings/domain/SettingsReducer.kt
```diff
@@ -46,14 +49,22 @@ class SettingsReducer @Inject constructor(
...
                 SettingsType.PrivacyPolicy -> TODO()
                 SettingsType.TermsOfService -> TODO()
                 SettingsType.Licenses -> TODO()
                 SettingsType.Help -> TODO()
                 SettingsType.SignOut -> effectOf(SettingsAction.SignOutTapped)
             }
+            SettingsAction.BackToMainScreenTapped -> {
+                if (state().backStack.last() !is Route.Settings) {
+                    state.mutate {
+                        copy(backStack = backStack.pop())
```
daividssilverio, line 63: Ideally, we should also clean the `sendFeedbackRequest` state when navigating back and forth to the send feedback page 🤔

semanticer: Good point, I added un-initialization, however, we should also cancel any ongoing effect which is something we haven't figure out yet right?

daividssilverio: Yup, we need to do something about it. We could maybe add a reference to the task that was set to run in the Loadable 🤔, keep it in the state so that we can later cancel it? Something to think about. @heytherewill @semanticer

settings/src/main/java/com/toggl/settings/ui/composables/FeedbackPage.kt
```diff
@@ -71,13 +75,20 @@ fun FeedbackForm(
...
-            modifier = Modifier.fillMaxWidth().preferredHeight(140.dp)
-        )
+        Stack(modifier = Modifier.fillMaxWidth().preferredHeight(140.dp)) {
+            OutlinedTextField(
+                value = textState,
+                onValueChange = { textState = it },
+                label = { Text(text = stringResource(R.string.submit_feedback)) },
+                modifier = Modifier.matchParentSize()
+            )
+            if (sendFeedbackRequestState is Loadable.Loading) {
```
daividssilverio, line 87: Compose is so cool <3

settings/src/main/java/com/toggl/settings/ui/FeedbackFragment.kt
```diff
@@ -29,14 +36,27 @@ class FeedbackFragment : Fragment() {
...
                 onFeedbackSent = { feedbackText ->
                     store.dispatch(SettingsAction.SendFeedbackTapped(feedbackText))
                 }
             )
+
+            sendFeedbackRequest
+                .onEach { request ->
+                    val msg = when (request) {
+                        is Loadable.Loaded -> getString(R.string.feedback_thank_you)
+                        is Loadable.Error -> request.failure.errorMessage
```
daividssilverio, line 53: Actually, the error message should've been a view instead, not a toast. I guess a toast is fine for now. Not sure if the designs for this screen will change ever.

The current implementation looks like this.

image

ping @heytherewill

PR #645, opened 2020-07-24
🔥 Remove optics to update compose

daividssilverio:
### Summary
This is removing ours optics so that we can compile the apps using the newest compose version.

### Technical considerations
It looks messier, more copies around.
Idk... but it also feels faster... check it for yourself.

Make your suggestion.

### Relationships
Closes #Nothing

### Pull Request checklist

### Pointers
You might need to update the bundled kotlin plugin in your AndroidStudio to use 1.4 as well.

### Review template
Compile and run the project. See if everything still works.
The tests should still pass alright.

- [ ] The code considers i18n

## Merge permissions
🦑 after at least two reviewers ✅ it


daividssilverio:
Or we can merge this into @semanticer's branch and review directly from there. Up to you @semanticer .

PR #669, opened 2020-07-26
🚧 Handle contextual menu & bottom nav bar interactions

### LINKED ISSUE #496: Handle contextual menu & calendar day pages bottom paddings
## Description
We need to handle adjusting the visible frame of the calendar day widget when the contextual menu is visible. Whenever the contextual menu slide offset changes, the calendar day needs to adjust itself so that its contents are reachable.

This issue was labeled as `size-M` due to bad experience we had while implementing the first version. The idea is to have the contextual menu at the same hierarchy level as the calendar day pages, which might cause some problems or could make it simpler.

## Dependencies
#494 #372

## Definition of done
- [ ] It's possible to scroll the whole calendar widget when the contextual menu is visible

### LINKED ISSUE #497: Handle hiding the application bottom bar when showing the calendar contextual menu

daividssilverio:
### Summary
This PR fixes the behaviour of the calendar widget when its size changes in order to make it play well with bottom navigation visibility changes when the contextual menu appears.

### Technical considerations
The delay added to the bottom sheet navigator was added so that the calendar fragment has time to have its layout bounds updated when the bottom navigation bar visibility is changed. The delay value was arbitrary.

Feel free to give your suggestion,

### Relationships
Closes #496 
Closes #497

### Pull Request checklist

### Pointers
You may review by commit, but there are not a lot of changes.
Please do a functional test.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it


common-feature/src/main/java/com/toggl/common/feature/navigation/BottomSheetNavigator.kt
```diff
@@ -26,8 +29,10 @@ class BottomSheetNavigator @Inject constructor() : Navigator<BottomSheetNavigato
         navigatorExtras: Extras?
     ): NavDestination? {
         bottomSheetBehavior?.let {
-            it.state = BottomSheetBehavior.STATE_EXPANDED
-            it.addBottomSheetCallback(dismissCallback)
+            handler.postDelayed({
+                it.state = BottomSheetBehavior.STATE_EXPANDED
+                it.addBottomSheetCallback(dismissCallback)
+            }, 150)
```
heytherewill, line 35: Extract as a constant inside the class

calendar/src/main/java/com/toggl/calendar/contextualmenu/ui/ContextualMenuStateViewModel.kt
```diff
@@ -0,0 +1,8 @@
+package com.toggl.calendar.contextualmenu.ui
+
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.ViewModel
+
+class ContextualMenuStateViewModel : ViewModel() {
+    val slideOffsetInPixels = MutableLiveData<Int>(0)
```
heytherewill, line 7: Couldn't you have added this data to the already existing viewModel, which is currently empty and just a store?

heytherewill: Also, why live data and a MutableStateFlow, since we're fully commited to the Flow Api?

daividssilverio: > Couldn't you have added this data to the already existing viewModel, which is currently empty and just a store?

Yes, I can, but should I? Not sure if we should add completely UI only stuff to something that is related to our domain.

> Also, why live data and a MutableStateFlow, since we're fully committed to the Flow Api? 

Ease of use. I don't see why not use both if it makes our lives easier. But I can look into converting it to flow. It was just so easy to use.

heytherewill: The Store is related to the domain, yes, but the ViewModel is not the store

I think using a single api has the benefit of making combining simpler and removes the cognitive load of figuring out what's the correct way of subscribing. If we could have a linter to prevent live data from existing at all, I'd add it 😄

daividssilverio: Done, live data 🔥 ed. `MutableStateFlow` was pretty much the same thing.

PR #673, opened 2020-07-27
✨ Add smart alerts settings, update ktlint

### LINKED ISSUE #523: Settings: Implement Smart Alerts setting
## Description
I suggest using the `Settings: Implement select from list sub-setting` (check dependencies below for a link) to show the selectable options. And to use the type 1 cells created in `Settings: Add settings view` issue

## Dependencies
`Settings: Implement select from list sub-setting`: #517
## Definition of done
- [ ] Implement the logic
- [ ] implement the UI

pdrj:
### Summary
Adds smart alerts settings

### Technical considerations
It's treated as a single-choice setting. I see in the rebrand iOS designs (https://toggl.invisionapp.com/d/main?origin=v7#/projects/prototypes/19800511) that it is a two-step process there, but I don't see a reason for it ATM, plus we don't have any Android designs.

### Relationships

Closes #523 

### Pull Request checklist

### Pointers
It's pretty straightforward. Play around with it.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


pdrj:
This had to be changed a bit because of the reworked settings navigation in #624 

Because of these changes, there's now a weird blink when setting a value in a dialog and the scroll position in settings list is not remembered :(. 

I've also reused the settings deep link to navigate to the dialogs, not sure if that's the way to go 🤔

daividssilverio:
> Because of these changes, there's now a weird blink when setting a value in a dialog and the scroll position in settings list is not remembered :(.

Are we missing some equals implementation or something like that? It looks like a full recomposition is happening when it shouldn't. 🤔

pdrj:
Yeah it looks like it @daividssilverio. I'll see if I can do something about it, but I'd rather not spend too much time on it.

pdrj:
I've found a solution @daividssilverio. Because of the switch back to using routes to show the compose dialogs, we can no longer show them in a 'react-like' way, embedded in the settings fragment :( I'll push some changes soon that use a separate dialog fragment holder (with a separate route) to show the compose dialogs.

pdrj:
@daividssilverio see how you like the fix. IMO it's OK to not have the embedded composable and we shouldn't spend more time on it.

The last commit is ktlint update - otherwise it complains about the extension that uses @Composable in lambda.

settings/src/main/java/com/toggl/settings/compose/extensions/FragmentExtensions.kt
```diff
@@ -3,28 +3,35 @@ package com.toggl.settings.compose.extensions
...
-    with(composeContainer) {
-        isClickable = false
-        isFocusable = false
-        layoutParams = ViewGroup.LayoutParams(ViewGroup.LayoutParams.MATCH_PARENT, ViewGroup.LayoutParams.MATCH_PARENT)
+}
 
-        doOnInsetsChanged { statusBarHeight, navigationBarHeight ->
-            callback(statusBarHeight.pixelsToDp(context).dp, navigationBarHeight.pixelsToDp(context).dp)
-        }
+fun Fragment.createComposeDialog(
```
daividssilverio, line 24: 🤔 this isn't creating a dialog per se... it's just not using the `doOnInsetsChanged` part of the other extension, right? Consider renaming it to something else.

pdrj:
> Please undo the newline changes

Can't really undo them because the lint task will fail 😞. I can modify the config to not care about newlines at end of the file.

pdrj:
@heytherewill I disabled `final-newline,import-ordering` rules in ktlint and you can see the resulting diff. If you want to limit it further, that would mean disabling `indent` rule https://github.com/pinterest/ktlint/blob/master/ktlint-ruleset-standard/src/main/kotlin/com/pinterest/ktlint/ruleset/standard/IndentationRule.kt and I'm pretty sure we don't want that.

The other options are disabling spotless entirely, or using something more configurable. Even though ktlint supports editorconfig, when wrapped in spotless, it doesn't support it anymore 🤷 

If we want to keep using ktlint+spotless combo, we'd have to update it at some point anyway.

settings/src/main/java/com/toggl/settings/compose/extensions/FragmentExtensions.kt
pdrj: Yep, that is the case. I renamed it to createComposeFullscreenView.

pdrj:
Superseded by #698

PR #675, opened 2020-07-28
Update compose, kotlin and migrate tests to the truth

daividssilverio:
### Summary
This is one way to go about updating compose.

### Technical considerations
Well... Truth doesn't look great. As I was going about migrating the tests... I started to see how nice the infix notations were... 
And how I miss the collection assertions... I'm not sure if using the Truth is the way to go...

With the Kotlin gradle plugin 1.4+, we don't need to specify the stdlib for our android projects.
I think we don't need to specify it for the kotlin-only projecs as well, but I couldn't make it work.

If we we manage to fix up the usage of kotest assertions instead (or maybe use some other library), I think I'd be ok with thrashing this PR.

### Relationships
Closes #Nothing

### Pull Request checklist

### Pointers
- Take a look at some of the tests, see if you think we should go that way.
- Run the app, see if it works properly.
- Open the project on AS, do a clean build and see if there are any problems.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it, but only if you want to write tests using the Truth. I'm not sure anymore.
Also, let me convert the old Spec tests before proceeding as well.


daividssilverio:
🤔 the build was quite slower on bitrise in this branch for some reason... Make sure to check your local build time.

heytherewill:
If this is the Truth I wanna live IN DENIAL

Issue #605, opened 2020-07-17
Figure out a better structure for our nav graph


heytherewill:
## Description
We need to figure out a better way of defining the starting destination of our app. Right now, it's the main log, which means there are some glitches when the user is logged out.

## Dependencies
Follow up of #598

## Definition of done
- [ ] Our nav graph is more consistent
- [ ] The user is unable to see the main log if they open the app, send it to the background and then open it again

PR #681, opened 2020-07-28
🔨 Add and wire nav between login & signup actions

### LINKED ISSUE #635: Implement GoToLoginAction
## Description
Implement the GoToLoginAction, wire it up to the SignUp UI
## Dependencies
#631 
## Definition of done

Action is added
Tests are added
Action is wired to the UI

### LINKED ISSUE #636: Implement GoToSignUpAction
## Description
Implement the GoToSignUpAction, wire it up to the Login UI
## Definition of done
Action is added
Tests are added
Action is wired to the UI

daividssilverio:
### Summary
This adds the boilerplate to navigate between the sign up and login screens.

### Technical considerations
I didn't push/pop the backstack to make the navigation happen to avoid creating unnecessary navigation ([Welcome, Login, SignUp, Login, SignUp] for example).

It seemed like the right thing to do. Feel free to make your suggestion.

### Relationships
Closes #635
Closes #636 

### Pull Request checklist

### Pointers
Review by commit
Test it functionally

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it.


onboarding/src/main/java/com/toggl/onboarding/signup/domain/SignUpReducer.kt
```diff
@@ -15,5 +18,8 @@ class SignUpReducer @Inject constructor() : Reducer<SignUpState, SignUpAction> {
             SignUpAction.SignUpButtonTapped -> TODO()
             is SignUpAction.EmailEntered -> TODO()
             is SignUpAction.PasswordEntered -> TODO()
+            SignUpAction.GoToLoginTapped -> state.mutateWithoutEffects {
+                copy(backStack = backStackOf(Route.Welcome, Route.Login))
```
pdrj, line 22: How do you feel about, rather than replacing the whole `backStack`, have an extension `BackStack.navigateAndPop` or `BackStack.replaceTop`? The resulting `backStack` is going to be the same of course, but we would keep the manual backstack creation hidden.

pdrj:
Should we also add some animations when switching between the screens? It feels kinda unnatural now 🤔

onboarding/src/main/java/com/toggl/onboarding/signup/domain/SignUpReducer.kt
daividssilverio: 🤔 I think we can do that. But there's no guarantee the `Welcome` screen will be there, so I think we still need to recreate the stack. Again, a problem with handling two stacks.

daividssilverio:
> The change request is for the fact that the email value is not passed between the screens.

🤔 The backstack being set to `[Welcome, (Login|SignUp)]` shouldn't make the onboarding state change, so the e-mail value should still be there and appear on the screen.

The detail is that we didn't wire it up it yet on either screen.

daividssilverio:
> Should we also add some animations when switching between the screens? It feels kinda unnatural now 🤔

Yes. I'll try to figure it out using our navigation graph, but it looks like something isn't setup right to make it work out of the box.

onboarding/src/main/java/com/toggl/onboarding/signup/domain/SignUpReducer.kt
daividssilverio: Can we maybe come back at this later? We need to do some refactoring in our nav graph & router to be able to pop the stack safely. I suppose #605 should handle that.

daividssilverio:
@pdrj Also, let's go back to the navigation's transitions when the layouts are final? Not sure if they are going to be modals or not.

onboarding/src/main/java/com/toggl/onboarding/signup/domain/SignUpReducer.kt
pdrj: Sure, let's leave it as is.

pdrj:
> The detail is that we didn't wire it up it yet on either screen.

Login is wired up now from what I see, can we do so for the signup as well?

daividssilverio:
> Login is wired up now from what I see, can we do so for the signup as well?

Might as well, should've done that from the start.

pdrj:
I hope you don't mind my commit @daividssilverio, I need these changes in my branch (so in master too) to not duplicate work.

PR #690, opened 2020-07-31
✨ Add user preferences subscription

### LINKED ISSUE #617: Implement user preferences subscription
## Description
Implement a subscription that will constantly update user preferences whenever there is a change in the database. 

## Dependencies
#329 

## Definition of done
- [ ] Refactor dao and repository to provide Flow 
- [ ] Implement subscription that will provide continuous user preferences updates
- [ ] 🔥 Old code that was responsible for userPrefs updates (`LoadUserPreferencesEffect` etc)
- [ ] Add tests

daividssilverio:
### Summary
This adds the subscription for the user preferences o/

### Technical considerations
Not really, the user preferences were already added in the AppState, I just had to use them in other states and migrate some services that were using the settings repository directly to use the state instead.

Feel free to give your suggestions.

### Relationships
Closes #617

### Pull Request checklist

### Pointers
You may review by commit.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it


repository/src/main/java/com/toggl/repository/Repository.kt
```diff
@@ -72,6 +74,18 @@ class Repository(
...
+    private val sharedPreferencesListener: SharedPreferences.OnSharedPreferenceChangeListener =
+        SharedPreferences.OnSharedPreferenceChangeListener { _, _ ->
+            sharedPreferencesFlow.value = readUserPreferences(sharedPreferences)
+        }
+
+    @ExperimentalCoroutinesApi
+    private val sharedPreferencesFlow: MutableStateFlow<UserPreferences> by lazy {
+        sharedPreferences.registerOnSharedPreferenceChangeListener(sharedPreferencesListener)
+        MutableStateFlow(readUserPreferences(sharedPreferences))
+    }
```
semanticer, line 87: You should probably also call `unregisterOnSharedPreferenceChangeListener` somewhere but since this code will be probably 🔥 soon when we move shared preferences to DB it doesn't really matter.

daividssilverio: Actually, the docs says they don't hold a strong reference to the listener, so everything should be alright.
Anyway, if we are moving to the DB, this doesn't really matter as you said.

PR #699, opened 2020-07-31
🔒 Add strong password indicators

### LINKED ISSUE #632: Add password strength indicators to signup screen
## Description
Add password strength indicators. Check out Xamarin app to see what they are and how they behave. The specs can be found in the old issues here https://github.com/toggl/mobileapp/projects/106
## Dependencies
#631 
## Definition of done
- Password strength criteria are added and behave in the same way as in Xamarin

pdrj:
### Summary
Adds strong password indicators

### Technical considerations
I had to make the Password classes not data classes, otherwise the password value on login screen is initially set to "Invalid("")", aka the string representation of invalid 😄 

### Relationships

Closes #632 

### Pull Request checklist

### Pointers
Validate it behaves the same as in xamarin app. 
Validate you can still log in.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


onboarding/src/main/java/com/toggl/onboarding/login/domain/LoginReducer.kt
```diff
@@ -36,9 +36,9 @@ class LoginReducer @Inject constructor(
                     is Email.Invalid -> noEffect()
                     is Email.Valid -> when (val password = currentState.password) {
                         is Password.Invalid -> noEffect()
-                        is Password.Valid -> {
+                        is Password.Valid, is Password.Strong -> {
```
heytherewill, line 39: A Strong password is already valid, I don't think this change is even needed 🤔

onboarding/src/main/java/com/toggl/onboarding/signup/ui/SignUpFragment.kt
```diff
@@ -36,6 +44,24 @@ class SignUpFragment : Fragment(R.layout.fragment_signup) {
...
+
+        val defaultCriteriaColor = ContextCompat.getColor(requireContext(), R.color.authentication_text)
+        val errorCriteriaColor = ContextCompat.getColor(requireContext(), R.color.error)
+
+        val isPasswordStrongChannel = Channel<Boolean>()
+
+        passwordStrengthWatcher = password.doAfterTextChanged { isPasswordStrongChannel.offer(it.representsStrongPassword()) }
+
+        isPasswordStrongChannel
+            .consumeAsFlow()
```
heytherewill, line 56: Why is this at all needed when the password already is in the state? 🤔 

```suggestion
        store.state
            .map { it.password is Password.Strong }
```

onboarding/src/main/java/com/toggl/onboarding/login/domain/LoginReducer.kt
pdrj: That was my understanding as well, but I'm pretty sure I couldn't login with a strong pass last week 😕 Today I can though, so I'll roll this change back 😄

models/src/main/java/com/toggl/models/validation/Password.kt
```diff
@@ -3,13 +3,13 @@ package com.toggl.models.validation
 sealed class Password {
     abstract val password: String
 
-    data class Invalid internal constructor(override val password: String) : Password()
+    class Invalid internal constructor(override val password: String) : Password()
     open class Valid internal constructor(override val password: String) : Password() {
         override fun equals(other: Any?) = other is Valid && other.password == password
 
         override fun hashCode() = password.hashCode()
```
heytherewill, line 10: I think the equality comparison needs to be on the strong password as well

pdrj: https://toggl.slack.com/archives/CD40QCV25/p1596452350195700

PR #706, opened 2020-08-04
🔁 Moshi, Base Api models and mock data

### LINKED ISSUE #693: Move away from Gson to Moshi
## Description
We decided to move away from Gson and replace it with Moshi. Preferably we want to use Moshi's code-gen variant and get rid of the reflection code. 

## Dependencies
none

## Definition of done
- [ ] Add Moshi dependency
- [ ] Preferably use Moshi code-gen variant instead of reflection
- [ ] Refactor existing Gson parsing to Moshi
- [ ] GSON dependency is completely removed from the project

### LINKED ISSUE #694: Create API models for base entities
## Description
Create API model versions of our base(database) entities. This should already done with Moshi annotations. 

## Dependencies
#693

## Definition of done
- [ ] Add `ApiClient`
- [ ] Add `ApiProject`
- [ ] Add `ApiTag`
- [ ] Add `ApiTask`
- [ ] Add `ApiTimeEntry`
- [ ] Add `ApiUser`
- [ ] Add `ApiWorkspace`

### LINKED ISSUE #695: Provide easy access to mock database data for testing
## Description
Ideally, we should add settings section that will appear only in `debug` variant of the app and will provide debug-related options. The first thing we should add to this section is option to insert the given number or TimeEntries from test data json files to our database. 

## Dependencies
#694 #693 

## Definition of done
- [ ] Add json test data files to project's asset folder. They should be included only in the `debug` build variant
- [ ] Add parsing logic to import data from JSON files directly to the database 
- [ ] Add debug settings section
- [ ] Add a way to specify the number of `TimeEntries` to import

### LINKED ISSUE #708: Sync pull response deserialization
## Description
Prepare necessary logic for pull response deserialization

## Dependencies
#694  #693 

## Definition of done
- [ ] Add necessary objects for deserialization
- [ ] Add tests

semanticer:
### Summary
This PR

- handles the transition to Moshi
- implements basic API models
- handles pull response deserialization
- implements a way to import mock data
- adds debug settings section 

### Technical considerations
We are using Moshi's code-generation kotlin flavor which should have the fastest runtime performance. 
Mock data pull-response file is humongous but it's included only in debug version build.
I'm intentionally parsing only properties that are relevant for mobile apps.
`ApiToDatabaseExtensions.kt ` is probably only preliminary or it should be used for mock data import only, the real sync will be doing something much more sophisticated. 

Let me know in case you have any ideas on how to make this even simpler. 

### Relationships

Closes #693 
Closes #694 
Closes #695 
Closes #708

### Pull Request checklist

### Pointers
Make sure we are parsing everything that we need.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


api/src/main/java/com/toggl/api/models/ApiTag.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import com.squareup.moshi.Json
+
+
+@JsonClass(generateAdapter = true)
+data class ApiTag(
+    val at: String,
+    val id: Int,
+    val name: String,
+    @Json(name = "workspace_id")
```
heytherewill, line 11: Isn't there a way to tell moshi to use snake_case everywhere?

semanticer: Nope, it's against its philosophy: https://github.com/square/moshi/issues/644 and https://publicobject.com/2016/01/20/strict-naming-conventions-are-a-liability/

Also, it wouldn't work for us - the API is inconsistent in many places, take a look at `ApiPreferences` for example.

PR #721, opened 2020-08-07
Make database time entries syncable and extend sync properties

### LINKED ISSUE #671: Make DatabaseTimeEntry properties syncable
## Description
Refactor all DatabaseTimeEntry properties to `*SyncProperty` types with the exception of `Id` and `serverId`

## Dependencies
#672  

## Definition of done
- [ ] Refactor all properties to corresponding `*SyncProperty` variants.
- [ ] Update all converter extension methods and make sure our domain models stay the same and are using "current" variant of `SyncProperty`

semanticer:
### Summary
This makes our DatabaseTimeEntries ready for the syncing age. 

### Technical considerations
Since the DatabaseTimeEntry is now superset of regular domain TimeEntry we need to change repository a little bit where we need to first query current DatabaseTimeEntry and update it with the new TimeEntry domain model rather than simply transforming TimeEntry to DatabaseTimeEntry as we did before.

I'm not yet sure how to handle relation to tags. Maybe I should do this in #711 ?

Unfortunately, I couldn't find a way how to make some abstraction over `*SyncProperties` so there is some repetitive code. Room and the fact that data classes with their `copy` method are not implementing any common interface is making this abstraction very hard to do, but let me know if you have any suggestions.  

### Relationships

Closes #671 

### Pull Request checklist

### Pointers
I've made some effort to separate commits logically so you can take advantage of that.  

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


repository/src/main/java/com/toggl/repository/Repository.kt
```diff
@@ -154,12 +156,20 @@ class Repository(
     }
 
     override suspend fun editTimeEntry(timeEntry: TimeEntry): TimeEntry {
-        return timeEntryDao.updateTimeEntryWithTags(timeEntry.toDatabaseModel()).run { timeEntry }
+        val oldDatabaseTimeEntry = timeEntryDao.getOneTimeEntryWithTags(timeEntry.id)
+        val newDatabaseTimeEntry = oldDatabaseTimeEntry updateWith timeEntry
```
heytherewill, line 160: That's :sogood:

PR #725, opened 2020-08-10
💭 Suggestions subscription

### LINKED ISSUE #639: Implement suggestions subscription
## Description
Implement a subscription that will constantly update suggestions list whenever any of the relevant entities changes

## Dependencies
#329 

## Definition of done
- [ ] Refactor existing code to provide Flow of suggestions that updates automatically
- [ ] Implement subscription that will provide continuous suggestion list updates
- [ ] 🔥 Old code that was responsible for suggestions updates
- [ ] Add tests

semanticer:
### Summary
This PR is implementing suggestions subscriptions and cleaning original implementation.

### Technical considerations
I guess it's pretty straightforward. I moved suggestions to top level AppState because that's how our subscriptions are implemented right now.

🤷‍♂️ 

### Relationships
Closes #639

### Pull Request checklist

### Pointers
Going commit by commit should make the review easier. Use mock data import from settings to hopefully see some real suggestions. 

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


timer/src/main/java/com/toggl/timer/suggestions/domain/SuggestionProvider.kt
```diff
@@ -1,16 +1,27 @@
...
+import com.toggl.common.feature.services.calendar.CalendarEvent
+import com.toggl.models.domain.Project
+import com.toggl.models.domain.TimeEntry
+
 interface SuggestionProvider {
-    suspend fun getSuggestions(suggestionsState: SuggestionsState): List<Suggestion>
+    suspend fun getSuggestions(suggestionSources: SuggestionSources): List<Suggestion>
 }
 
+data class SuggestionSources(
```
heytherewill, line 11: This name is a tad ambiguous. I think SuggestionData might work better

PR #728, opened 2020-08-12
Sync pull endpoint

### LINKED ISSUE #717: ⬇️ Add sync Pull request
## Description
We need to define retrofit interface and client for sync endpoints and add pull method.
TODO add docs link when it's accessible 

## Definition of done
- [x] Add sync Retrofit interface
- [x] Define pull method in sync Retrofit interface
- [x] Add sync interface client and its implementation
- [x] Wire everything up in DI module
- [ ] Add tests

heytherewill:
### Summary
This adds the pull sync endpoint to our apps

### Technical considerations
Not much, really. The deserialization was already there, so this one was simpler than anticipated. I based it on the Xamarin implementation; apparently we need that extra header

### Relationships
Closes #717 

### Pull Request checklist

### Pointers
Go for it commit by commit

### Review template

## Merge permissions
Anyone, anytime


api/src/main/java/com/toggl/api/di/ApiModule.kt
```diff
@@ -44,6 +47,13 @@ object ApiModule {
     @BaseReportsUrl
     fun baseReportsUrl(@BaseUrl baseUrl: String): String = "$baseUrl/reports/api/v3/"
 
+    @Provides
+    @Singleton
+    @BaseSyncUrl
+    fun baseSyncUrl(): String =
+        if (AppBuildConfig.isBuildTypeRelease) "https://sync.toggl.com"
+        else "https://sync.toggl.space"
```
semanticer, line 55: I think we should extract all these build/flavor dependent constants into some config class that will be defined for every version of our app - alternatively, we can put these constants to Gradle as well.

PR #729, opened 2020-08-12
📅 Calendar Events Subscription

### LINKED ISSUE #641: Implement calendar events subscription
## Description
Implement a subscription that will constantly update calendar event list whenever anything changes

## Dependencies
#329 

## Definition of done
- [ ] Refactor existing code to provide Flow of calendar events
- [ ] Implement subscription that will provide continuous calendar event list updates
- [ ] 🔥 Old code that was responsible for calendar event updates
- [ ] Add tests

semanticer:
### Summary
This implements subscription for updating our calendar events list

### Relationships
Closes #641
 
Refs #731 

### Pull Request checklist

### Pointers
Our permission logic is unfortunately in deep 💩 so you have to manually change `val calendarPermissionWasGranted: Boolean = false,` in AppState to `true` to see some calendar events. I will create a separate issue for fixing permissions.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


app/src/main/java/com/toggl/domain/loading/LoadCalendarEventsSubscription.kt
```diff
@@ -0,0 +1,143 @@
...
+import java.time.OffsetDateTime
+import java.time.ZoneOffset
+
+private class CalendarEventsContentObserver(
+    private val contentResolver: ContentResolver,
+    private val fromStartDate: OffsetDateTime,
+    private val toEndDate: OffsetDateTime,
+    private val fromCalendars: List<Calendar>,
+    private val calendarEventsFlow: MutableStateFlow<List<CalendarEvent>>
+) : ContentObserver(null) {
```
heytherewill, line 84: Can you add the label here so we know what null is? 🤔

app/src/main/java/com/toggl/domain/loading/LoadCalendarEventsSubscription.kt
```diff
@@ -0,0 +1,143 @@
...
+import kotlinx.coroutines.flow.flatMapLatest
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.onCompletion
+import kotlinx.coroutines.withContext
+import java.time.Duration
+import java.time.Instant
+import java.time.OffsetDateTime
+import java.time.ZoneOffset
+
+private class CalendarEventsContentObserver(
```
heytherewill, line 78: Shouldn't this be in another file? 🤔

app/src/main/java/com/toggl/domain/loading/LoadCalendarEventsSubscription.kt
```diff
@@ -0,0 +1,155 @@
...
+    private val dispatcherProvider: DispatcherProvider
+) {
+    private var calendarEventsContentObserver: CalendarEventsContentObserver? = null
+
+    suspend fun toCalendarEventsFlow(appState: AppState): Flow<List<CalendarEvent>> = withContext(dispatcherProvider.io) {
+        val calendarsFlow = MutableStateFlow<List<CalendarEvent>>(emptyList())
+        calendarEventsContentObserver = CalendarEventsContentObserver(
+            contentResolver,
+            appState.calendarLocalState.selectedDate.toBeginningOfTheDay(),
+            appState.calendarLocalState.selectedDate.toEndOfTheDay(),
```
heytherewill, line 60: I don't think that's how we should fetch calendar items. We should just fetch everything that is within the range we support (which currently is two weeks in the past but will soon be two month in the past and two months in the future) and put all that in the state. The UI can then figure out what it needs to display

app/src/main/java/com/toggl/domain/loading/LoadCalendarEventsSubscription.kt
```diff
@@ -0,0 +1,155 @@
...
+import java.time.OffsetDateTime
+import java.time.ZoneOffset
+
+class LoadCalendarsEventsSubscription(
+    private val calendarEventsProvider: CalendarEventsProvider
+) : Subscription<AppState, AppAction> {
+
+    override fun subscribe(state: Flow<AppState>): Flow<AppAction.Loading> =
+        state.map { s -> if (s.calendarPermissionWasGranted && s.user is Loadable.Loaded) s else null }
+            .distinctUntilChanged()
```
heytherewill, line 36: This distinctUntilChanged is wrong, it won't re-request the calendar events when the selected date changes. Nonetheless, I think we should change the way we fetch anyways

app/src/test/java/com/toggl/domain/loading/LoadCalendarEventsSubscriptionTests.kt
```diff
@@ -0,0 +1,165 @@
...
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flowOf
+import kotlinx.coroutines.flow.toList
+import kotlinx.coroutines.test.runBlockingTest
+import org.junit.jupiter.api.DisplayName
+import org.junit.jupiter.api.Test
+import java.time.Duration
+import java.time.OffsetDateTime
+
+@ExperimentalCoroutinesApi
```
heytherewill, line 23: This shouldn't be needed after Daivid's changes

app/src/main/java/com/toggl/domain/loading/LoadCalendarEventsSubscription.kt
heytherewill: The more I look at it the more I think it *shouldn't* be in another file, so maybe ignore this comment 😄

app/src/main/java/com/toggl/domain/loading/LoadCalendarEventsSubscription.kt
semanticer: You mean to use named argument to show `handler = null` that's unfortunatelly not supported for java objects.

app/src/main/java/com/toggl/domain/loading/LoadCalendarEventsSubscription.kt
semanticer: Yeah, I would keep it here, it is very closely coupled with the rest.

app/src/main/java/com/toggl/di/SubscriptionModule.kt
```diff
@@ -113,7 +113,16 @@ object SubscriptionModule {
...
     fun loadCalendarsEventsSubscription(
         @ApplicationContext context: Context,
+        timeService: TimeService,
         dispatcherProvider: DispatcherProvider
     ): Subscription<AppState, AppAction> =
-        LoadCalendarsEventsSubscription(CalendarEventsProvider(context.contentResolver, dispatcherProvider))
+        LoadCalendarsEventsSubscription(
+            timeService,
+            { minusMonths(2) },
+            { plusMonths(2) },
```
heytherewill, line 123: Extract these into constants

heytherewill: I think there already are some constants that are used by the UI to control exactly that; the idea is that the UI and this service should share the same set of constants related to dates available

semanticer: The only other Calendar related constant I could find is `datePickerRowItemsCount` which is something different

PR #735, opened 2020-08-13
Make the database user syncable

### LINKED ISSUE #714: Make DatabaseUser properties syncable
## Description
Refactor all DatabaseUser properties to `*SyncProperty` types with the exception of `Id`, `apiToken` and `serverId`

## Dependencies
#672  

## Definition of done
- [ ] Refactor all properties to corresponding `*SyncProperty` variants.
- [ ] Update all converter extension methods and make sure our domain models stay the same and are using "current" variant of `SyncProperty`

### LINKED ISSUE #716: Persist UserPreferences in DatabaseUser instead of shared preferences
## Description
All of `UserPreferences` should be persisted in DatabaseUser instead of shared preferences. All syncable properties should be changed to `*SyncProperty`

 
## Dependencies
#687 

## Definition of done
- [ ] Remove shared preferences code 
- [ ] Add user preferences properties to `DatabaseUser`
- [ ] Make `manualModeEnabled`, `twentyFourHourClockEnabled`, `groupSimilarTimeEntriesEnabled`, `selectedWorkspaceId`, `dateFormat`, `durationFormat` and `firstDayOfTheWeek` as `*SyncProperty` types
- [ ] Fix existing code
- [ ] Fix tests

heytherewill:
### Summary
This changes the way we persist our user settings and updates the entire user object to use the new sync property database scheme

### Technical considerations
I added different methods to update the database user based on what we need to update. Massive props to @semanticer for figuring out those `updateWith` functions; they read and work really well!

### Relationships
Closes #714
Closes #716

### Pull Request checklist

### Pointers
Review it commit by commit. Pay extra attention to whether I'm saving the right things at the right place

### Review template

## Merge permissions
Anyone, anytime


database/src/main/java/com/toggl/database/dao/UserDao.kt
```diff
@@ -5,12 +5,16 @@ import androidx.room.Insert
...
 import com.toggl.database.models.DatabaseUser
+import kotlinx.coroutines.flow.Flow
 
 @Dao
 interface UserDao {
     @Query("SELECT * FROM users")
     fun getAll(): List<DatabaseUser>
 
+    @Query("SELECT * FROM users")
+    fun getAllFlow(): Flow<List<DatabaseUser>>
```
semanticer, line 16: Maybe we should just LIMIT 1 and return a single user rather than the list of 1? It would be also nice to make sure the only way to insert a new user is to clear the previous one - both of these improvements could prevent higher levels of the app from shooting themselves in the foot.

heytherewill: I didn't know we could `LIMIT 1` :sogood:

heytherewill: About the limiting the way we insert, is there a way to run the insert without adding an insert method to the DAO? If there is, we could do it. Otherwise, that's impossible :/

PR #751, opened 2020-08-13
🛰️ Sync Push

### LINKED ISSUE #732: 🆙 Add Payload(Body) request object for POST /push/{id} method
## Description
Prepare the set of classes we'll need as **body** in  **POST /push/{id}**  method

## Dependencies
None

## Definition of done
- [ ] Prepare classes as defined in Docs and use existing API* classes as payloads if possible
- [ ] Add necessary Moshi annotations and custom adapters if needed
- [ ] Add serialization tests

### LINKED ISSUE #733: 🆙 Add response object for POST /push/{id} method
## Description
Add a set of classes for parsing response from **POST /push/{id}** method

## Dependencies
None

## Definition of done
- [ ] Prepare classes as defined in Docs for **POST /push/{id}** response
- [ ] Add necessary Moshi annotations and custom adapters if needed
- [ ] Add deserialization tests

### LINKED ISSUE #734: 🆙  Define sync POST /push/{id} method in our sync interface
## Description
Add **push** method to our sync retrofit interface and client

## Dependencies
#717 #732 #733 

## Definition of done
- [ ] Define push method in sync Retrofit interface
- [ ] Use request and response objects created in #732 and #733 
- [ ] Define push method in sync client
- [ ] Add tests if possible

heytherewill:
### Summary
This adds the sync push method and the things needed to serialize it

### Technical considerations
We still need to figure out a database view to fetch all data that needs syncing
For all other questions, ask @simonrozsival 😂 

This is different from the Xamarin implementation, but the idea is to make things simpler

### Relationships
Closes #732
Closes #733
Closes #734

### Pull Request checklist

### Pointers
Review commit by commit

### Review template

## Merge permissions
Anyone. anytime


api/src/main/java/com/toggl/api/network/models/sync/PushAction.kt
```diff
@@ -0,0 +1,28 @@
...
+    val type: ActionType,
+    val meta: MetaInfo?,
+    val payload: ActionResult<T>?
+)
+
+@JsonClass(generateAdapter = true)
+data class MetaInfo(
+    val id: Long?,
+    @Json(name = "client_assigned_id")
+    val clientAssignedId: Long?,
```
simonrozsival, line 17: This has to be a `String?` - the tests even contain the value as a string.

api/src/main/java/com/toggl/api/network/models/sync/PushResponse.kt
```diff
@@ -0,0 +1,25 @@
...
+import com.toggl.api.models.SyncApiTag
+import com.toggl.api.models.SyncApiTask
+import com.toggl.api.models.SyncApiTimeEntry
+import com.toggl.api.models.SyncApiUser
+import com.toggl.api.models.SyncApiWorkspace
+
+@JsonClass(generateAdapter = true)
+data class PushResponse(
+    @Json(name = "time_entries")
+    val timeEntries: List<PushAction<SyncApiTimeEntry>>,
```
simonrozsival, line 17: This should be lists of `ActionResult<T>`, not `PushAction<T>`.

api/src/main/java/com/toggl/api/network/models/sync/PushAction.kt
heytherewill: Why though?Our client will only ever use longs; could it be that we get client ids from other clients? 🤔

simonrozsival: How does it work then? It has to be serialized into a string and then when it's deserialized, it has to be parsed and converted from string into a long somewhere. Where is (or will) that implemented?

image

You can never get a `client_assigned_id` from any other client.

simonrozsival: I just noticed commit 37853df in which you fixed it.

heytherewill:
@daividssilverio It was addressed internally; the structure indeed is different

semanticer (approved):
It's looking good. And again, great job with that crazy adapter 👍

api/src/main/java/com/toggl/api/network/models/sync/PushResponse.kt
simonrozsival: As @heytherewill pointed out, we discussed this internally, and it should actually be `PushAction<T>` (the `ActionResult<T>` is in fact the `payload` of the action).

simonrozsival (approved):
The `ActionResultJsonAdapter` is scary, but it seems to work just fine. I think that there are test cases for all important scenarios.

PR #753, opened 2020-08-17
Add three way merge extensions

### LINKED ISSUE #737: Implement three way merge utility
## Description
We need to implement a three-way merge resolver. This could be done via some extension methods on our sync properties, static (object) methods or whatever Kolin gives us but the logic must be equivalent to [C# implementation](https://github.com/toggl/mobileapp/blob/a4db134ff16d28abbf662bba4cdfbd02d648f89c/Toggl.Storage.Realm/Sync/ThreeWayMerge.cs)

## Dependencies
None

## Definition of done
- [ ] Implement three way merge
- [ ] Add tests (unlike Sync guys who didn't test anything in Xamarin app 😁 )

heytherewill:
### Summary
This adds extensions for 3-way merging our props

### Technical considerations
I've added some extensions methods over the sync properties themselves. This way we can run it on top of the very database entities. Good stuff

Feel free to suggest some other way

### Relationships
Closes #737 

### Pull Request checklist

### Pointers
Just look at the code

### Review template

## Merge permissions
Anyone. anytime


repository/src/main/java/com/toggl/repository/sync/ThreeWayMergeExtensions.kt
```diff
@@ -0,0 +1,111 @@
...
+    val resolvedStatus = if (server == resolvedValue) PropertySyncStatus.InSync else PropertySyncStatus.SyncNeeded
+    return PropertyResolutionResult(resolvedStatus, resolvedValue)
+}
+
+data class PropertyResolutionResult<out T>(
+    val status: PropertySyncStatus,
+    val value: T
+)
+
+private fun <T> merge(original: T, local: T, server: T) = when {
```
simonrozsival, line 107: This is really nice!

repository/src/main/java/com/toggl/repository/sync/ThreeWayMergeExtensions.kt
```diff
@@ -0,0 +1,111 @@
...
+import com.toggl.database.properties.PropertySyncStatus
+import com.toggl.database.properties.StringSyncProperty
+import java.time.Duration
+import java.time.OffsetDateTime
+
+fun BooleanSyncProperty.resolve(server: Boolean): PropertyResolutionResult<Boolean> {
+    val original = if (status == PropertySyncStatus.InSync) current else backup
+    val resolvedValue = merge(original, current, server)
+    val resolvedStatus = if (server == resolvedValue) PropertySyncStatus.InSync else PropertySyncStatus.SyncNeeded
+    return PropertyResolutionResult(resolvedStatus, resolvedValue)
```
simonrozsival, line 23: Can't we move this block of code to some "generic" function `resolve<T>(PropertySyncStatus localState, T current, T backup, T server): PropertyResolutionResult<T>`? I understand that we need all the different non-generic `*SyncProperty` classes to store them in the DB in a straightforward way, but I believe there's a way to reduce the code duplication in this file, isn't there?

simonrozsival: Otherwise the logic seems to be identical to https://github.com/toggl/mobileapp/blob/develop/Toggl.Storage.Realm/Sync/ThreeWayMerge.cs#L52 :+1:

heytherewill: Good catch, Simon! I've done exactly that :)

PR #758, opened 2020-08-18
👨‍🎨 Add terms UI

### LINKED ISSUE #646: Add Terms and Conditions, Privacy Policy UI
## Description
During third party logins, we can detect there’s no account associated with the email address. We then switch to signup, but before actually signing the user up, we need to show them a screen where t&c + pp can be accepted. Note that we no longer present the country selection list and we have no designs for this screen, so please talk to product about how this should look. Maybe a simple popup would be sufficient, or just what we had before with country selection, but without the actual country list.

## Definition of done

- [ ] UI is added, where t&c + pp is linked and can be accepted

pdrj:
### Summary
This adds terms UI

### Technical considerations
This lives in `common` package for now. I'm not sure if it's going to be kept so 🤔 We will have to handle the 'Get started' tapped action differently when it comes from regular or third party signup. I'm not sure if that can be elegantly solved in a top-level onboarding reducer.

I'm open to suggestions.

### Relationships

Closes #646 

### Pull Request checklist

### Pointers
Navigate to the fragment and make sure it works and looks as intended. Use the links as well.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


common-feature/src/main/java/com/toggl/common/feature/navigation/Route.kt
```diff
@@ -11,6 +11,7 @@ sealed class Route {
     object Welcome : Route()
     object Login : Route()
     object SignUp : Route()
+    object Terms : Route()
```
heytherewill, line 14: Since #752 you also need to provide the animations for this route

common-android/src/main/res/values/strings.xml
```diff
@@ -138,4 +138,7 @@
     <string name="projects">Projects</string>
     <string name="tracked_hours">Tracked Hours</string>
     <string name="total">Total</string>
+    <string name="terms_and_conditions_header">Welcome to $\nGet tracking.</string>
+    <string name="terms_and_conditions_content">By tapping “%1$s”, you understand and agree to Toggl’s %2$s and %3$s.</string>
+    <string name="get_started">Get started</string>
```
heytherewill, line 143: Do we have 0 translations for these?

onboarding/src/main/java/com/toggl/onboarding/common/ui/TermsFragment.kt
```diff
@@ -0,0 +1,60 @@
...
+        headerBuilder.setSpan(ImageSpan(logoDrawable), newlineIndex - 1, newlineIndex, Spannable.SPAN_EXCLUSIVE_EXCLUSIVE)
+        terms_header.text = headerBuilder
+    }
+
+    private fun prepareBodyText() {
+        val termsAndConditions = requireContext().getString(R.string.terms_of_service)
+        val privacyPolicy = requireContext().getString(R.string.privacy_policy)
+        val getStarted = requireContext().getString(R.string.get_started)
+        val termsText =
+            requireContext().getString(R.string.terms_and_conditions_content, getStarted, termsAndConditions, privacyPolicy)
```
heytherewill, line 51: You could require context just once at the beginning and reuse the result of the function call

heytherewill:
Do merge master in this branch so I can see what it looks like with the routing changes

common-android/src/main/res/values/strings.xml
pdrj: I was able to adapt the `terms_and_conditions_content` from xamarin. The rest we don't have.

PR #713, opened 2020-08-06
Add text editing settings

heytherewill:
### Summary
This adds the settings that need a text editor

### Technical considerations
I need @semanticer to chime in, since apparently EditTexts are broken on Compose 😂 

### Why is it done like this and not some other way?
We still need final designs, but we can test the implementation with this partial layout and figure apply the final layout later :)

### Relationships
Closes #528
Closes #529 
Closes #530 

### Pull Request checklist
- [X] The code does not hinder the app's performance
- [X] The code does not introduce regressions
- [ ] The code in this PR is tested thoroughly enough that changing the behavior to an incorrect one breaks tests
- [ ] One of our [live or file custom templates](https://github.com/toggl/android/tree/master/.idea/templates) was used in this PR

### Pointers
Review it commit by commit
Don't functional test until this PR is no longer a draft

### Review template

- [ ] The code is thoroughly tested in well written and resilient manner
- [ ] The code is performant and leads to an optimal user experience
- [ ] The code is testable and does not cut corners for the sake of delivering
- [ ] The code considers older platforms well and will work in older devices just like it will in new ones

- [ ] The code considers light and dark theme
- [ ] The code considers RTL and does not mention left/right
- [ ] The code considers i18n
  - [ ] There are no hardcoded strings in the UI
  - [ ] All strings are included in every language we support OR there's an issue requesting new translations
- [ ] There are no hard-coded numbers referring to dimensions or font sizes
- [ ] The code considers UI elements that are similar and extracts their similarities into styles to avoid duplication

## Merge permissions
Only myself

PR #763, opened 2020-08-19
📦Update compose to dev17

heytherewill:
### Summary
Another day, another compose update

### Technical considerations
This one is kind of a big deal because it fixes the crash in #713 and finally unblocks that PR :sogood:

### Pull Request checklist

### Pointers
All changes are needed to fix warnings, so there's not much to see

### Review template

## Merge permissions
Anyone. anytime


semanticer (changes requested):
Let's also update kotlin to the same version https://github.com/toggl/android/commit/5febc52c2142a2d10993fd06af7ea38fe36f7496 

Also, toggles are now looking quite funky
ezgif com-video-to-gif

semanticer (approved):
👍

PR #784, opened 2020-08-21
✨ Process pulled workspaces

### LINKED ISSUE #748: Implement WorkspaceFeatures response processing
## Description
Process Workspace features response after a successful pull request and update given workspaces in the DB. 
Please note that the `workspace_id` you'll get from pull response will contain `serverId` of the given Workspace, not the local one so make sure you update the correct workspace.

## Dependencies
#737  #741

## Definition of done
- [ ] Add Tests

### LINKED ISSUE #749: Implement Workspace response processing
## Description
Process Client response after a successful pull request. Use 3WM extension methods for every sync-able property.

Xamarin implementation is using one method for processing **tags**, **clients**, **project**, **tasks** and **workspaces** it's called [processEntities](https://github.com/toggl/mobileapp/blob/8779617e49c7da8c9899d36061784dcbb13e4ebf/Toggl.Storage.Realm/Queries/WritePulledData.cs#L220). We probably won't be able to do the same but still have a look for a reference.

## Dependencies
#737  #741

## Definition of done
- [ ] Add Tests

heytherewill:
### Summary
This adds the code that actually persists the pulled workspaces

### Technical considerations
I noticed that the `projectsBillableByDefault` flag is not being used at all in our codebase; I added a note to fix this before we release. I also added some properties to the database workspace that were needed

I've made the methods in the WorkspaceDao `suspend fun`s because [of these recommendations](https://developer.android.com/reference/kotlin/androidx/room/package-summary#(androidx.room.RoomDatabase).withTransaction(kotlin.coroutines.SuspendFunction0))

I had to bundle two methods because we are persisting all the workspace features inside the workspace database, unlike what we did in the Xamarin app where there were two separate tables

### Relationships
Closes #748
Closes #749

### Pull Request checklist

### Pointers
Commit by commit is the way

### Review template

## Merge permissions
Anyone. anytime


api/src/main/java/com/toggl/api/clients/ErrorHandlingProxyClient.kt
```diff
@@ -118,7 +118,7 @@ internal class ErrorHandlingProxyClient @Inject constructor(
 
     override suspend fun pull(since: OffsetDateTime?): PullResponse {
         try {
-            return syncApi.pull(since?.toEpochSecond())
+            return syncApi.pull(since?.toInstant()?.toEpochMilli())
```
simonrozsival, line 121: Why milliseconds? The `/pull` endpoint expects seconds.

common-android/src/main/java/com/toggl/common/services/time/LastSinceDateStorage.kt
```diff
@@ -0,0 +1,29 @@
...
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class LastSinceDateStorage @Inject constructor(
+    private val sharedPreferences: SharedPreferences
+) {
+    private val lastSinceSharedPreferencesKey = "last_since"
+
+    fun getLastSince(): OffsetDateTime? {
```
simonrozsival, line 17: In this case, I honestly don't think you should convert it to the `OffsetDateTime` type. It would make more sense to have some type alias (if that's possible in Kotlin) for `Long` and simply store that number in the DB and then send that same number to the server. The `since` parameter was supposed to be a token that you don't need to interpret. The way we implemented it in the C# code was mostly to be backwards compatible to the old sync (which used `DateTimeOffset` based API in the repository). On the other hand, if you treat that timestamp you get from the server as milliseconds and convert it to the `OffsetDateTime` and then later you convert it back to _milliseconds_, you're going to get the same number of seconds in the end 🤷 😄

database/src/main/java/com/toggl/database/models/DatabaseWorkspace.kt
```diff
@@ -2,16 +2,24 @@ package com.toggl.database.models
...
     @PrimaryKey
     val id: Long = 0,
     val serverId: Long?,
     @Embedded(prefix = "name_")
     val name: StringSyncProperty,
-    val features: List<WorkspaceFeature>
+    val features: List<WorkspaceFeature>,
+    val projectsBillableByDefault: Boolean,
+    val isDeleted: Boolean,
+    val syncStatus: EntitySyncStatus
```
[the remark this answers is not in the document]
simonrozsival: Doing it explicitly is better IMO. Also the `EntitySyncStatus` also contains the failure state which property states can't have.

database/src/main/java/com/toggl/database/properties/EntitySyncStatus.kt
```diff
@@ -0,0 +1,7 @@
+package com.toggl.database.properties
+
+enum class EntitySyncStatus {
```
simonrozsival, line 3: `SyncFailed` should go here as well.

database/src/main/java/com/toggl/database/models/DatabaseWorkspace.kt
simonrozsival, line 24: In Realm entities we also had a `string LastSyncErrorMessage` property which would be `null` if there wasn't any error and the text of the error message returned from the API if sync of that entity failed. It was then displayed in the edit view:
image

I'm not saying we should implement it the same way, but there should be a way of storing the sync error for each creatable/editable entity.

repository/src/test/java/com/toggl/repository/RepositoryTest.kt
```diff
@@ -378,4 +381,28 @@ class RepositoryTest : CoroutineTest() {
             editor.putString(ApiTokenProvider.apiToken, token.toString())
         }
     }
+
+    @Test
+    fun `the clear all method cleans the sync api token`() = runBlockingTest {
```
simonrozsival, line 386: The test name doesn't seem related to what the test does, is that a copy&pasta issue or am I missing something? Shouldn't it be `the clear all method cleans the last since timestamp` or something like that?

sync/src/main/java/com/toggl/sync/processing/ProcessWorkspaces.kt
```diff
@@ -0,0 +1,45 @@
...
+
+import com.toggl.api.models.ApiWorkspace
+import com.toggl.api.models.ApiWorkspaceFeature
+import com.toggl.database.dao.WorkspaceDao
+import com.toggl.database.models.DatabaseWorkspace
+import com.toggl.database.properties.EntitySyncStatus
+import com.toggl.database.properties.StringSyncProperty
+import com.toggl.models.domain.WorkspaceFeature
+
+internal suspend fun processWorkspaces(
```
simonrozsival, line 11: If this is a "free floating function" here, I think the name should mention that it's supposed to be used for processing the workspace after the data was _pulled_ from the server. In Xamarin, I believe this method was under a query specifically intended for saving _pull_ response.
```suggestion
internal suspend fun processPulledWorkspaces(
```

simonrozsival (changes requested):

sync/src/main/java/com/toggl/sync/processing/ProcessWorkspaces.kt
```diff
@@ -0,0 +1,47 @@
...
+        val apiWorkspaceFeature = apiWorkspaceFeatures.firstOrNull { it.workspaceId == apiWorkspace.id }
+
+        if (localWorkspace == null) {
+            if (isServerDeleted)
+                continue
+
+            val newWorkspace = apiWorkspace.toDatabaseWorkspace(apiWorkspaceFeature)
+            workspaceDao.insert(newWorkspace)
+            continue
+        }
```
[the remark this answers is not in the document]
heytherewill: For papertrail reasons: This is fixed by the outstanding push step

api/src/main/java/com/toggl/api/clients/ErrorHandlingProxyClient.kt
heytherewill: Oh, then the bug wasn't here, it was in the other point of the app 😄

simonrozsival (approved):
Yep, LGTM 👍

Issue #654, opened 2020-07-24
Add GetSsoUrlEffect


pdrj:
## Description
Implement the effect, it should talk to the sso url endpoint and emit AuthenticateWithSso or SetError. The endpoint and message structure should be the same as in Xamarin app.

## Dependencies
#653 
## Definition of done
Effect is added
API communication happens
Errors are handled
Tests are added

- [ ] The documentation is updated to reflect the changes done in this PR

PR #786, opened 2020-08-24
✨ Add SSO auth effect and deep link reducer

### LINKED ISSUE #656: Add PerformSsoAuthFlowEffect
## Description
Navigate user to the URL returned by GetSsoUrlEffect. Use https://developer.apple.com/documentation/authenticationservices/aswebauthenticationsession or https://developers.google.com/web/android/custom-tabs/implementation-guide Then either emit SetError / SetUser (if the flow returned an api token) / MustLinkSsoAccounts(email, confirmationCode) - if the account needs to be sso linked. More info on params and how the flow can go are in https://www.notion.so/toggl/SSO-d8929cc996084f3db6fdd08637080f1c and in the Xamarin app

## Dependencies
#655 
## Definition of done
Effect is added
It triggers the auth flow 
It handles the flow so that errors are handled and params extracted from the togglauth:// deep link
It emits the correct actions
Tests are added

pdrj:
### Summary
Adds SSO auth effect, deep link reducer, auth with api token and other missing stuff

### Technical considerations
This probably won't compile as is, needs Daivid's get sso url effect to then use the context supplied to the reducer. I have not added tests yet - I wanted to get someone's opinion on the approach first to not redo the work.

This seemed like the way to go, but I'm open to suggestions.

### Relationships

Closes #656

### Pull Request checklist

### Pointers

Try to login via sso on web staging. Copy the okta url. Use that url to trigger the effect in SSO reducer. That's how I do it:
```
            SsoAction.ContinueButtonTapped -> when (val email = state().email) {
                is Email.Valid -> listOf(
                    GetSsoUrlEffect(
                        email,
                        apiClient,
                        userRepository,
                        getSsoUrlErrorMessages,
                        dispatcherProvider
                    )
                )
                else -> effect(
                    PerformSsoAuthFlowEffect(
                        Uri.parse("<the url copied from web app>"),
                        context
                    )
                )
            }
```

Then go through the okta flow (make sure you are added to the sso test workspace in staging and to okta itself). If the okta auth passes, you should be logged in. If your account needs linking, you have to link it in the xamarin app or elsewhere. It's not implemented here yet.

Try an error-triggering flow as well.

### Review template

- [ ] The code considers i18n

## Merge permissions
Me


app/src/main/AndroidManifest.xml
```diff
@@ -16,12 +16,21 @@
         <meta-data
             android:name="firebase_crashlytics_collection_enabled"
             android:value="${crashlytics}" />
-        <activity android:name=".ui.MainActivity">
+        <activity
+            android:name=".ui.MainActivity"
+            android:launchMode="singleTask">
```
daividssilverio, line 21: It always makes me go see the docs when I see `launchMode` different than the standard.
This should be alright.

onboarding/src/main/java/com/toggl/onboarding/sso/domain/PerformSsoAuthFlowEffect.kt
```diff
@@ -0,0 +1,20 @@
...
+import android.content.Intent
+import android.net.Uri
+import androidx.browser.customtabs.CustomTabsIntent
+import com.toggl.architecture.core.Effect
+
+class PerformSsoAuthFlowEffect(
+    private val idpUrl: Uri,
+    private val context: Context
+) : Effect<SsoAction> {
+    override suspend fun execute(): SsoAction? {
```
daividssilverio, line 13: Is this even regularly testable if we add references to the android framework?
I was under the impression that we our reducers and effects weren't supposed to have direct references to the android framework.

onboarding/src/main/java/com/toggl/onboarding/sso/domain/SsoReducer.kt
```diff
@@ -1,35 +1,98 @@
...
 import com.toggl.models.validation.Email
 import com.toggl.models.validation.toEmail
 import com.toggl.repository.interfaces.UserRepository
+import dagger.hilt.android.qualifiers.ApplicationContext
 import javax.inject.Inject
 import javax.inject.Singleton
 
 @Singleton
 class SsoReducer @Inject constructor(
+    @ApplicationContext private val context: Context,
```
daividssilverio, line 28: Yeah... adding the context to the reducer complicates testing, doesn't it?

daividssilverio:
After plugging in the changes from #654 I've managed to at least test one unhappy path. The deeplink seems to work at least when it fails. I've managed to get the "Single Sign On is not configured..." error. Same thing happens if I try to login on the xamarin app or the web app. I did accept the invite to the `Toggl SSO Test` workspace.

One problem, after this failure, the reducer enters in a infinite loop of `SsoAction.HandleDeepLink`s.

onboarding/src/main/java/com/toggl/onboarding/sso/ui/SsoFragment.kt
```diff
@@ -31,11 +38,28 @@ class SsoFragment : Fragment(R.layout.fragment_sso) {
         }
 
         emailWatcher = email.doSafeAfterTextChanged { store.dispatch(SsoAction.EmailEntered(it.toString())) }
+
+        store.state
+            .map { it.deepLinkData }
+            .filter { it != Uri.EMPTY }
```
daividssilverio, line 44: I'm guessing this is causing the infinite loop. We should probably check if there's an error in the state before trying to handle the deep link, or something like that.

onboarding/src/main/java/com/toggl/onboarding/sso/domain/SsoReducer.kt
```diff
@@ -1,35 +1,98 @@
...
+                )
                 else -> noEffect()
             }
             is SsoAction.EmailEntered -> state.mutateWithoutEffects { copy(email = action.email.toEmail()) }
+            is SsoAction.HandleDeepLink -> {
+                val params =
+                    action.deepLinkData.queryParameterNames.associateWith { action.deepLinkData.getQueryParameter(it)!! }
+                val keys = params.keys
+
+                if (keys.containsAll(loginWithApiTokenKeys)) {
```
daividssilverio, line 63: it would be nicer if you could create some functions to hide the complexity of these `keys.containsAll` checks.
Something like: `ssoLoginSucceeded(keys)`, `needsAccountLinking(keys)`, or something like that.

Plus, extracting some functions would help reducing the 波動拳(Hadouken) effect here.

app/src/main/AndroidManifest.xml
pdrj: Yep I changed it because I don't think we want to have more than one instance at a time.

onboarding/src/main/java/com/toggl/onboarding/sso/domain/SsoReducer.kt
pdrj: I haven't tested it, but we should be able to just use a mock, right? That being said, I also felt odd about having the context injected into the reducer. I'll change it so that there's a flag in state that tells the fragment to launch the browser.

onboarding/src/main/java/com/toggl/onboarding/sso/ui/SsoFragment.kt
pdrj: The link is not cleared in the unhappy path. I'll fix it, thanks 👍

pdrj:
> Requesting changes for the missing MustLinkSsoAccounts that should be emitted based on the description of the issue this PR is closing.

When creating the issue I thought we'll be able to handle the deep link in effect and emit actions from there. It is not the case, so I have not added the action, just modified the state accordingly in reducer.

onboarding/src/main/java/com/toggl/onboarding/sso/domain/SsoReducer.kt
```diff
@@ -18,22 +28,87 @@ import javax.inject.Singleton
...
-    private val getSsoUrlEffectErrorMessages: GetSsoUrlEffect.ErrorMessages,
+    private val getSsoUrlErrorMessages: GetSsoUrlEffect.ErrorMessages,
+    private val logUserInWithAuthTokenErrorMessages: LogUserInWithAuthTokenEffect.ErrorMessages,
     private val linkSsoEffectOutcomeMessages: LinkSsoEffect.OutcomeMessages,
     private val dispatcherProvider: DispatcherProvider
 ) : Reducer<SsoState, SsoAction> {
 
+    private val mustLinkKeys = listOf("confirmation_code", "email")
+    private val loginWithApiTokenKeys = listOf("apiToken")
+    private val errorKeys = listOf("ssoError")
```
daividssilverio, line 39: Hey @pdrj, I know it's merged, but do we need this?

pdrj: We could delete the last one. I just left it there in case we actually want to get the value at some point.

PR #788, opened 2020-08-25
🐛 Fix sign in and sign out

heytherewill:
### Summary
This fixes a tiny bug in our sign in/out flow

### Technical considerations
We were persisting the user with the defaultWorkspaceId set as the serverId rather than the localId. I fixed that. I also did not add a test because this code will be 🔥d in the future when sync is completed

When deleting our entities, we need to it in the opposite order of the sync algorithm so that deleting does not create FK violations. I did that and created a test to ensure such behavior

### Pull Request checklist

### Pointers
Commit by commit

### Review template

## Merge permissions
Anyone, anytime


simonrozsival (approved):
LGTM

PR #789, opened 2020-08-25
🔨 Add, Reduce and Bind LinkSsoButtonTapped and CloseLinkSsoTapped actions

### LINKED ISSUE #658: Add LinkSsoButtonTapped action
## Description
Add the action. Reduce it so that user is taken to the custom welcome screen prepared in #648. The email and confirmation code params have to be preserved somewhere so they can be used in effect later
## Dependencies
#648 
#647 
## Definition of done
UI is wired up

### LINKED ISSUE #659: Add CloseLinkSsoTapped action
## Description
Add the action, reduce it so that user is taken back to the main welcome screen.

## Dependencies
#647 

## Definition of done
UI is wired up

daividssilverio:
### Summary
This adds, reduces and binds the `LinkSsoButtonTapped` action.

### Technical considerations
The flow to test this action isn't ready yet.
The user needs to go through the login with sso flow to be navigated to the `SsoLinkFragment`.

I think it makes sense to store the e-mail and confirmation code directly in the state, then the `LinkSsoButtonTapped` doesn't need to take in any parameters and can be simply fired when the "log in to enable sso" button is tapped in the `SsoLinkFragment`.

Feel free to make your suggestions.

### Relationships
Closes #658
Closes #659

### Pull Request checklist

### Pointers
eh... hard to test without the rest of the flow. But take a look at the tests and the notion page: https://www.notion.so/toggl/SSO-d8929cc996084f3db6fdd08637080f1c.

From my understanding, `MustLinkSsoAccounts` will trigger the navigation to the `SsoLinkFragment` and set the confirmation code to the state.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it


pdrj:
>From my understanding, MustLinkSsoAccounts will trigger the navigation to the SsoLinkFragment and set the confirmation code to the state.

Yeah, that is correct. Mostly, because we don't have the MustLinkSsoAccounts action, as described in my sso effect PR.

onboarding/src/main/java/com/toggl/onboarding/common/domain/OnboardingState.kt
```diff
@@ -18,13 +18,15 @@ data class OnboardingState(
         internal val password: Password,
         internal val resetPasswordResult: Loadable<String>,
         internal val welcomeScreenConfiguration: WelcomeScreenConfiguration,
+        internal val confirmationCode: String?,
```
pdrj, line 21: I actually approached it in a similar way in my sso effect PR. I just have a wrapper over confirmation code and email (we need both of these for linking).

onboarding/src/main/res/layout/fragment_ssolink.xml
```diff
@@ -58,15 +58,24 @@
...
+            android:layout_marginBottom="@dimen/grid_1"
             android:text="@string/login_to_enable_sso"
             android:textColor="@color/authentication_button_text"
             app:backgroundTint="@color/primary_dark"
-            app:layout_constraintBottom_toBottomOf="parent"
+            app:layout_constraintBottom_toTopOf="@+id/sso_link_cancel"
             app:layout_constraintEnd_toEndOf="parent"
             app:layout_constraintStart_toStartOf="parent" />
 
+        <Button
```
pdrj, line 69: We could reuse the @style/TextOnboardingButton style and the other button style for the button above

pdrj (changes requested):
Just minor suggestions.

onboarding/src/main/java/com/toggl/onboarding/common/domain/OnboardingState.kt
pdrj: It seems like we don't need email actually (during the SSO project I vaguely recall it was decided that backend doesn't want that parameter). So your approach without wrapper is OK.

onboarding/src/main/java/com/toggl/onboarding/common/domain/OnboardingState.kt
```diff
@@ -18,13 +18,15 @@ data class OnboardingState(
...
         internal val ssoError: Failure?
     ) {
         constructor() : this(
             Email.Invalid(""),
             Password.from(""),
             Loadable.Uninitialized,
-            WelcomeScreenConfiguration.Regular,
+            WelcomeScreenConfiguration.SsoLink,
```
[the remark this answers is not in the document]
daividssilverio: Oops, yeah.

onboarding/src/main/res/layout/fragment_ssolink.xml
daividssilverio: Alright. I think we'll have to come back to this later though.

pdrj (approved):
Please apply the small changes before merging. Approving for 🐎

pdrj:
Hm, the comment about 'small changes' didn't register. Basically, please use the `OnboardingButton` style for the link button (it uses the`Text` one now), the `TextOnboardingButton` for the Cancel button. Then, don't override the width/height and margins if not necessary - this should keep the buttons' dimensions in line with other screens.

daividssilverio:
@pdrj the cancel button was already like the other usages of "cancel buttons" on onboarding.
I've updated the action button as you requested.

PR #790, opened 2020-08-26
🔨 Add link sso action and effect

### LINKED ISSUE #660: Add LinkSsoAction
## Description
Add the action, this is a proxy that calls LinkSsoEffect

## Definition of done

### LINKED ISSUE #661: Add LinkSsoEffect
## Description
This effect should perform the account linking. See Xamarin app for endpoint spec. Some info is here https://github.com/toggl/mobileapp/pull/7519/files and https://www.notion.so/toggl/SSO-d8929cc996084f3db6fdd08637080f1c

## Definition of done
Effect is added
API is integrated
Results are correctly propagated

daividssilverio:
### Summary
This adds the `LinkSso` action and the `LinkSsoEffect` with some alterations from what was described on https://www.notion.so/toggl/SSO-d8929cc996084f3db6fdd08637080f1c.
Another action was added as a way to store the result of the `LinkSsoEffect` so that we can show some feedback to the user about the result of the linking.

### Technical considerations
The `LinkSsoEffect` is assuming the user has already been set in the repository. That ensures that the user's api token has already been set in the local storage and the call to the link sso api is authenticated.

Feel free to make your suggestions. This made sense after all the discussions with @pdrj .
The idea is that the `LinkSso` is emitted sometime after a successful login (regular or 3rd party) that happens after a `MustLinkSsoAccounts` has been emitted.

The `SetSsoLinkResult` is emitted and `ssoLinkResult: Either<Unit, Failure>` and stored in the state. From there we can fire the toast from somewhere in the flow (maybe directly from one of the onboarding screens, or we can move the `ssoLinkResult` up the state and fire the toast in the timer.

### Relationships
Closes #660 
Closes #661

### Pull Request checklist

### Pointers
Also a bit hard to test without the rest of the sso authentication flow.
But you can use the xamarin app to go half way through the flow and manually emit the `LinkSso` action to see the magic.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it


pdrj (commented):
The code looks good, but I've been unable to test this :( 

I've reached out to backend to see if they can help https://toggl.slack.com/archives/C010JBG1KTK/p1598443888002600

pdrj (approved):
That's the deal: SSO for our toggl emails can only work on staging XOR prod. It is currently configured to work on prod, so the invites I sent you on staging couldn't have worked.

In order to test this, we need to be hitting prod endpoints (can be from a debug app) and the tested email has to be a part of prod sso test workspace. I did that and was able to test the 'invalid code', 'bad email' and happy scenario.

Note that you don't have to create a new account to test the linking flow. It is enough to leave the test sso workspace and then rejoin it. This will re-trigger the linking request.

PR #787, opened 2020-08-25
Process Pulled User

heytherewill:
### Summary
This adds the method to process the user that was pulled from the api

### Technical considerations
I've bundled the processing of the user and preferences because they are represented by the same entity. Also, the line is really grey there.

The first commit has most changes. I needed to make the `defaultWorkspace` property optional. That's by design; the defaultWorkspace property is nullable in the API, we simply didn't make it optional in our models because it was simpler not to.

### Why is it done like this and not some other way?
There are some properties that we are fetching but not using, but that's by design.

### Relationships
Closes #742
Closes #743

### Pull Request checklist
- [X] The code does not hinder the app's performance
- [X] The code does not introduce regressions
- [X] The code in this PR is tested thoroughly enough that changing the behavior to an incorrect one breaks tests

### Pointers
Review commit by commit, since most of the files touched are needed to change the defaultWorkspaceId property to Nullable

### Review template

- [ ] The code is thoroughly tested in well written and resilient manner
- [ ] The code is performant and leads to an optimal user experience
- [ ] The code is testable and does not cut corners for the sake of delivering
- [ ] The code considers older platforms well and will work in older devices just like it will in new ones

- [ ] The code considers light and dark theme
- [ ] The code considers RTL and does not mention left/right
- [ ] The code considers i18n
  - [ ] There are no hardcoded strings in the UI
  - [ ] All strings are included in every language we support OR there's an issue requesting new translations
- [ ] There are no hard-coded numbers referring to dimensions or font sizes
- [ ] The code considers UI elements that are similar and extracts their similarities into styles to avoid duplication

## Merge permissions
Anyone, anytime

PR #791, opened 2020-08-26
🎇 Process pulled tags and clients

### LINKED ISSUE #744: Implement Tags response processing
## Description
Process Tags response after a successful pull request. Use 3WM extension methods for every sync-able property.
Please note that the `workspace_id` you'll get from pull response will contain `serverId` of the given Workspace not the local one so you will need one extra query to get the workspace's `localId` based on the `serverId` you already have and save the retrieved `localId` with the tag. Here is a pseudocode explaining this operation:

```kotlin
val defaultWorkspaceServerId = tagFromPullResponse.default_workspace_id
val defaultWorkspaceLocalId = workspaceDao.getLocalIdByServerId(defaultWorkspaceServerId)
val localtagWithCorrectRelationToWorkspace = tagFromPullResponse.toLocalTag(defaultWorkspaceLocalId)
```

Xamarin implementation is using one method for processing **tags**, **clients**, **project**, **tasks** and **workspaces** it's called [processEntities](https://github.com/toggl/mobileapp/blob/8779617e49c7da8c9899d36061784dcbb13e4ebf/Toggl.Storage.Realm/Queries/WritePulledData.cs#L220). We probably won't be able to do the same but still have a look for a reference.

## Dependencies
#737  #741

## Definition of done
- [ ] Add Tests

### LINKED ISSUE #747: Implement Client response processing
## Description
Process Client response after a successful pull request. Use 3WM extension methods for every sync-able property.
Please note that the `workspace_id` you'll get from pull response will contain `serverId` of the given Workspace not the local one so you will need one extra query to get the workspace's `localId` based on the `serverId` you already have and save the retrieved `localId` with the client. Here is a pseudocode explaining this operation:

```kotlin
val defaultWorkspaceServerId = projectFromPullResponse.default_workspace_id
val defaultWorkspaceLocalId = workspaceDao.getLocalIdByServerId(defaultWorkspaceServerId)
val localClientWithCorrectRelationToWorkspace = clientFromPullResponse.toLocalProject(defaultWorkspaceLocalId)
```

Xamarin implementation is using one method for processing **tags**, **clients**, **project**, **tasks** and **workspaces** it's called [processEntities](https://github.com/toggl/mobileapp/blob/8779617e49c7da8c9899d36061784dcbb13e4ebf/Toggl.Storage.Realm/Queries/WritePulledData.cs#L220). We probably won't be able to do the same but still have a look for a reference.

## Dependencies
#737  #741

## Definition of done
- [ ] Add Tests

heytherewill:
### Summary
This adds processing of tags and clients

### Technical considerations
Since these are the "simple" entities (as in, they can be created but not edited) I've added them in bundle.

The code is very similar, but I'd rather not have generic methods because it'd make things much more complex (since there's no common interface for entities like in C#)

### Relationships
Please merge after #787 
Closes #744 
Closes #747 

### Pull Request checklist

### Pointers
This is so straightforward that you can just look at the code 

### Review template

## Merge permissions
Anyone, anytime


sync/src/main/java/com/toggl/sync/processing/ProcessPulledClients.kt
```diff
@@ -0,0 +1,41 @@
...
+    clientDao: ClientDao,
+    apiClients: List<ApiClient>
+) {
+    for (apiClient in apiClients) {
+        val localClient = syncDao.getClientByServerId(apiClient.id)
+        val isServerDeleted = apiClient.serverDeletedAt != null
+
+        if (localClient == null) {
+            if (!isServerDeleted) {
+                val localWorkspaceId = syncDao.getWorkspaceIdByServerId(apiClient.workspaceId) ?: continue
```
simonrozsival, line 21: The `continue` is kinda scary 🤔 If it ever occurs that we should save a client from a WS that we don't have locally, then there was a major problem that we didn't handle:
- either the server returned other other user's data (the worst case of them all)
- or the server didn't return the workspace itself
- or we deleted the workspace while we shouldn't
- or we messed up the IDs (local vs. server)
- or there's some other problem I can't think of right now.

We should at least log and report that this occurred.

sync/src/main/java/com/toggl/sync/processing/ProcessPulledClients.kt
```diff
@@ -0,0 +1,41 @@
...
+                val newWorkspace = apiClient.toDatabaseClient(localWorkspaceId)
+                clientDao.insert(newWorkspace)
+            }
+        } else {
+            if (isServerDeleted) {
+                clientDao.delete(localClient)
+            } else {
+                val localWorkspaceId = syncDao.getWorkspaceIdByServerId(apiClient.workspaceId) ?: continue
+                val updatedWorkspace = apiClient.toDatabaseClient(localWorkspaceId).copy(id = localClient.id)
+                clientDao.update(updatedWorkspace)
```
simonrozsival, line 31:
```suggestion
                val updatedClient = apiClient.toDatabaseClient(localWorkspaceId).copy(id = localClient.id)
                clientDao.update(updatedClient)
```

sync/src/main/java/com/toggl/sync/processing/ProcessPulledTags.kt
```diff
@@ -0,0 +1,41 @@
...
+                val newWorkspace = apiTag.toDatabaseTag(localWorkspaceId)
+                tagDao.insert(newWorkspace)
+            }
+        } else {
+            if (isServerDeleted) {
+                tagDao.delete(localTag)
+            } else {
+                val localWorkspaceId = syncDao.getWorkspaceIdByServerId(apiTag.workspaceId) ?: continue
+                val updatedWorkspace = apiTag.toDatabaseTag(localWorkspaceId).copy(id = localTag.id)
+                tagDao.update(updatedWorkspace)
```
simonrozsival, line 31:
```suggestion
                val updatedTag = apiTag.toDatabaseTag(localWorkspaceId).copy(id = localTag.id)
                tagDao.update(updatedTag)
```

sync/src/main/java/com/toggl/sync/processing/ProcessPulledTags.kt
```diff
@@ -0,0 +1,41 @@
...
+internal suspend fun processPulledTags(
+    syncDao: SyncDao,
+    tagDao: TagDao,
+    apiTags: List<ApiTag>
+) {
+    for (apiTag in apiTags) {
+        val localTag = syncDao.getTagByServerId(apiTag.id)
+        val isServerDeleted = apiTag.serverDeletedAt != null
+
+        if (localTag == null) {
```
simonrozsival, line 19: Now that I'm looking at this, we might abstract this flow:

```
if (!existsLocally) {
  if (!isServerDeleted) { insert() }
} else {
  if (isServerDeleted) { delete() }
  else { update() }
}
```

... and pass just those three functions/object with those three functions into the generic "decision tree"? It would make the code much simpler and easier to understand.

simonrozsival (changes requested):

sync/src/main/java/com/toggl/sync/processing/ProcessPulledTags.kt
heytherewill: I thought about that as well. The problem is that we don't have (and we don't wanna have) a common interface, so the code to actually enable this generic will be really bloated (since different entities need . It's not that much repetition anyways and we'll only touch this code when we add new entities that can be pushed, so we're fine repeating this logic for each of the entities and keeping readability high

sync/src/main/java/com/toggl/sync/processing/ProcessPulledClients.kt
heytherewill: I've created an issue so we can handle this once all entities are being properly pulled: https://github.com/toggl/android/issues/797

simonrozsival (approved):
LGTM now 👍

PR #793, opened 2020-08-27
🆙 Update AS gradle plugin & gradle to 4.2-alpha08 and 6.6-rc-6

daividssilverio:
### Summary
With compose alpha release, they've fixed our marvellous previews!!! YES!!!

### Technical considerations
None. The previews are working again.

This is the way.

### Relationships
Closes #Nothing

### Pull Request checklist

### Pointers
Upgrade your AS. You can download it here https://developer.android.com/studio/preview . For some reason, the update wasn't detected on my Jetbrains' toolbox.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it


pdrj (approved):
A clean build worked fine, settings rendered fine, as did reports. I have no previews yet, but the latest canary is still downloading.

Issue #750, opened 2020-08-13
Implement TimeEntries response processing


semanticer:
## Description
Process Time Entry response after a successful pull request. Use 3WM extension methods for every sync-able property.

[Xamarin implementation is here](https://github.com/toggl/mobileapp/blob/8779617e49c7da8c9899d36061784dcbb13e4ebf/Toggl.Storage.Realm/Queries/WritePulledData.cs#L37)

Please note that the `workspace_id` you'll get from pull response will contain `serverId` of the given Workspace not the local one so you will need one extra query to get the workspace's `localId` based on the `serverId` you already have and save the retrieved `localId` with the client. Here is a pseudocode explaining this operation:

```kotlin
val defaultWorkspaceServerId = timeEntryFromPullResponse.default_workspace_id
val defaultWorkspaceLocalId = workspaceDao.getLocalIdByServerId(defaultWorkspaceServerId)
val localTimeEntryWithCorrectRelationToWorkspace = timeEntryFromPullResponse.toLocalProject(defaultWorkspaceLocalId)
```

Same goes for `project_id`, `task_id` and `tag_ids`

## Dependencies
#737  #741

## Definition of done
- [ ] Process sync-able properties using 3WM extension methods
- [ ] Add Tests

PR #799, opened 2020-09-01
Simplify tag persisting mechanism

heytherewill:
### Summary
Our tag persisting mechanism was ridiculously complex. I made it simpler. Normal forms were a mistake

### Technical considerations
This changes the way we persist tags from the convoluted SQL db to simply storing the tag ids inline on the time entries.

_William, you're a monster, why would you do that???_
There was no trivial way of having the current/backup/sync status data the way things were. The only benefit we were getting was the FK ensuring tags exist, but since we are already doing so much juggling with the serverIds anyways, that benefit gets completely outweighted by the complexity of having to deal with `DatabaseTimeEntry` `DatabaseTimeEntryTag` `DatabaseTimeEntryWithTags` and everything that comes with it.

This makes the changes to the DAO that are needed for #750. Also:
image

### Relationships
Refs #750

### Pull Request checklist

### Pointers
Lots of 🔥 and boilerplate changes, so reviewing it should be simple enough

### Review template

## Merge permissions
Anyone, anytime


database/src/main/java/com/toggl/database/dao/TimeEntryDao.kt
```diff
@@ -1,119 +1,65 @@
...
-        return id
+    suspend fun start(timeEntry: DatabaseTimeEntry): DatabaseTimeEntry {
+        stopRunningTimeEntries(timeEntry.startTime.current)
+        val id = insert(timeEntry)
+        return get(id)
     }
 
     @Transaction
-    fun stopRunningTimeEntries(now: OffsetDateTime): List<DatabaseTimeEntry> {
+    suspend fun stopRunningTimeEntries(now: OffsetDateTime): List<DatabaseTimeEntry> {
```
simonrozsival, line 33: I guess this doesn't concern this PR, but `now` should be actually called `stopTime` 😄

database/src/main/java/com/toggl/database/properties/LongListSyncProperty.kt
```diff
@@ -0,0 +1,26 @@
+package com.toggl.database.properties
+
+data class LongListSyncProperty(
```
simonrozsival, line 3: Shouldn't this be actually a `LongSetSyncProperty`? It's probably gonna be used just for tag IDs and that's not a list of longs, that's a set of longs.

database-android/src/main/java/com/toggl/database/TogglTypeConverters.kt
```diff
@@ -58,6 +58,15 @@ class TogglTypeConverters {
     fun toStringList(value: String?): List<String>? =
         value?.split(",")
 
+    @TypeConverter
+    fun fromLongList(value: List<Long>?): String? =
+        value?.joinToString(",") { it.toString() } ?: ""
```
simonrozsival, line 63: Are we going to query the list somehow? Like does the list contain this ID? If that were the case, we should add the separator also to the start and end of the string. That way we would have `"|123|456|789|"` and then we would check if it contains a substring `"|123|"`.

simonrozsival (commented):
It's kinda scary to have foreign keys stored in a comma separated list... I understand why you decided to do it this way though. I'm kinda unsure if it's a good thing or not. I would probably prefer having a DB view to do the flattening somehow... but it might be an overkill just to keep the list of IDs.

database-android/src/main/java/com/toggl/database/TogglTypeConverters.kt
heytherewill: I don't think there's ever a need to query the list for a specific tag id. And if there is, we can do it in code after fetching

database/src/main/java/com/toggl/database/properties/LongListSyncProperty.kt
heytherewill: You are right, tags are sets. I changed this everywhere

database/src/main/java/com/toggl/database/dao/TimeEntryDao.kt
heytherewill: I've made this change since why not

database-android/src/main/java/com/toggl/database/TogglTypeConverters.kt
simonrozsival: BTW how will we manage this list when a tag is deleted? Imagine this scenario:
- a TE has a list of tags A, B
- sync
- locally you edit it and add also the tag C
- there's a change on the server which deleted tag A
- sync
   - the tag A is deleted from the local database
   - the local change to the TE's set of tags survives 3WM, now it references A, B, and C, even though the A's ID in that string doesn't point to a valid row in the tags table

heytherewill: And how would that be prevented by having FKs? It'd prevent the tag from being deleted by leaving the whole app in an invalid state. It's weird.

simonrozsival: SQLite doesn't have the cascading FK deletion mechanism? Then it won't help anyway... we'll just have to be careful with FKs then and drop them if we can't find the referenced entity in the local DB.

simonrozsival: Cascading should be possible in SQLite: https://www.techonthenet.com/sqlite/foreign_keys/foreign_delete.php

simonrozsival: (cascading is the desired behavior for M:N relationships, for 1:1 and 1:N `SET NULL` is most likely the desired  behavior (https://www.techonthenet.com/sqlite/foreign_keys/foreign_null.php))

simonrozsival (approved):
Let's go this way and let's improve it if we run into real-world problems with this solution.

PR #800, opened 2020-09-01
🦴 Add push sync step skeleton 

### LINKED ISSUE #773: ⏫ Push Sync Step skeleton implementation
## Description
The basic of Push Sync Step should be implemented the same way [PushSyncInteractor is implemented in Xamarin](https://github.com/toggl/mobileapp/blob/54bb2f1194b97fac59113222caaf390a4acf1581/Toggl.Core/Interactors/Sync/PushSyncInteractor.cs#L35)

## Dependencies
None

## Definition of done
- [ ] `pushRequestIdentifier` is implemented
- [ ] `PushSync` is implemented the same way Xamarin does PushSyncInteractor
- [ ]  Empty implementations of `PreparePushRequest` (#774) `MarkEntitiesAsSyncing`(#775), `ProcessPushResult`(#779) and `Clear` methods are called correctly

heytherewill:
### Summary
This adds the structure needed for the push sync step of syncing 

### Technical considerations
I'm currently deleting the outstanding push request id when the user logs out. I think this is correct, but I'd love some confirmation on that.
Also, the idea on how to handle timeouts comes from [here](https://howtodoinjava.com/retrofit2/connection-timeout-exception/). I assume that all other exceptions that can happen are handled elsewhere rather than swallowed, right, @simonrozsival?

### Relationships
Closes #773 

### Pull Request checklist

### Pointers
Reviewing commit by commit is definitely simpler

### Review template

## Merge permissions
Anyone, anytime


sync/src/main/java/com/toggl/sync/sequence/PushSync.kt
```diff
@@ -1,10 +1,53 @@
...
+            )
+        }
+
+        try {
+            val id = UUID.randomUUID()
+            pushRequestIdStorage.set(id)
+            requestProcessor.markEntitiesAsSyncing(request)
+            val response = syncApiClient.push(id, request)
+            responseProcessor.process(response)
+            pushRequestIdStorage.clear()
```
simonrozsival, line 38: Now that I'm looking at this piece of code, it looks to me that there's a lot of things going on in the `run` function. It would be much better if at least this piece of code which is the actual important part that pushes data to the server was extracted to a private method.

simonrozsival (changes requested):
It looks great. Deleting the push req UUID on logout is correct. If we're not doing it in the Xamarin app, it's a bug.

All the other exceptions, except for the request timeout should be handled by some global sync error catching code (like the `processError` method in the `SyncManager` in the Xamarin app).

Now that you have the skeleton, you can already test it - especially how the push request ID is handled in different results of the HTTP request. Basically just port https://github.com/toggl/mobileapp/blob/develop/Toggl.Core.Tests/Interactors/Sync/PushSyncInteractorTests.cs

simonrozsival (approved):
LGTM

PR #801, opened 2020-09-01
Accessible database client view

### LINKED ISSUE #780: AccesableDatabaseClient view
## Description
We need to make sure we are accessing only Clients that are "accessible" outside of the sync logic (UI).
Accessible means that `isDeleted = false` and that the referenced workspace is also not deleted. 
The best way to implement this is probably to create an [database view](https://developer.android.com/training/data-storage/room/creating-views) called something like AccesableDatabaseClient and ClientDao should then mostly use this view instead of the real table. Alternatively, we could just put this join in every relevant query we make without using the view.

I gave an example of databaseView solution in #778

## Dependencies
None

## Definition of done
Either
- [x] `AccesableDatabaseClient` database view is created
- [x] The view is used in ClientDao and the rest of the related code is fixed
or 
- [x] All relevant queries in ClientDao are enhanced with join to workspaces and filtered by deleted workspace

simonrozsival:
### Summary
This is a straightforward implementation of the issue.

I struggled with the unit tests for some time. I was getting some cryptic errors from java which I didn't understand. Later I found out that the `database-android` module test use the Roboelectric test runner unlike the other and that the test function annotations must be `org.junit.Test` and not `org.junit.jupiter.api.Test` which are used for example in the `api` module and that the test functions can't be `suspend` and that I need to use `runBlocking` instead of `runBlockingTest` which didn't work for some reason.

Please let me know if I'm doing something incorrectly and if maybe there's a way to use `runBlockingTest` instead of `runBlocking` or if some other improvements are possible.

### Relationships
Closes #780 

### Pull Request checklist

### Pointers
Check the unit tests - can you think of some other test I could add to test this?

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑


database/src/main/java/com/toggl/database/dao/ClientDao.kt
```diff
@@ -13,6 +14,9 @@ interface ClientDao {
     @Query("SELECT * FROM clients")
     fun getAll(): Flow<List<DatabaseClient>>
```
heytherewill, line 15: Will this ever be used? We should delete this method if using it is deemed as wrong

simonrozsival: It will be used - it's necessary for syncing.

simonrozsival: It might be called `getAllIncludingInaccessible` to scare of people so that they use it only when they know what they're doing 😆

heytherewill: Used for syncing where?
For pushing we wanna have a complex DB query that returns exactly what we want
For pulling we fetch items as they are needed. When will we ever need a flow that updates whenever anything changes?

simonrozsival: We definitely need to fetch all inaccessible (just inaccessible) for clean up. Maybe the `getAll` method isn't needed afterall. But now that I'm thinking about it, I'm not sure if we need the DB view or if a simple query with the `JOIN` and the `WHERE NOT w.isDeleted`, but I'm not sure actually... the `accessible_clients` (and the other `accessible_*` views) were supposed to be used for more complicated queries, so that we don't forget to add the filtering somewhere where it would be necessary.

Maybe we should remove _this_ `getAll` and we would only bring it back if we actually needed it. Currently we have a DAO which has all the basic CRUD methods whether they're already needed or not so we should maybe do some clean up for all the DAOs after we finish working on syncing or something like that.

I skipped the part of the issue where it says: 
```
- [ ] The view is used in ClientDao and the rest of the related code is fixed
or
- [ ] All relevant queries in ClientDao are enhanced with join to workspaces and filtered by deleted workspace
```
... so I'll do that first and then we'll see if some use case for `getAll` is still there or not.

simonrozsival:
It's actually kinda strange now 🤔 I don't like that the DAO returns either the `DatabaseClient` or `AccessibleDatabaseClient`, which are internally exactly the same. On the other hand, the UI layer can be sure that they're getting exactly what they asked for - either something that can safely be shown in the UI or everything that's in the DB.

database/src/main/java/com/toggl/database/dao/ClientDao.kt
heytherewill: Oh, I think you are missing an important detail: This specific query returns a flow that updates whenever there's a change in the DB. This is used by subscriptions, which are entirely UI related. This means we will never need something like that. For cleanup etc we would need a method that returns a snapshot of the database and not a live flow that updates whenever something changes

simonrozsival: Yeah, that's a good catch. I'll remove it. Also we might use the `SyncDao` for clean-up rather than this DAO directly.

simonrozsival:
This PR is no longer needed.

PR #802, opened 2020-09-01
✨ Add Google signup and integrate terms acceptance/rejection

### LINKED ISSUE #665: Add AcceptTermsButtonTapped action
## Description
Add the action, wire it up to the UI. It should be emitted when user accepts terms and conditions. The action emits SignUserUpWithThirdPartyEffect or SignUserUpEffect, depending on where it originated (welcome screen / email, password signup screen).

## Dependencies
#646 
## Definition of done
It’s wired with the UI

### LINKED ISSUE #666: SignUserUpWithThirdPartyEffect - google
## Description
Implement the signup flow with google. Emit SetUser/SetError
## Dependencies
#664 
## Definition of done
It is possible to sign up with google
Tests are added

### LINKED ISSUE #667: Add MustAcceptTerms action
## Description
Add the action and reduce it so that the terms and conditions view from #646   is shown to the user

This action should be accessible in welcome and signup submodules.

## Dependencies
#646
## Definition of done
- Action is added, reduced, tested

### LINKED ISSUE #719: Add RejectTerms action
## Description
If user rejects the terms (by closing the UI #646 or hitting back button) during third party signup, we need to emit this action and reduce it so the welcome screen is shown and user is not signed up.

## Dependencies
#646 

## Definition of done
- [ ] action is added, reduced, tested
- [ ] UI is wired up

### LINKED ISSUE #720: Introduce accept terms screen to the email/password signup flow
## Description
Depending on how the discussion here https://toggl.slack.com/archives/CUHSVKLUV/p1596792481053600?thread_ts=1596013389.005900&cid=CUHSVKLUV goes, we'll have to add some copy to the signup screen, or introduce another step during signup - showing #646 

This issue is here so we don't forget to take action, but is not actionable until decisions are made in the slack thread.

**Update**
Decision has been made to show the modal from #646 during email/pass signup. Therefore, signup has to be changed so that (see the definition of done section):

## Dependencies
#646 
#667 
#665 

## Definition of done
- [ ] tapping signup button emits MustAcceptTerms
- [ ] after accepting terms, signupeffect is emitted
- [ ] when terms are not accepted, user is taken back to the signup screen

### LINKED ISSUE #810: Make sso linking happen
## Description
☝️

pdrj:
### Summary
This adds google signup and integrates accepting terms in both signup flows (google and email/password)

### Technical considerations
I've added a terms reducer in the common package. Then welcome and signup fragments pickup the terms acceptance event and fire up actions that sign the user up.

### Relationships

Closes #665
Closes #667
Closes #666 
Closes #719 
Closes #720 
Closes #810

### Pull Request checklist

### Pointers
Add your debug sha1 hash to the staging firebase project. Then use google signup and regular signup. Validate you have to accept terms and that you can also reject them via the back button.

### Review template

- [ ] The code considers i18n

## Merge permissions
Me


onboarding/src/test/java/com/toggl/onboarding/common/domain/TermsAcceptedActionTest.kt
```diff
@@ -0,0 +1,27 @@
...
+
+class TermsAcceptedActionTest : CoroutineTest() {
+    private val reducer = TermsReducer()
+    private val state = onboardingStateWithTermsRoute()
+
+    @Test
+    fun `sets terms accepted to true and pops backstack`() = runBlocking {
+        reducer.testReduceState(state, OnboardingAction.TermsAccepted) { newState ->
+            newState.localState.termsAccepted shouldBe true
+            newState.backStack shouldBe state.backStack.pop()
```
daividssilverio, line 19: I think this test would be more robust if you explicitly specified the expected route here and on the other similar tests.

onboarding/src/main/java/com/toggl/onboarding/signup/domain/SignUpState.kt
```diff
@@ -11,15 +11,17 @@ data class SignUpState(
     val user: Loadable<User>,
     val email: Email,
     val password: Password,
-    val backStack: BackStack
+    val backStack: BackStack,
+    val termsAccepted: Boolean?
```
daividssilverio, line 15: The `termsAccepted` is a nullable boolean, don't you think that's a bit weird? It's also never set to false.
Reading the tests and usage, it sort of makes sense, but I thought it was awkward. Maybe we could use an enum instead?

onboarding/src/main/java/com/toggl/onboarding/signup/domain/SignUpReducer.kt
```diff
@@ -33,23 +34,28 @@ class SignUpReducer @Inject constructor(
...
+                    state.mutate { copy(user = Loadable.Loading, termsAccepted = null) } returnEffect signUserUpEffect(
+                        currentState.email as Email.Valid,
+                        currentState.password as Password.Strong
                     )
                 } else {
                     noEffect()
                 }
             }
+            SignUpAction.SignUpButtonTapped -> if (state().isSignUpAllowed()) state.mutateWithoutEffects {
+                copy(
```
daividssilverio, line 49: Don't we have to accept the terms first in the terms screen for this function `isSignUpAllowed()` to be evaluable to true?

pdrj: Ah yes, you're right. I'll change the check there.

onboarding/src/main/java/com/toggl/onboarding/signup/domain/SignUpState.kt
pdrj: Yeah, that's a bit weird 😄 I'll change it to a regular bool and init to false.

onboarding/src/test/java/com/toggl/onboarding/common/domain/TermsAcceptedActionTest.kt
pdrj: Not sure I agree with this one. The terms screen can appear in the email flow and in the google flow. The bottom of the stack will be different in both these cases. That's why my intention is that we only care to check if we popped the top of the backstack (ie. dismissed ourselves) and we don't care what's underneath. What do you think?

pdrj:
> I couldn't test the SSO flow though.

No signup can happen during an SSO flow. Are you referring to the linking flow and linking via google login? If so, that's something still todo. Maybe I'll manage to squeeze it into this PR by the time you get here ;)

pdrj:
@daividssilverio I've added the SSO linking stuff. I'll make sure you're setup such that you can test this. On your side, you just have to point the app to prod endpoints.

The easiest way to repetitively test linking is to have two accounts in the sso test workspace on prod. One of these accounts should have admin rights. Then, leave the workspace with one account and invite it with the other. Accept the invite in web app. Then, when you sso-login with the newly invited account via the app, there should be a linking request.

pdrj:
You're setup in okta and in toggl with your main and +1 email. Let me know if stuff works.

onboarding/src/test/java/com/toggl/onboarding/common/domain/TermsAcceptedActionTest.kt
daividssilverio: 🤔 I see. That sounds alright, although we could also check that the size of the backstack changed just in case.

daividssilverio:
> > I couldn't test the SSO flow though.

> No signup can happen during an SSO flow. Are you referring to the linking flow and linking via google login? If so, that's something still todo. Maybe I'll manage to squeeze it into this PR by the time you get here ;)

Yes. That's what I meant. I tried going through the link SSO flow.

onboarding/src/test/java/com/toggl/onboarding/common/domain/TermsAcceptedActionTest.kt
pdrj: But then we'd be essentially testing that `pop()` works as it should, won't we?

daividssilverio: Not really, we wouldn't know if the backstack actually changed if the backstack was initially empty for example (it doesn't make much sense for this exact example though).

What I wanted to see tested is that the backstack actually changed, or that we actually called `pop`.
The problem I see is that
```
backStack(emptyList()) == backStack(emptyList()).pop()
```

pdrj: OK, I see what you mean. As you noted, the backstack is not initially empty here, but I'll add the size check nonetheless.

daividssilverio:
All the login & linking flows I could test worked! This is working great!

onboarding/src/main/java/com/toggl/onboarding/di/OnboardingModule.kt
```diff
@@ -162,9 +173,13 @@ object OnboardingApplicationModule {
...
         signUpReducer: SignUpReducer,
         ssoReducer: SsoReducer,
-        passwordResetReducer: PasswordResetReducer
+        passwordResetReducer: PasswordResetReducer,
+        termsReducer: TermsReducer,
+        commonSsoReducer: CommonSsoReducer
     ): OnboardingReducer {
         return combine<OnboardingState, OnboardingAction>(
+            termsReducer,
+            commonSsoReducer,
```
daividssilverio, line 182: I was wondering... the order we put these reducers matter, right?
Shouldn't this `commonSsoReducer` be put after the other reducers? do we want to run the reduce the `*.SetUser` actions first on the common reducer and then on the login/welcome reducers or the other way around?

pdrj: Yes, it does matter, but I don't think we care too much about ordering in this case 😄 I'll move it to the end though, it feels more appropriate to have it there.

PR #803, opened 2020-09-01
✨ Add the date range picker view and bind it to the reports state

### LINKED ISSUE #569: Create the date range picker view and bind it to the reports state
## What needs to be done?
- [x] Create a pop-up that is visible when there's a non-nil `selection`
- [x] The pop-up should have a close and accept buttons, which should trigger `.dateRangePickerAcceptButtonTapped` and `.dateRangePickerCloseButtonTapped` respectively
    - [x] The accept button should be disabled when the selection does not have an end date
- [x] The pop-up should hold two collection/recycler views:
    - [x] A multi-page/row/column one for the calendar
    - [x] A single row scrollable view for the shortcuts
- [x] The calendar view should show 12 months, with the latest on being the current one
- [ ] Each day cell of the calendar should have 6 variants: enabled, disabled, selected as a single cell, selected as a range range, selected as a range end, and selected as a middle part of a range.
- [x] Each shortcut cell should have two variants: selected and not
    - [x] If the range in `selection` matches the shortcut's `dateRange`, then the shortcut should also be rendered as selected

### Dependencies
Depends on #565, #566, #567 and #568

daividssilverio:
### Summary
This adds the date range picker composable for the reports page! In compose! COMPOSE!

### Technical considerations
The pager was shamelessly copied from the jetpack compose samples (Jetcaster) (it's a bit glitchy for now). I'm hoping we'll be getting an official pager api and we can use it later.
Everything else was inspired a lot from the jetpack samples. Specially the calendar from Crane sample.

Feel free to make your suggestion.

### Relationships
Closes #569

### Pull Request checklist

### Pointers
Please note that you might need to set the default workspace id for the `loadReportEffect` call on `ReportsReducer` for a workspace id where you have some data to actually see something working on the reports.

### Review template
Check the previews. They are helpful to see how stuff is working.

- [ ] The code considers i18n

## Merge permissions
🦑 it


reports/src/main/java/com/toggl/reports/domain/ReportsSelector.kt
```diff
@@ -43,6 +52,106 @@ class ReportsSelector @Inject constructor(
         )
     }
 
+    private fun dateRangePicker(state: ReportsState): List<ReportsViewModel> = state.let {
```
heytherewill, line 55: Why include the picker like this, as part of the list? 🤔 

Since it overlays on top of everything else, it feels weird to have it as part of the list. I'd rather have another selector for it.

reports/src/main/java/com/toggl/reports/domain/ReportsViewModel.kt
```diff
@@ -1,7 +1,16 @@
 package com.toggl.reports.domain
 
+import com.toggl.models.common.DateRange
+import java.time.OffsetDateTime
+
 sealed class ReportsViewModel {
     data class DatePicker(val formattedDate: String) : ReportsViewModel()
+    data class DateRangePicker(
```
heytherewill, line 8: Again, it's really weird to have this as part of the picker since this is a separate thing that is overlaid on top of the actual list of report data

reports/src/main/java/com/toggl/reports/ui/composables/Pager.kt
```diff
@@ -0,0 +1,272 @@
...
+import kotlin.math.roundToInt
+
+/**
+ * This is a modified version of
+ * https://gist.github.com/adamp/07d468f4bcfe632670f305ce3734f511,
+ * copied from the compose samples Jetcaster sample app
+ * https://github.com/android/compose-samples/blob/master/Jetcaster/app/src/main/java/com/example/jetcaster/util/Pager.kt
+ */
+
+class PagerState(
```
heytherewill, line 52: A pager is such a common thing that it should be moved to the `common-feature` module

reports/src/main/java/com/toggl/reports/ui/composables/ReportsPage.kt
```diff
@@ -67,13 +68,18 @@ fun ReportsContent(
                     is ReportsViewModel.DonutChartLegend -> DonutChartLegend(viewModel.legend)
                     is ReportsViewModel.Error -> Error(viewModel.errorMessage)
                     is ReportsViewModel.AdvancedReportsOnWeb -> AdvancedReportsOnWeb(dispatcher)
+                    is ReportsViewModel.DateRangePicker ->
+                        TogglDialog(onDismissRequest = { dispatcher(ReportsAction.DateRangePickerCloseButtonTapped) }) {
+                            CalendarDateRangePicker(viewModel = viewModel, dispatcher = dispatcher)
+                        }
```
heytherewill, line 74: Just so I stress all places you have to change: This should not happen as part of the list, but rather on top of the list: https://toggl.invisionapp.com/d/main#/console/19932960/425826079/preview

reports/src/main/java/com/toggl/reports/ui/composables/CalendarDateRangePicker.kt
```diff
@@ -0,0 +1,533 @@
...
+            Offset(size.width / 2, 0f)
+        }
+        val size = Size(width = size.width / 2, height = size.height)
+
+        drawRect(size = size, topLeft = offset, color = color)
+    }
+}
+
+@Composable
+private fun ChoiceChip(
```
heytherewill, line 366: This (and everything that is only used to render it) can also probably be extracted to the `common-feature` package

reports/src/main/java/com/toggl/reports/ui/composables/ReportsPage.kt
daividssilverio: The dialog appears on top of the content. I've made it part of the list to avoid creating another selector. It looked alright to have all he view models together in this `when`.

daividssilverio:
@heytherewill I've fixed the ripples and addressed your other comments o/
Thanks for the review o/

reports/src/main/java/com/toggl/reports/domain/ReportsDateRangePickerViewModel.kt
```diff
@@ -0,0 +1,48 @@
...
+import java.time.OffsetDateTime
+
+data class ReportsDateRangePickerViewModel(
+    val months: List<DateRangePickerMonthInfo>,
+    val shortcuts: List<DateRangePickerShortcutInfo>,
+    val daysOfTheWeekLabels: List<String>,
+    val selectionIsCommittable: Boolean
+) {
+    companion object {
+        val Empty = ReportsDateRangePickerViewModel(emptyList(), emptyList(), emptyList(), false)
```
heytherewill, line 13:
```suggestion
        val empty = ReportsDateRangePickerViewModel(emptyList(), emptyList(), emptyList(), false)
```

heytherewill (approved):
I'm approving for speed, but please fix this before we proceed (seems to be some leftover from Xamarin)

PR #805, opened 2020-09-02
Process pulled time entries

### LINKED ISSUE #750: Implement TimeEntries response processing

heytherewill:
### Summary
This allows us to fetch time entries from the sync server!!

### Technical considerations
This is a rather simple one. It's similar to the project processing PR: We check the important properties and override the ones that are not editable. Pretty straight-forward.

I also added the needed check that makes us only persist the since parameter when processing works.

I need @simonrozsival to confirm that this works. It's, apparently, missing the check that verifies if the time entry is relevant according to some time entry since parameter. I don't fully understand how that system works or how to port it to our new solution

### Relationships
Closes #750

### Pull Request checklist

### Pointers
Review each commit individually

### Review template

## Merge permissions
Anyone, anytime


sync/src/main/java/com/toggl/sync/processing/ProcessPulledTimeEntries.kt
```diff
@@ -0,0 +1,95 @@
...
+        projectId = projectId updateWith resolvedProjectId,
+        taskId = taskId updateWith resolvedTaskId,
+        tagIds = tagIds updateWith resolvedTagIds,
+        isDeleted = isDeleted updateWith resolvedIsDeleted,
+        syncStatus = if (isDirty) EntitySyncStatus.SyncNeeded else EntitySyncStatus.InSync
+    )
+}
+
+private fun Long.toRunningTimeEntryAwareDuration() =
+    if (this < 0) null else Duration.ofSeconds(this)
```
simonrozsival, line 95: Could we move this to the API module maybe? The V9's implementation detail, that for the running TE the duration is negative should not leak into the sync layer IMO. We don't have to let this slow us down right now, but maybe we could at least create an issue to revisit it later?

simonrozsival (approved):
I think this can move ahead. The "is relevant for syncing" thing was added to the Xamarin app because the list of pulled TEs always contained the currently running TE (if any) even when it hasn't changed since `since`. If we run into problems with the running TEs (you stop a TE and it reappears), we'll have to add it back. I don't remember if we didn't solve the issue with the 3WM and it's not necessary anymore (and it's just a leftover), or if it will be necessary. If we run into the issue, we'll know what to change in order to fix it.

Issue #781, opened 2020-08-21
AccesableDatabaseProject view


semanticer:
## Description
We need to make sure we are accessing only Projects that are "accessible" outside of the sync logic (UI).
Accessible means that `isDeleted = false` and that the referenced workspace is also not deleted. 
The best way to implement this is probably to create an [database view](https://developer.android.com/training/data-storage/room/creating-views) called something like AccesableDatabaseProject and ProjectDao should then mostly use this view instead of the real table. Alternatively, we could just put this join in every relevant query we make without using the view.

I gave an example of databaseView solution in #778

## Dependencies
None

## Definition of done
Either
- [ ] `AccesableDatabaseProject` database view is created
- [ ] The view is used in ProjectDao and the rest of the related code is fixed
or 
- [ ] All relevant queries in ProjectDao are enhanced with join to workspaces and filtered by deleted workspace

simonrozsival:
Superseded by #807

Issue #778, opened 2020-08-21
AccesableDatabaseTimeEntry view


semanticer:
## Description
We need to make sure we are accessing only Time Entries that are "accessible" outside of the sync logic (UI).
Accessible means that `isDeleted = false` and that the referenced workspace is also not deleted. 
The best way to implement this is probably to create an [database view ](https://developer.android.com/training/data-storage/room/creating-views) called something like AccesableDatabaseTimeEntry and TimeEntryDao should then mostly use this view instead of the real table. Alternatively, we could just put this join in every relevant query we make without using the view.

Here is the example how the `@DatabaseView` solution could look like:

```kotlin
@DatabaseView("SELECT * FROM time_entries AS te INNER JOIN workspaces AS w ON te.workspaceId_current = w.id" +
    "WHERE te.is_deleted = 0 AND w.is_deleted = 0")
data class AccesableDatabaseTimeEntry(
    val id: Long = 0,
    val serverId: Long?,
    @Embedded(prefix = "description_")
    val description: StringSyncProperty,
    @Embedded(prefix = "startTime_")
    val startTime: OffsetDateTimeSyncProperty,
    @Embedded(prefix = "duration_")
    val duration: NullableDurationSyncProperty,
    @Embedded(prefix = "billable_")
    val billable: BooleanSyncProperty,
    @Embedded(prefix = "workspaceId_")
    val workspaceId: LongSyncProperty,
    @Embedded(prefix = "projectId_")
    val projectId: NullableLongSyncProperty,
    @Embedded(prefix = "taskId_")
    val taskId: NullableLongSyncProperty,
    @Embedded(prefix = "isDeleted_")
    val isDeleted: BooleanSyncProperty
)
```

Or just take the join and use it in every single query

## Dependencies
None

## Definition of done
Either
- [ ] `AccesableDatabaseTimeEntry` database view is created
- [ ] The view is used in TimeEntryDao and the rest of the related code is fixed
or 
- [ ] All relevant queries in TimeEntryDao are enhanced with join to workspaces and filtered by deleted workspace

simonrozsival:
Superseded by #807

PR #811, opened 2020-09-02
Filter accessible entities

### LINKED ISSUE #807: Return only accessible entities to the UI
As a result of an [internal discussion](https://toggl.slack.com/archives/C02APMYTM/p1599038414135000) we decided not to use DB views for selecting just accessible entities.

We must make sure that we don't leak inaccessible entities to the UI by changing the `SELECT`s in all of our DAO which return entities to the UI:

- [ ] `ClientDao`
- [ ] `ProjectDao`
- [ ] `TagDao`
- [ ] `TaskDao`
- [ ] `TimeEntryDao`
- [ ] `WorkspaceDao`

This issue supersedes these issues: #778 #780 #781 #782 #783

simonrozsival:
### Summary
This PR changes the `SELECT` queries to return only accessible entities.

This was done instead of adding DB views (#778, #780, #781, #782, #783).

### Relationships
Closes #807 

### Pull Request checklist

### Pointers
Check the unit tests. Don't get scared by the fact that they're all the same.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑


database-android/src/test/java/com/toggl/database/dao/TimeEntryDaoTest.kt
```diff
@@ -0,0 +1,49 @@
...
+            val workspaceId = database.createDummyWorkspace(deleted = true)
+            database.createDummyTimeEntries(workspaceId, count = 20)
+
+            val accessibleDatabaseTasks = database.timeEntryDao().getAllAccessible().first()
+            accessibleDatabaseTasks.size shouldBe 0
+        }
+    }
+
+    @Test
+    fun `selects tasks time entries an accessible workspace`() {
```
heytherewill, line 35:
```suggestion
    fun `selects time entries from an accessible workspace`() {
```

heytherewill (dismissed):
Please make sure to test that the `isDeleted` property of time entries as well

PR #812, opened 2020-09-02
🤗 Group similar time entries according to settings

### LINKED ISSUE #806: Properly react to the group time entries setting
## Description
Right now the setting is persisted (and even synced) but the log ignores the flag

## Definition of done
- [ ] The log groups/ungroups entries based on the selected setting value

heytherewill:
### Summary
This makes the log actually respect the settings that dictate whether time entries should be grouped or not

### Technical considerations
This is dead simple

### Relationships
Closes #806

### Pull Request checklist

### Pointers
This is fairly straight forward

### Review template

## Merge permissions
Anyone, anytime


timer/src/test/java/com/toggl/timer/log/domain/TimeEntriesLogSelectorTests.kt
```diff
@@ -438,16 +467,28 @@ class TimeEntriesLogSelectorTests : CoroutineTest() {
         groupedTimeEntries shouldBe expectedGroupedTimeEntries
     }
 
+    @Test
+    fun `does not group similar time entries as per the setting`() = runBlockingTest {
```
pdrj, line 471: Rather than having two separate tests, you can consider `@ParametrizedTest` + `@ValueSource(booleans = [true, false]` for the grouping param. Up to you to decide what's cleaner ;)

pdrj (approved):
Couldn't actually functionally test by setting the setting as our toggles don't work 😄 Worked as advertised when hardcoded.

Only one tiny thing to consider.

PR #821, opened 2020-09-03
Mark pushed entities as Syncing

### LINKED ISSUE #775: Mark Pushed Entities as Syncing
## Description
Mark all entities from push request as `syncing` in the database. 

[Xamarin code](https://github.com/toggl/mobileapp/blob/54bb2f1194b97fac59113222caaf390a4acf1581/Toggl.Storage.Realm/Queries/MarkeEntitiesAsSyncingQuery.cs)

## Dependencies
#773 #774 

## Definition of done
- [ ] Method for marking entities as `synced` is implemented
- [ ] Add tests if possible

simonrozsival:
### Summary
This PR adds a transaction which marks all of the entities referenced in the `PushBody` with the entity sync status `Syncing`. This is necessary to identify whether the entity has changed locally while the server was processing the push request.

### Technical considerations
- In order to easily extract the local IDs from the `PushBody`, I added a new type `TemporaryId` data class and I added methods necessary to serialize and deserialize it + ✅ 
- I added code to get the list of local and server IDs from the a list of `PushAction`s + ✅ 
- I added the transaction itself, ~tests are currently missing, they still need to be added, that's why this PR is still just a draft~ - tests are already added ✅

I decided to include also the entities which are being created, unlike in the Xamarin app, where we skipped these entities. In principle, we could still skip this step and always overwrite the local state of created entities but I think it's better to be consistent and modify the state of all entities no matter what action is pushed.

### Relationships
Closes #775 

### Pull Request checklist

### Review template

- [ ] The code considers i18n

## Merge permissions
🚫


heytherewill (approved):
Really straightforward!

PR #825, opened 2020-09-04
🚨 Improve onboarding error handling 

### LINKED ISSUE #818: Display email errors on login screen
## Description
Display email errors on login screen (as we do in the signup one)

## Definition of done

- [ ] invalid emails produce an error displayed below the email field

### LINKED ISSUE #819: Display error when google login fails
## Definition of done

- [ ] when google login fails, an error toast is shown

### LINKED ISSUE #820: Clear user errors when navigating
## Definition of done

- [ ] when error happens on sso screen, it should not carry over to the email login screen
- [ ] when error happens during google login it should not carry over to sso and email login screens
- [ ] when error happens during regular signup/login it is not carried over to sso screen
- [ ] scenarios not described above are indentified and handled too

pdrj:
### Summary
- Clears errors when moving between various onboarding screens.
- Displays google login errors via toast. 
- Displays email validation errors on login screen.

### Technical considerations
I introduced another common reducer to handle the cleanup on back presses.

### Relationships

Closes #819 
Closes #820 
Closes #818 

### Pull Request checklist

### Pointers
Trigger errors in various screens and validate you can't see them in other screens. Trigger a google error and validate you see a toast.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


onboarding/src/main/java/com/toggl/onboarding/welcome/ui/WelcomeFragment.kt
```diff
@@ -93,9 +97,14 @@ class WelcomeFragment : Fragment(R.layout.fragment_welcome) {
...
         store.state
             .map { it.termsAccepted }
-            .filter { it == true }
+            .filter { it }
             .onEach { store.dispatch(WelcomeAction.SignUserUpWithThirdParty) }
             .launchIn(lifecycleScope)
+
+        store.state
+            .mapNotNull { if (it.backStack.last() is Route.Welcome) it.user as? Loadable.Error else null }
+            .onEach { context?.run { Toast.makeText(this, it.failure.errorMessage, Toast.LENGTH_LONG).show() } }
```
daividssilverio, line 106: The only problem I see with this approach is that if the state somehow changes while the user error is still in the state, the toast will be duplicated. But it works for now.

daividssilverio (approved):
Minor concern about duplicated toasts, but LGTM.

onboarding/src/main/java/com/toggl/onboarding/welcome/ui/WelcomeFragment.kt
pdrj: That's a good point. For this to happen though, first we'd have to have an error during google login and the second attempt would have to go through an action that modifies state and not the user, so for instance the 'must accept terms' one. Then, when we're back to the welcome screen, a toast should appear again. 

I'll add a cleanup step that happens each time the google button is pressed.

pdrj:
I've set the user to loading on each google login attempt @daividssilverio. Is there any other scenario you can think of that would trigger the double toast?

daividssilverio:
> Is there any other scenario you can think of that would trigger the double toast?

Not really @pdrj. :shipit:

PR #828, opened 2020-09-04
🔃 Prepare push request

### LINKED ISSUE #774: Prepare Push Request
## Description
Implement a method that will get all dirty entities from the database and construct **pushRequest**.

[Xamarin code](https://github.com/toggl/mobileapp/blob/54bb2f1194b97fac59113222caaf390a4acf1581/Toggl.Core/Interactors/Sync/PreparePushRequestInteractor.cs#L30) for reference

We should transform these entities into "sync ready" entities. This means that all ids must be either server ids or negative local ids if server id is not yet available (that is true for entities created locally and synced for the first time). This is also true for foreign keys, so for example `Project` needs to be joined with `Client` and `Project.clientId` is then either Client server id or a negative local one if the `Client` hasn't been synced yet. The more you can do within sqlite the better, [ifnull](https://www.sqlitetutorial.net/sqlite-functions/sqlite-ifnull/) could be really useful here. 

Consider creating [database view](https://developer.android.com/training/data-storage/room/creating-views) for every "SyncReadyEntity".

## Dependencies
#773 

## Definition of done
- [ ] Method for getting dirty "sync ready" `Workspaces` is implemented
- [ ] Method for getting dirty "sync ready" `Clients` is implemented
- [ ] Method for getting dirty "sync ready" `Projects` is implemented
- [ ] Method for getting dirty "sync ready" `Tags` is implemented
- [ ] Method for getting dirty "sync ready" `Preferences` is implemented
- [ ] Method for getting dirty "sync ready" `User` is implemented
- [ ] Method for getting dirty "sync ready" `TimeEntries` is implemented
- [ ] All entities are used to create `pushRequest`
- [ ] Add tests if possible

simonrozsival:
### Summary
This PR is part of the Push Sync implementation.

#### Serialization

The `OptionalProperty<T>` is a type I added in order to serialize "nullable nullables" - there's a difference between putting a `"project_id": null` into the payload of an update action and omitting the property:

- in the first case, the ID will be set to `null` (we remove the project reference)
- in the second case, the referenced project is not changed

In order to serialize the property correctly in `OptionalPropertyAdapterFactory`, I took advantage of the way [Moshi implements writing `null`s](https://github.com/square/moshi/blob/master/moshi/src/main/java/com/squareup/moshi/JsonUtf8Writer.java#L201) and implemented a few handy extension methods to do what we need to do. If the internal implementation of Moshi changes some day, our unit tests will catch it.

#### Assembling `SyncApi*` models
Inside of the `RequestProcessor` class there are `forSyncApi` extension methods for all `Database*` models which convert the DB records it into the `SyncApi*` models. This conversion needs to achieve two things:

1. select which properties to _push_ in the payload
    - in the case when we're _creating_ stuff, we want to pass all _current_ to the API using `Some`
    - in the case when we're _updating_ stuff, we want to pass only _dirty_ properties using `Some` and the rest set to `None` and keep it unchanged
2. connect the entities with correct IDs
    - for _local_ entities, we want to use _negative_ local IDs
    - for entities which already have a counterpart on the server, we want to use _positive_ server IDs

I put most of the logic inside private (extension) functions inside the `RequestProcessor` class. This way I don't need to pass the injected DAOs as arguments. Let me know if you dislike that there's so much logic in a single class. It might also make sense to split the class in two and make the `RequestProcessor` responsible solely for assembling `PushBody` and move `markEntitiesAsSyncing` somewhere else.

### Relationships
Closes #774 

### Pull Request checklist

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑


api/src/main/java/com/toggl/api/network/models/sync/OptionalProperty.kt
```diff
@@ -0,0 +1,18 @@
...
+sealed class OptionalProperty<T> {
+    fun <U> map(fn: (T) -> U): OptionalProperty<U> =
+        if (this is Some<T>) Some(fn(value)) else None()
+
+    suspend fun <U> mapSuspend(fn: suspend (T) -> U): OptionalProperty<U> =
+        if (this is Some<T>) Some(fn(value)) else None()
+}
+
+data class Some<T>(val value: T) : OptionalProperty<T>()
+class None<T> : OptionalProperty<T>()
```
heytherewill: Won't `object None : OptionalProperty<Nothing>` work and actually be better?

sync/src/main/java/com/toggl/sync/processing/RequestProcessor.kt
```diff
@@ -31,4 +81,196 @@ class RequestProcessor @Inject constructor(
...
+            start = startTime.toOptional(justModified),
+            duration = duration.toOptional(justModified).map { it?.seconds ?: -startTime.current.toEpochSecond() }
+        )
+
+    private fun DatabaseUser.toSyncApiPreferences(): SyncApiPreferences =
+        SyncApiPreferences(
+            timeOfDayFormat = None(), // time of day format is missing completely currently (12h/24h)
+            // dateFormat = dateFormat.toOptional(justModified = true).map(DateFormat::valueOf),
+            // durationFormat = durationFormat.toOptional(justModified = true).map(DurationFormat::valueOf),
+            // collapseTimeEntries = groupSimilarTimeEntriesEnabled.toOptional(justModified = true)
```
heytherewill, line 223: Why are these lines commented out and the ones below are marked as none?

sync/src/main/java/com/toggl/sync/processing/RequestProcessor.kt
```diff
@@ -31,4 +81,196 @@ class RequestProcessor @Inject constructor(
...
+        anyIsDirty(defaultWorkspaceId.status, firstDayOfTheWeek.status)
+
+    private fun DatabaseUser.userNeedsSync(): Boolean =
+        anyIsDirty(groupSimilarTimeEntriesEnabled.status, dateFormat.status, durationFormat.status)
+
+    private fun anyIsDirty(vararg states: PropertySyncStatus): Boolean =
+        listOf(*states).contains(PropertySyncStatus.SyncNeeded)
+
+    private suspend fun <T, U> List<T>.mapSuspend(fn: suspend (T) -> U): List<U> =
+        asFlow().map(fn).toList()
```
heytherewill, line 269: Why is this at all needed? It feels really off

heytherewill (changes requested):

sync/src/main/java/com/toggl/sync/processing/RequestProcessor.kt
simonrozsival: Oh, oopsie. I need to figure these mappings, there's something wrong with them..

api/src/main/java/com/toggl/api/network/models/sync/OptionalProperty.kt
simonrozsival: It might be! I was struggling with this one and I was hoping someone would help me figure this out 😄 I'll try the `object` thing.

sync/src/main/java/com/toggl/sync/processing/RequestProcessor.kt
simonrozsival: It screamed at me when I wanted to plug a `suspend` lambda into normal `map`, so I added the extension method to make the `fun preparePushRequest()` lines short.

heytherewill: You just need to `map { suspendFn }` rather than `map(::suspendFn)`. If the parent scope is not a suspend function, you should make it suspend (a lot like `async Task`). If this is the reasoning, this extension should go

sync/src/main/java/com/toggl/sync/processing/RequestProcessor.kt
simonrozsival: Fixed.

api/src/main/java/com/toggl/api/network/models/sync/OptionalProperty.kt
simonrozsival: Done.

sync/src/main/java/com/toggl/sync/processing/RequestProcessor.kt
simonrozsival: It wasn't straightforward, but I believe I got the job done.

semanticer (changes requested):
Only some minor things to consider. I haven't done any functional tests yet (if they should be possible).

I'm also starting to think about some kind of abstraction we could create around our syncable entities because they have a lot in common but let's leave that for some other issue.

api/src/main/java/com/toggl/api/network/models/sync/MetaInfo.kt
```diff
@@ -0,0 +1,22 @@
...
+    @Json(name = "workspace_id")
+    val workspaceId: Long? = null,
+    @Json(name = "project_id")
+    val projectId: Long? = null
+) {
+    internal fun extractId(type: ActionType): Either<Long, Long>? =
+        when (type) {
+            ActionType.Create -> clientAssignedId?.id?.let { Either.Left(it) }
+            ActionType.Update, ActionType.Delete -> id?.let { Either.Right(it) }
+        }
```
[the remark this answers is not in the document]
simonrozsival: Yeah, I guess you're right 😄

database/src/main/java/com/toggl/database/dao/ClientDao.kt
```diff
@@ -4,21 +4,30 @@ import androidx.room.Dao
...
         JOIN workspaces w ON w.id = c.workspaceId_current
         WHERE NOT w.isDeleted
         """
     )
     fun getAllAccessible(): Flow<List<DatabaseClient>>
 
+    @Query("SELECT * FROM clients WHERE syncStatus = :syncStatus")
+    suspend fun getAllBySyncStatus(syncStatus: EntitySyncStatus): List<DatabaseClient>
+
+    @Transaction
```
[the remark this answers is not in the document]
simonrozsival: Hmm, I guess it's not needed after all 🤔

api/src/main/java/com/toggl/api/network/models/sync/MetaInfo.kt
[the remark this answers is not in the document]
simonrozsival: But I think I'll get rid of the "ID extraction" part completely. It seems unnecessary.

sync/src/main/java/com/toggl/sync/processing/RequestProcessor.kt
```diff
@@ -59,16 +59,16 @@ class RequestProcessor @Inject constructor(
         PushBody(
             timeEntries =
                 timeEntryDao.getAllDirty()
-                    .filter { !it.isDeleted.current || it.serverId != null }
-                    .mapSuspend { it.toPushAction() },
+                    .filter { !it.isLocalOnlyAndDeleted() }
```
heytherewill, line 73: No need to change, but there's a `filterNot` method

heytherewill (approved):

semanticer (approved):

PR #822, opened 2020-09-04
Workspace Plan View

heytherewill:
### Summary
This adds the workspace plan view

### Technical considerations
This version is more complex than the Xamarin version. In order to fetch the workspace plan data, we need to hit the API, something that is much harder to do in the Xamarin app. With Aurora's architecture, though, this is quite simple.

We need to poke the plans/subscriptions endpoints in order to calculate the user's plan. I also have a backup to calculate a best effort version with what we have locally so we can degrade gracefully when offline.

### Relationships
Closes #808

### Pull Request checklist
- [ ] The code does not hinder the app's performance
- [ ] The code does not introduce regressions
- [ ] The code in this PR is tested thoroughly enough that changing the behavior to an incorrect one breaks tests

### Pointers
There's a lot of changes that are structural changes, so reviewing commit by commit will be beneficial

### Review template

- [ ] The code is thoroughly tested in well written and resilient manner
- [ ] The code is performant and leads to an optimal user experience
- [ ] The code is testable and does not cut corners for the sake of delivering
- [ ] The code considers older platforms well and will work in older devices just like it will in new ones
- [ ] The code considers light and dark theme
- [ ] The code considers RTL and does not mention left/right
- [ ] The code considers i18n
  - [ ] There are no hardcoded strings in the UI
  - [ ] All strings are included in every language we support OR there's an issue requesting new translations
- [ ] There are no hard-coded numbers referring to dimensions or font sizes
- [ ] The code considers UI elements that are similar and extracts their similarities into styles to avoid duplication

## Merge permissions
Anyone, anytime

PR #829, opened 2020-09-04
🔧 Only show the advanced reports card for non pro workspaces

### LINKED ISSUE #578: Create the advanced reports on web cell
## What needs to be done?
- [ ] Create a `shouldShowAdvancedReportsOnWeb` selector in the reports feature. This selector should check whether we need to display the advanced report on web card and return a `ReportsCellViewModel.AdvancedReportsOnWeb`
- [ ] Create the actual cell/view, and populate it with the appropriate data. Designs are are still pending for both platforms.
- [ ] Discuss possible value animation with the team in #mobile-team

### Dependencies
Depends on #570

daividssilverio:
### Summary
This adds the missing part of what had to be done for #578
The layout was already there.

### Technical considerations
🤔 Not sure if we should make an api request like it's being done on #822 for the settings screen.
Should we be using the workspace plans instead of the `isPro` workspace extension?
Calling @dshr and @heytherewill to answer that question.

### Relationships
Closes #578
Maybe depends on #822

### Pull Request checklist

### Pointers
None.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it after we get a confirmation about the reasonability of the check being made to show the card.


heytherewill (changes requested):
We were showing to pro users only because we weren't doing the networking to fetch the actual user plan (and detecting a free plan can be done offline). With the code added in the other PR, we can easily how this screen to everyone, so closing the linked issue is just a matter of navigating to the correct route and adding tests

daividssilverio:
So, the workspace plan view currently works with the default workspace id from the user and there's no UI to change your workspace for visualization purposes only in said view.

@heytherewill  Should we make some adhoc adaptations directly on this PR? Or is it better to create some issues to handle it?

I think we would need to:
- Make the route `Route.WorkspacePlan` accept a parameter.
- Make the Settings state hold the `currently selected workspace` for the workspace plan view.
- Clean up the `currently selected workspace` when leaving the workspace plan view.
- Add some loading state in the workspace plan view...
- Change how we fire the workspace plan loading effect (I think we can fire it when the workspace plan view appears) instead of when the default workspace changes.

daividssilverio:
For now, I'm using the `isPro` offline extension and navigating to the workspace plan view as it currently works.

reports/src/main/java/com/toggl/reports/domain/ReportsReducer.kt
```diff
@@ -44,8 +44,12 @@ class ReportsReducer @Inject constructor(
             }
             is ReportsAction.ReportLoaded -> state.setReportData(Loadable.Loaded(action.reportData))
             is ReportsAction.ReportFailed -> state.setReportData(Loadable.Error(action.failure))
-            ReportsAction.AvailableOnOtherPlansTapped -> state.mutateWithoutEffects {
-                copy(backStack = backStack.push(Route.WorkspacePlan))
+            ReportsAction.AvailableOnOtherPlansTapped -> {
+                val selectedWorkspaceId = state().localState.selectedWorkspaceId
+                if (selectedWorkspaceId == null) noEffect()
```
heytherewill, line 49: If the user has not yet selected a workspaceId, we default to showing data of the default workspaceId (be it the user's default workspaceId or the first workspaceId when that is null). Why not do the same here?

daividssilverio: Good catch. As it is, if the user never selected a workspace and saw the `AvailableOnOtherPlans` card, it would be unresponsive.

heytherewill (approved):

PR #830, opened 2020-09-05
🐛 Fix user initial default workspace id setup

### LINKED ISSUE #642: Workspace not selected in settings on fresh install
## Description
On fresh app installs, when you go to settings, the default workspace name is not visible in the main settings screen and when you click on the row, the radio next to the name is not selected.

## Definition of done

- [ ] the workspace name is visible and selected on fresh installs

daividssilverio:
### Summary
This fixes the initial user setup.
We were setting the user received from the api into the state with outdated data. 
The database was setup with the correct default local workspace id, but the state was set with default server workspace id.

### Technical considerations
None.

I just renamed the the `set` method to `setAndConfigureDefaultWorkspace` to be more explicit.
But we could also have gone like this in all effects:
```
userRepository.set(user)
updatedUser = userRepository.get()
```

What do you think?

### Relationships
Closes #642

### Pull Request checklist

### Pointers
Test it out.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it


heytherewill (changes requested):
Do add tests to prevent us from making this mistake again

pdrj (approved):
👍

It'd be nice to have tests, as Will said.

daividssilverio:
I've also added the missing test for the sign up with 3rd party effect.

heytherewill (approved):

PR #834, opened 2020-09-09
⏳ Resolve outstanding push request

### LINKED ISSUE #770: 🦦 ResolveOutstandingPushRequest Sync Step
## Description
ResolveOutstandingPushRequest is the [first and fourth step in sync sequence ](https://github.com/toggl/mobileapp/blob/54bb2f1194b97fac59113222caaf390a4acf1581/Toggl.Core/Sync/v2/SyncManagerV2.cs#L191)

See [Xamarin code](https://github.com/toggl/mobileapp/blob/54bb2f1194b97fac59113222caaf390a4acf1581/Toggl.Core/Interactors/Sync/ResolveOutstandingPushRequestInteractor.cs#L31) for reference

## Dependencies
Everything connected to push result processing must be ready 

## Definition of done
- [ ] `OutstandingPush` request in our `SyncApi` and network clients is implemented
- [ ] OutstandingPush request is called
- [ ] OutstandingPush response is processed
- [ ] pushRequestIdentifier is cleared 
- [ ] Add tests if possible

simonrozsival:
### Summary
This is a port of the `ResolveOutstandingPushRequestInteractor` from the Xamarin app.

### Relationships
Closes #770 

### Pull Request checklist

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑


common-android/src/main/java/com/toggl/common/services/PushRequestIdStorage.kt
```diff
@@ -21,7 +21,8 @@ class PushRequestIdStorage @Inject constructor(
     fun set(id: UUID?) {
         sharedPreferences.edit { putString(pushRequestIdSharedPreferencesKey, id?.toString() ?: noPushRequestIdDefaultValue) }
     }
+
+    fun clear() = set(null)
+    fun hasOutstandingRequest() = get() != null
 }
```
heytherewill: I don't like this change. This basically means now we have to mock the behavior of `clear` and `hasOutstandingRequest`, when we could much easier have then configured by simply messing with the behavior of `set`

simonrozsival: I actually wanted to do this in order to be able to mock `hasOutstandingRequest`. I didn't like that I had to mock the dependency of `PushRequestIdStorage` in order to test how it is being used. That's one level of nesting too much :-/

I changed all the places where this needs to be mocked, it's was just about changing `set(null)` to `clear()` in `RepositoryTest.kt`. It doesn't require any additional changes.

Are we gonna use `PushRequestIdStorage` anywhere else except for `PushSync`, `ResolveOutstandingPushRequest`, and `Repository`? I don't think we will need it anywhere else 🤔

heytherewill (approved):
Looks good. I won't fight for the mocking thing because, as you said, it's quite simple, but in general I still prefer extensions for keeping the API sane

heytherewill:
One question, though: Should we not retry only a certain amount of times and then give up? 🤔

sync/src/test/java/com/toggl/sync/sequence/ResolveOutstandingPushRequestTest.kt
```diff
@@ -0,0 +1,156 @@
...
+import kotlinx.coroutines.test.runBlockingTest
+import org.junit.jupiter.api.DisplayName
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.assertDoesNotThrow
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.MethodSource
+import java.util.UUID
+import java.util.stream.Stream
+
+@DisplayName("The PushSync step")
```
semanticer, line 31: Probably © & 🍝 error?
```suggestion
@DisplayName("The ResolveOutstandingPushRequest step")
```

semanticer (approved):
Good 👍

sync/src/test/java/com/toggl/sync/sequence/ResolveOutstandingPushRequestTest.kt
simonrozsival: You're right, thanks!

simonrozsival:
@heytherewill Yep, we might do that. Let's finish the initial implementation and add that once we start testing it 👍 (please create an issue)

PR #835, opened 2020-09-09
Sync after login

### LINKED ISSUE #832: Create the sync screen
## Description
We need to show the view that awaits for sync to run before we log the user in

This view will be reused when we migrate users from the Xamarin app (since we will only have the api token and will need to re-fetch entities.

## Definition of done
- [ ] The view is added
- [ ] The login view navigates to this view instead of the main view
- [ ] The third party logins also go through the sync view 
- [ ] Sign ups do not go through the sync view

heytherewill:
### Summary
This adds a view that fetches user data right after we get the api token.

### Technical considerations
I had to increase the timeout limit on OkHttp because of how long pull sync takes when you have a bigger amount of entities.
I also had to change the way we inject the WorkManager because early initialization of it would cause dagger to fail

### Relationships
Closes #832

### Pull Request checklist

### Pointers
Go at it commit by commit. It will be a lot easier this way

### Review template
- [ ] The code considers i18n

## Merge permissions
Anyone, anytime


semanticer (approved):
I tried hard but couldn't find anything wrong with this. The sync fails because of `OptionalPropertyAdapterFactory` issues again but this should be already solved in Master right?

PR #838, opened 2020-09-09
✨ Add push notification handling (sync + tokens)

### LINKED ISSUE #768: Setup Push Notification Handling
## Description
The push notification should fire sync every time we receive a push notification.  

Optimally we don't want to launch sync when it's already running or a few seconds after it finished since the push notification might be just a response to our own sync. 

## Dependencies
#738 #767

## Definition of done
- [ ] Add necessary [dependencies](https://firebase.google.com/docs/cloud-messaging/android/client#add_firebase_sdks_to_your_app)
- [ ] Edit [manifest](https://firebase.google.com/docs/cloud-messaging/android/client#manifest)
- [ ] Use PushServicesApi(created in #767) to subscribe/unsubscribe etc with your token.  
- [ ] Setup [message receiving](https://firebase.google.com/docs/cloud-messaging/android/receive) and launch the sync via `SyncLauncher`
- [ ] (Optional) Experiment with some optimizations to minimize the syncing as mentioned in the description

pdrj:
### Summary
This adds push notifications support. It registers/de-registers the tokens and launches sync.

### Technical considerations
- In the onboarding module I've added a reducer that takes care of registering the token on logins. This reducer is sort of exploited to do stuff that probably doesn't belong to a reducer. The alternative would be to call token registration api in each place we do `setuser` - not sure what's better 🤷 
- I don't enqueue sync when a previous sync finished in last 10s - that is quite an arbitrary number, I'm happy to hear suggestions on what this number should be.
- I didn't add anything about remote config (which can enable/disable sync), as we have in xamarin, not sure if we need this
- I have not added tests yet, need to think what we want to actually test here

### Relationships

Closes #768 

### Pull Request checklist

### Pointers
- in :app build.gradle, comment out line 155
- get yourself the firebase json from the staging project in firebase, add your debug sha1 there too
- copy over the json into the app module
- clean the app module

Validate that:
- when new token is received by the service, there is an attempt to register it with the api (this will most likely fail as you won't be logged in)
- when logging in, the token is registered
- when logging out, the token is de-registered
- when logging in again, the token is registered again
- push some notifications to yourself: https://toggl.slack.com/archives/C02APMYTM/p1599569943386000?thread_ts=1599567270.376200&cid=C02APMYTM
- validate that the first notification schedules sync
- validate that subsequent ones don't enqueue sync in the 10s after the last sync finished
- validate that 10s after last sync finished, sync is enqueued again when push notification is received

### Review template

- [ ] The code considers i18n

## Merge permissions
Me


pdrj:
One more thing to mention is that the app now crashes without a proper firebase application id set. The dirty way to solve this is by everyone having a local copy of the debug section of firebase json inside our json template - it should get copied over to the generated json and local debug builds should work fine.

simonrozsival:
I'm not sure if it's better to test for 10s _after_ sync finishes, or for the previous _start_ of sync - if we get a PN while sync is already running, we probably shouldn't queue another sync (it was probably caused by the changes we pushed to the server).

We don't need remote config for PNs anymore, that was just for the rollout last year.

pdrj:
> if we get a PN while sync is already running, we probably shouldn't queue another sync

we have this already in SyncLauncher:
```
        val wasQueued = syncSequence.queueSyncIfAlreadyRunning()
        if (wasQueued) return
```

I added the check on finish time, because otherwise if the first sync took long to finish, we could end up with the second one executing right after 🤔 What do you think?

simonrozsival:
> I added the check on finish time, because otherwise if the first sync took long to finish, we could end up with the second one executing right after 🤔 What do you think?

Your reasoning is correct 👍 We should not unnecessarily sync right after we've just synced. That snippet you mention does something else than I had in mind though - if we're currently syncing, we'll _queue_ the intent to sync right after this current sync finishes, but we won't run a second sync process in parallel. Check what `SyncSequence` `sync` and `shouldKeepSyncing` do.

Let's keep the "throttling" for sync invocations for incoming PNs, but let's also add a different sync invocation method to `SyncLauncher`, something like `startSyncIfNotRunning`, or add a parameter to `enqueueSync(skipIfAlreadyRunning: Boolean = false)` so that PNs don't enqueue another sync loop.

daividssilverio:
@pdrj @simonrozsival 

We should also handle retrying to subscribe to the push services api in case it fails during login or during the `MessagingService:onNewToken`, right?

In the old implementation we had a [sync step just for that](https://github.com/toggl/mobileapp/blob/359e131d078af4a315a9fef455ed2712bc015363/Toggl.Core/Sync/States/Push/SyncPushNotificationsTokenState.cs).

Or will that be handled in another issue?

daividssilverio:
Hey @pdrj, about the pointers. I didn't have to add my debug sha1 to the staging firebase project.
> in :app build.gradle, comment out line 155
> get yourself the firebase json from the staging project in firebase, add your debug sha1 there too
> copy over the json into the app module
> clean the app module

All I had to do was:
- Get `google-services.json` from the staging firebase project
- Place it at `app/src/debug/google-services.json` (found this in the firebase docs samples)
- Clean the project

So, regarding your comment:
> One more thing to mention is that the app now crashes without a proper firebase application id set. The dirty way to solve this is by everyone having a local copy of the debug section of firebase json inside our json template - it should get copied over to the generated json and local debug builds should work fine.

Everybody having a local copy of the `google-services.json`  on `app/src/debug` doesn't sound too bad, our .gitignore already ignores that file anyway. We'd just have to add something to our README.

daividssilverio:
As for the tests:
- [x] when new token is received by the service, there is an attempt to register it with the api (this will most likely fail as you won't be logged in)
- [x] when logging in, the token is registered
- [x] when logging out, the token is de-registered
- [x] when logging in again, the token is registered again
- [x] push some notifications to yourself: https://toggl.slack.com/archives/C02APMYTM/p1599569943386000?thread_ts=1599567270.376200&cid=C02APMYTM
- [x] validate that the first notification schedules sync
- [ ] validate that subsequent ones don't enqueue sync in the 10s after the last sync finished
- [x] validate that 10s after last sync finished, sync is enqueued again when push notification is received

> - [ ] validate that subsequent ones don't enqueue sync in the 10s after the last sync finished

Was it expected that the sync manually triggered `enqueueSync`s from the `MainActivity` would run after a sync triggered by PNs is finished? Cause it seems to be running. Based on the firebase http logs.

daividssilverio (commented):
The code looks good to me.
One thing that is bothering me is the lack of logs.
Like the `CommonSsoReducer`, the `CommonTokensReducer` also handles actions that were already handled by some other reducer and it almost looks like those `SetUser` actions are doing too much stuff.

Maybe we should create specific actions that could be fired as effects of the `SetUser` actions, which would then fire specific actions that would be logged, maybe something like `MustLinkSso`, `PushNotificationSubscriptionFired`. Or we could add a custom logger service just for that.

I think the syncing could benefit from some logging as well...

pdrj:
> Let's keep the "throttling" for sync invocations for incoming PNs, but let's also add a different sync invocation method to SyncLauncher, something like startSyncIfNotRunning, or add a parameter to enqueueSync(skipIfAlreadyRunning: Boolean = false) so that PNs don't enqueue another sync loop. What do you think?

I did that in the most recent (https://github.com/toggl/android/pull/838/commits/b96fc88ebf5eb9b701095349e7796768e57fd744) commit @simonrozsival. Thanks for the explanation and suggestion.

> Maybe we should create specific actions that could be fired as effects of the SetUser actions, which would then fire specific actions that would be logged, maybe something like MustLinkSso, PushNotificationSubscriptionFired. Or we could add a custom logger service just for that.

That is not great, yeah. But on the other hand, having a proxy effect to just fire an action, might be too much boilerplate for the cause? 🤔 Maybe we should have a separate issue to think about that?

> Everybody having a local copy of the google-services.json on app/src/debug doesn't sound too bad, our .gitignore already ignores that file anyway. We'd just have to add something to our README.

Sounds good to me 👍 

> Was it expected that the sync manually triggered enqueueSyncs from the MainActivity would run after a sync triggered by PNs is finished? Cause it seems to be running. Based on the firebase http logs.

I now enqueue the push-triggered sync in a different way, but still the ones triggered manually from the main activity would just get enqueued.

daividssilverio:
>Maybe we should create specific actions that could be fired as effects of the SetUser actions, which would then fire specific actions that would be logged, maybe something like MustLinkSso, PushNotificationSubscriptionFired. Or we could add a custom logger service just for that.

>>That is not great, yeah. But on the other hand, having a proxy effect to just fire an action, might be too much boilerplate for the cause? 🤔 Maybe we should have a separate issue to think about that?

Another issue/a discussion sounds good to me.

sync/src/main/java/com/toggl/sync/SyncLauncher.kt
```diff
@@ -9,6 +9,11 @@ import com.toggl.sync.sequence.SyncSequence
 import javax.inject.Inject
 
 class SyncLauncher @Inject constructor(private val syncSequence: SyncSequence, private val workManager: WorkManager) {
+
+    fun startSyncIfNotRunning() {
+        if(syncSequence.isSyncAlreadyRunning()) enqueueSync()
```
simonrozsival, line 14:
```suggestion
        if(!syncSequence.isSyncAlreadyRunning()) enqueueSync()
```

pdrj:
> Or maybe there's a different and better way to run some code like this regularly (like once a day) using the work manager or something like that?

I've changed it so that now registering tokens is done via periodic work. We trigger this work on logins and when a new token arrives. Then it is scheduled to run daily. For currently logged in users the work won't be triggered (if the token doesn't change). We can think about having the work registered on each app launch for instance.

simonrozsival (approved):
Periodic work seems exectly like what I had in mind :+1:

common-android/src/main/java/com/toggl/common/services/tokens/FcmTokenSubscribeWorker.kt
```diff
@@ -0,0 +1,29 @@
...
+) : CoroutineWorker(context, parameters) {
+    override suspend fun doWork(): Result {
+        return try {
+            val previousToken = tokenStorage.getPreviouslyRegisteredToken()
+            val currentToken = tokenStorage.getCurrentToken()
+            if (previousToken == currentToken) return Result.success()
+
+            pushServicesApiClient.subscribe(currentToken)
+            tokenStorage.setPreviouslyRegisteredToken(currentToken)
+            Result.success()
```
daividssilverio, line 24: 🤔 I wonder if we shouldn't cancel the repeating work if we succeed registering the token.
Also, do you know if the worker will survive a device reboot?

daividssilverio: > Also, do you know if the worker will survive a device reboot?

Got my answer from the docs.

daividssilverio (changes requested):
> We can think about having the work registered on each app launch for instance.

I think we should add a check for that when the application launches. If not here, another issue is fine.

Now about the recurring work, I think we could cancel it when we succeed registering the token. Otherwise, LGTM.

common-android/src/main/java/com/toggl/common/services/tokens/FcmTokenSubscribeWorker.kt
simonrozsival: If we're doing one req per day, I believe we don't need to cancel the work and we can keep it just to be sure. One of the reasons we had the repeated request in place was to re-register the token if the server lost the token (e.g., the Redis that holds the tokens crashes and is restarted) so that the user doesn't stop getting the PNs all of a sudden.

pdrj: Ah, I didn't know about that scenario where server can lose the token. In this case, I'll remove (in the subscribe worker) the check comparing previously registered token and the one we want to save so that we always call the api.

daividssilverio: > Ah, I didn't know about that scenario where server can lose the token

I didn't know about that either.

pdrj:
@simonrozsival @daividssilverio I've removed the check comparing previously subscribed token to we always call the api. I've also made it so that the service (the local service) is called each time the TE list fragment resumes. That will cause the local service to be called quite a lot, but since the work request is going to be crated with a `KEEP` policy, no actual API calls will happen. We only call the API straight away if we receive a new token and on logins.

simonrozsival (approved):

daividssilverio (approved):
Works great. The only I would consider changing is moving the `        fcmTokenService.subscribeToken(executeImmediately = false)` call from the `TimeEntriesLogFragment` to the `MainActivity`.

pdrj:
> from the TimeEntriesLogFragment to the MainActivity.

The reason I didn't do it is because in MainActivity we might not be authenticated and the call will fail anyway.

daividssilverio:
> The reason I didn't do it is because in MainActivity we might not be authenticated and the call will fail anyway.

I'm alright with that for now. If the need comes, we can change it.

PR #839, opened 2020-09-10
Realm migration

### LINKED ISSUE #836: Migrate data from Realm
## Description
We need to make sure that users who transition from the Xamarin app into Aurora are eased in and barely notice it. Since we already have the API token stored in shared preferences, we can easily fetch it, sync the user data again into our database and let push sync do its job.

## Dependencies
We need the sync view from #832

## Definition of done
- [ ] Migrating the ApiToken to our new way of storing it
- [ ] Making sure the user won't be logged out
- [ ] Re-fetch entities as needed
- [ ] Reimport the unsynced entities that are in the unsynced data dump
- [ ] The import should only ever happen once

heytherewill:
### Summary
This adds that sweet sweet Realm migration so that existing users coming from the Xamarin app won't suffer :tm:

### Technical considerations
Everything here is temporary code, so I've laid things out in a way that makes the code easier to delete; There aren't many dependencies and we can pretty much just revert this commit as a mean of cleaning up the code added here (hopefully)

We first migrate the shared preferences and run a pull sync so that the database is properly populated with user data.
Then, we insert the entities from the data dump in the database and mark them as needing sync. There's a bit of logic to ensure the consistency of server/local ids there, so please pay attention to it
Once all that is done, the app behaves as if none of this has ever happened and the user is properly logged in

Feel free to suggest changes here. Also, we need the code to handle errors to act on sync (as in: retry the sync when it fails)

### Relationships
Closes #836

### Pull Request checklist

### Pointers
Review each commit individually, since I've separated things nicely.

For a functional test, do the following:

- Install the app and login to your account 
- Replace `TogglApplication.kt` with the following:

```kotlin
package com.toggl

import android.app.Application
import android.content.SharedPreferences
import androidx.core.content.edit
import androidx.hilt.work.HiltWorkerFactory
import androidx.work.Configuration
import com.toggl.api.ApiTokenProvider
import com.toggl.architecture.DispatcherProvider
import com.toggl.architecture.StoreScopeProvider
import com.toggl.architecture.core.Store
import com.toggl.domain.AppAction
import com.toggl.domain.AppState
import com.toggl.domain.loading.LoadingAction
import com.toggl.initializers.AppInitializers
import com.toggl.repository.Repository
import com.toggl.sync.migration.SharedPreferencesMigrator
import dagger.hilt.android.HiltAndroidApp
import kotlinx.coroutines.CoroutineScope
import kotlinx.coroutines.launch
import org.intellij.lang.annotations.Language
import java.io.File
import javax.inject.Inject
import kotlin.coroutines.CoroutineContext

@HiltAndroidApp
class TogglApplication : Application(), CoroutineScope, StoreScopeProvider, Configuration.Provider {

    @Inject lateinit var appInitializers: AppInitializers
    @Inject lateinit var dispatchersProviders: DispatcherProvider
    @Inject lateinit var store: Store<AppState, AppAction>
    @Inject lateinit var workerFactory: HiltWorkerFactory
    @Inject lateinit var sharedPreferencesMigrator: SharedPreferencesMigrator
    @Inject lateinit var sharedPreferences: SharedPreferences
    @Inject lateinit var repository: Repository

    override val coroutineContext: CoroutineContext by lazy {
        dispatchersProviders.main
    }

    override fun onCreate() {
        super.onCreate()
        appInitializers.initialize(this)

        // This entire region of code can be deleted once we
        // completely deprecate the Xamarin versions of our app.
        launch {
            setupUserThatNeedsMigration()
            if (sharedPreferencesMigrator.syncNeeded()) {
                sharedPreferencesMigrator.migrateApiToken()
                store.dispatch(AppAction.NavigateToSyncForRealmMigration)
            } else {
                store.dispatch(AppAction.Loading(LoadingAction.StartLoading))
            }
        }
    }

    override fun getStoreScope(): CoroutineScope =
        this

    override fun getWorkManagerConfiguration() =
        Configuration.Builder()
            .setWorkerFactory(workerFactory)
            .build()

    private suspend fun setupUserThatNeedsMigration() {
        val nope = "NO_TOKEN"
        val legacyApiTokenKey = "SharedStorageApiTokenKey"
        val token = sharedPreferences.getString(ApiTokenProvider.apiToken, nope)
        if (token != nope) {
            repository.clearAllData()

            setupFile()

            sharedPreferences.edit {
                putString(legacyApiTokenKey, token)
                putString(ApiTokenProvider.apiToken, nope)
                putBoolean("realm_migration_done", false)
            }
        }
    }

    private fun setupFile() {
        val file = File("${applicationContext.filesDir}/unsynced-migration-dump.json").outputStream()
        val bytes = serializedDump.toByteArray()
        file.write(bytes)
        file.close()
    }
}

@Language("JSON")
private const val serializedDump = """{YOUR_DUMP_FILE_HERE}"""
```
- Replace `{YOUR_DUMP_FILE_HERE}` with the contents of an actual dump file
- Now each time you run the app it will behave as if this is coming from an empty canvas and will run all the migration logic.

In order to get a valid dump file generated from Xamarin:

- Open the Xamarin app
- Login into the same account that you wanna test on Aurora
- Turn airplane mode once logged in
- Create a bunch of entities
- Go to the home screen (IOW, send to app to the background)

Once you've done that, a valid dump file exists. You can either get it using adb or by putting a breakpoint in line 46 of `UnsyncedDataPersistenceService.cs` once VS hits the breakpoint, you can get the file from the `serializedUnsyncedDataDump` variable.

### Review template

## Merge permissions
⚠️ ONLY MYSELF ⚠️


sync/src/main/java/com/toggl/sync/migration/RealmMigrator.kt
```diff
@@ -0,0 +1,170 @@
...
+
+            return
+        }
+
+        try {
+            database.withTransaction {
+                val syncDao = database.syncDao()
+
+                database.workspaceDao().getAllAccessible().onEach {
+                    Log.d("asd", it.toString())
```
daividssilverio, line 65: 👀

heytherewill: :eyes:

sync/src/main/java/com/toggl/sync/migration/RealmMigrator.kt
```diff
@@ -0,0 +1,170 @@
...
+        } catch (ex: Exception) {
+            // TODO: Log something here
+        }
+    }
+
+    private fun dataImportIsNeeded(): Boolean =
+        !sharedPreferences.getBoolean(realmMigrationAlreadyDoneKey, false)
+
+    private fun RealmTimeEntry.toDatabaseModel(localWorkspaceId: Long, localProjectId: Long?, localTagIds: Set<Long>) = DatabaseTimeEntry.from(
+        serverId = null,
```
daividssilverio, line 128: Are the `serverId`s supposed to be always be `null` for the time entries?
I know we don't make changes to projects, clients and tags, but unsynced time entries could already have a server Id, couldn't they? Wouldn't this cause new time entries to be created with the unsynced changes?

daividssilverio (commented):
Tested and it seems the migration seems to be working alright.
I'd like to see it working the completed push sync though. I left a question about the always null `serverId` for the time entries.

daividssilverio:
Btw, for other fellow reviewers:

```kotlin
private fun setupFile() {
        val file = File("${applicationContext.filesDir}/unsynced-migration-dump.json").outputStream()
        val input = assets.open("dump.json")
        val buffer = ByteArray(1024)
        var read: Int = input.read(buffer)
        while (read > 0) {
            file.write(buffer, 0, read)
            read = input.read(buffer)
        }
        input.close()
        file.close()
    }
```

Snippet to read a `dump.json` file from the assets folder in case your migration file is too big. 5k lines seems to be too much.

daividssilverio (approved):
LGTM.
Interesting thing I've noticed though. We are still not syncing up new time entries, right? We can only download them for now, right?

Not pertaining this PR, but Inspecting the database after the migration, I've noticed that a time entry I've created locally was marked with the syncStatus with `InSync`, while everything else was marked with `SyncNeeded`.

PR #841, opened 2020-09-10
Clean-up sync step

### LINKED ISSUE #769: 🧹 Cleanup Sync Step
## Description

The cleanup step is the last step from the [sync sequence.](https://github.com/toggl/mobileapp/blob/54bb2f1194b97fac59113222caaf390a4acf1581/Toggl.Core/Sync/v2/SyncManagerV2.cs#L195)
It should basically make sure we clean up all old or no longer accessible entities. Kinda like a good garbage collector does. See [Xamarin code](https://github.com/toggl/mobileapp/blob/54bb2f1194b97fac59113222caaf390a4acf1581/Toggl.Core/Interactors/Sync/CleanUpInteractor.cs#L30) here a summary of what it does:

1. [Delete time entries that are too old](https://github.com/toggl/mobileapp/blob/54bb2f1194b97fac59113222caaf390a4acf1581/Toggl.Core/Interactors/Sync/CleanUpInteractor.cs#L32)
2. [~~Delete placeholders~~](https://github.com/toggl/mobileapp/blob/54bb2f1194b97fac59113222caaf390a4acf1581/Toggl.Core/Interactors/Sync/CleanUpInteractor.cs#L33) this should no longer be needed. 
3. [Delete all inaccessible entities](https://github.com/toggl/mobileapp/blob/54bb2f1194b97fac59113222caaf390a4acf1581/Toggl.Core/Interactors/Sync/CleanUpInteractor.cs#L37) = entities that are deleted OR are referencing deleted workspace
4. [Delete those inaccessible workspaces](https://github.com/toggl/mobileapp/blob/54bb2f1194b97fac59113222caaf390a4acf1581/Toggl.Core/Interactors/Sync/CleanUpInteractor.cs#L42) (they shouldn't be referenced by any entity after step 3)

## Dependencies
None

## Definition of done
- [x] Cleanup step is deleting all old time entries
- [x] Cleanup step is deleting all no longer accesible entities
- [x] Cleanup step is deleting all inacesible workspaces
- [ ] Cleanup step is properly injected into sync sequence
- [x] Add tests if possible

simonrozsival:
### Summary
This is the final step to finish clean-up for syncing.

### Technical considerations
I somehow thought there will be more to do, so I didn't do it in #840, but I probably should have 😆 

### Relationships
Closes #769 

### Pull Request checklist

### Pointers
Did I froget to do something? I really thought there would be more to do 🤔 

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑


sync/src/main/java/com/toggl/sync/sequence/CleanUp.kt
```diff
@@ -1,10 +1,17 @@
...
+import com.toggl.database.dao.SyncDao
+import java.time.Duration
+import java.time.OffsetDateTime
 import javax.inject.Inject
 
-class CleanUp @Inject constructor() : SyncStep {
+class CleanUp @Inject constructor(
+    val syncDao: SyncDao
+) : SyncStep {
+    private val maximumTimeEntryAge = Duration.ofDays(8 * 7)
```
semanticer, line 11: Consider adding this to `com.toggl.common.Constants`

sync/src/main/java/com/toggl/sync/sequence/CleanUp.kt
```diff
@@ -1,10 +1,17 @@
...
+
     override suspend fun run() {
-        delay(2000)
+        val threshold = OffsetDateTime.now() - maximumTimeEntryAge
```
semanticer, line 14:
```suggestion
        val threshold = timeService.now() - maximumTimeEntryAge
```

semanticer (changes requested):
Consider these small suggestions 👇

sync/src/main/java/com/toggl/sync/sequence/CleanUp.kt
```diff
@@ -1,10 +1,17 @@
...
 
-import kotlinx.coroutines.delay
+import com.toggl.database.dao.SyncDao
+import java.time.Duration
+import java.time.OffsetDateTime
 import javax.inject.Inject
 
-class CleanUp @Inject constructor() : SyncStep {
+class CleanUp @Inject constructor(
+    val syncDao: SyncDao
```
semanticer, line 9:
```suggestion
    private val syncDao: SyncDao,
    private val timeService: TimeService
```

semanticer (changes requested):

semanticer (approved):

PR #845, opened 2020-09-11
🔄 Process push results

### LINKED ISSUE #779: Process Push Result
## Description
This piece of code should handle the processing of `PushResult`
[Xamarin code](https://github.com/toggl/mobileapp/blob/54bb2f1194b97fac59113222caaf390a4acf1581/Toggl.Storage.Realm/Queries/ProcessPushResultQuery.cs#L28) 

## Dependencies
#773 

## Definition of done
- [ ] Process Workspaces
- [ ] Process User
- [ ] Process Preferences
- [ ] Process Tags
- [ ] Process Clients
- [ ] Process Tasks
- [ ] Process Projects
- [ ] Process Time Entries
- [ ] All of the above are wrapped in db transaction
- [ ] Add tests

pdrj:
### Summary
Adds processing of the push results

### Technical considerations
This is a WIP since I lack the knowledge of how sync server is supposed to work and what can I expect there to be in the response. There are some casts that might not be safe, so I'll need someone more familiar with this stuff to tell me if I'm going the right way here.

I've tried to trigger an actual response from the server, but the best I can do is either to get a SyncError back, or a bad request 😄. I couldn't test it properly then.

It only covers workspaces for now, to get some validation that I'm doing the right thing.

### Relationships

Closes #779 

### Pull Request checklist

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


sync/src/main/java/com/toggl/sync/processing/ProcessPushedWorkspaces.kt
```diff
@@ -0,0 +1,55 @@
...
+    syncDao: SyncDao,
+    workspaceDao: WorkspaceDao,
+    workspaces: List<PushActionResult<SyncApiWorkspace>>
+) {
+    val errorCodeSkipped = 2L
+
+    for (apiWorkspace in workspaces) {
+        apiWorkspace.meta?.id?.let {
+            if (syncDao.getWorkspaceByServerId(it) == null) return
+        }
```
simonrozsival, line 24: This is kinda cryptic. I had to think for a while what's going on here. Let's move the logic into a separate function please (`if (doesNotExistLocally(action) continue`). Also should it really be a `return` and not a `continue`? 🤔

sync/src/main/java/com/toggl/sync/processing/ProcessPushedWorkspaces.kt
```diff
@@ -0,0 +1,55 @@
...
+
+        val localId = apiWorkspace.meta?.clientAssignedId?.id ?: return
```
simonrozsival, line 26: I think that this is wrong. If this was an `update` or `delete` action, you would have to find the `localId` using the server ID (`apiWorkspace.meta.id`). Here you would skip the results of all `delete` and `update` actions where `clientAssignedId` is not present in the response (the `MetaInfo` class is a "union" of all possible forms of the `"meta"` object values, it's a bit confusing..)

sync/src/main/java/com/toggl/sync/processing/ProcessPushedWorkspaces.kt
```diff
@@ -0,0 +1,55 @@
...
+import com.toggl.database.properties.StringSyncProperty
+
+internal suspend fun processPushedWorkspaces(
+    syncDao: SyncDao,
+    workspaceDao: WorkspaceDao,
+    workspaces: List<PushActionResult<SyncApiWorkspace>>
+) {
+    val errorCodeSkipped = 2L
+
+    for (apiWorkspace in workspaces) {
```
simonrozsival, line 21: I'd prefer if this was `actionResult` instead of `apiWorkspace`. The workspace itself should be in the `payload.result` for the `create` action results.

sync/src/main/java/com/toggl/sync/processing/ProcessPushedWorkspaces.kt
```diff
@@ -0,0 +1,55 @@
...
+                    syncDao.setWorkspaceSyncError(localId, payload.result.errorMessage.defaultMessage)
+                }
+            }
+            is ActionResult.EntitySuccessResult -> {
+                val result = payload.result!!
+                when (apiWorkspace.type) {
+                    ActionType.Create -> workspaceDao.updateWithoutFeatures(
+                        DatabaseWorkspaceWithoutFeatures(
+                            localId,
+                            (result.id as Some).value,
```
simonrozsival, line 40: This looks dangerous 👀

api/src/main/java/com/toggl/api/network/models/sync/PushResponse.kt
```diff
@@ -14,12 +14,12 @@ import com.toggl.api.models.SyncApiWorkspace
...
-    val user: ActionResult<SyncApiUser>,
-    val preferences: ActionResult<SyncApiPreferences>
+    val timeEntries: List<PushActionResult<SyncApiTimeEntry>>?,
+    val workspaces: List<PushActionResult<SyncApiWorkspace>>?,
+    val projects: List<PushActionResult<SyncApiProject>>?,
+    val clients: List<PushActionResult<SyncApiClient>>?,
+    val tasks: List<PushActionResult<SyncApiTask>>?,
+    val tags: List<PushActionResult<SyncApiTag>>?,
+    val user: ActionResult<SyncApiUser>?,
+    val preferences: ActionResult<SyncApiPreferences>?
```
simonrozsival, line 24: I think that this should actually contain `PushActionResult<ApiTimeEntry>` and not the `SyncApiTimeEntry` with the optional properties (which are intended for partial serialization of the push request body). The same would apply for the rest of the entity types, it's not just TEs. We wouldn't have to do the weird `(result.id as Some).value` cast to get the deserialized value.

sync/src/main/java/com/toggl/sync/processing/ProcessPushedWorkspaces.kt
```diff
@@ -0,0 +1,55 @@
...
+            is ActionResult.ErrorResult -> {
+                if (payload.result.errorMessage.code != errorCodeSkipped) {
+                    syncDao.setWorkspaceSyncError(localId, payload.result.errorMessage.defaultMessage)
+                }
+            }
+            is ActionResult.EntitySuccessResult -> {
+                val result = payload.result!!
+                when (apiWorkspace.type) {
+                    ActionType.Create -> workspaceDao.updateWithoutFeatures(
+                        DatabaseWorkspaceWithoutFeatures(
```
simonrozsival, line 38: This could be extracted into an extension function.

database/src/main/java/com/toggl/database/dao/SyncDao.kt
```diff
@@ -67,6 +67,16 @@ abstract class SyncDao {
     @Query("UPDATE time_entries SET syncStatus = :status WHERE id IN(:localIds)")
     abstract suspend fun changeTimeEntriesSyncStatus(localIds: List<Long>, status: EntitySyncStatus)
 
+    @Query("UPDATE workspaces SET lastSyncErrorMessage = :message, syncStatus = :status WHERE id = :localId")
+    abstract suspend fun setWorkspaceSyncError(localId: Long, message: String, status: EntitySyncStatus = EntitySyncStatus.SyncFailed)
+
+    @Query("UPDATE workspaces SET syncStatus = :syncStatusInSync WHERE id = :id AND syncStatus != :syncStatusSyncNeeded")
+    abstract suspend fun setSynced(
```
simonrozsival, line 74: This should be either called `setWorkspaceSynced` or something like that, or it should be in the `WorkspaceDao`

simonrozsival (changes requested):

sync/src/main/java/com/toggl/sync/processing/ProcessPushedWorkspaces.kt
```diff
@@ -0,0 +1,55 @@
...
+
+        val localId = apiWorkspace.meta?.clientAssignedId?.id ?: return
+
+        when (val payload = apiWorkspace.payload) {
+            is ActionResult.ErrorResult -> {
+                if (payload.result.errorMessage.code != errorCodeSkipped) {
+                    syncDao.setWorkspaceSyncError(localId, payload.result.errorMessage.defaultMessage)
+                }
+            }
+            is ActionResult.EntitySuccessResult -> {
```
simonrozsival, line 34: You're missing one more case here - `ActionResult.SuccessResult` - which is what you expect to get for successful `Update` and `Delete` actions (which have no payload when processing was successful).

simonrozsival:
@pdrj I made quite a few changes and I believe the implementation is almost complete but we still need to add a bunch of unit tests. I will work on that on Monday.

simonrozsival:
We should consider merging `PushSyncDao<T>` (which I added) and `EntityDao<T>` which was added by a different PR.

heytherewill (approved):

semanticer (approved):
I had the whole list of suggestions ready to throw at you when I realized that I'm looking at the old version of this PR and you had already made all the changes I wanted to suggest.  🤷

pdrj (commented):
I can't approve it, as I created it, but since pretty much all my WIP stuff is gone, I gave it a test and it is looking good 👍

simonrozsival (approved):

PR #849, opened 2020-09-14
📏 Improve the way we handle insets

heytherewill:
### Summary
This improves on the way we are handling our inset changes. Atm, we are using callbacks and discarding it where unneeded. After the changes, the insets will behave more like themes, which makes them accessible wherever we want and makes passing them around a lot nicer.

### Pointers
Do a functional test and make sure everything still looks good

### Review template
- [ ] The code considers i18n

## Merge permissions
Only myself


semanticer (approved):
It looks and works great. 

I tested this on API 30 with and without gesture navigation and on 21 and everything looks fine.
We have some issues with the nav and statusbar color on 21 tho but that's unrelated.
Screen Shot 2020-09-15 at 10 46 10
Screen Shot 2020-09-15 at 10 44 41

PR #879, opened 2020-09-16
📬 Add User subscription

### LINKED ISSUE #736: Create User Subscription
## Description
We need `LoadUserSubscription` that will update our state with the new user anytime there is a change

## Dependencies
None

## Definition of done
- [ ] Modify repository and dao if needed to provide user as a flow
- [ ] Implement `LoadUserSubscription`
- [ ] 🔥 Old update logic 
- [ ] Add tests

heytherewill:
### Summary
This adds the user subscription

##### Changing the way we handle the login 
We now only store the user's api token instead of persisting the user object. Why?
- Persisting the user object forces us to create a mock WS to ensure database consistency
- If the user interrupts login right after the user object is persisted, the app remains in an unusable state
- Sync already *has* to run anyway after logging in
- Login cancelling was a major problem back in the day because we had a partially dirty database. With this, cancelling login is as easy as not persisting/deleting a single key in shared prefs.

By persisting only the api token, we ensure that users whose first sync fail will remain logged in but will only ever use the app when we manage to properly fetch their entities, leaving them always in a valid state. I've tested this and we will likely not encounter this scenario very often because of background sync (it works really well even if you remove the app from the activity manager!), but it's always better to err on the side of caution

##### Removes other ways of setting the user
Since that's what subscriptions do, we no longer need to update the user in any other place

##### Changes in navigation
We no longer need an action to trigger the `StartLoading` action for initial navigation, since the setting of an user comes from the subscription. This means the migration from realm logic is now in a clearer place until we need to remove it!
We need to test the realm integration again just in case, but we should be fine

I've also changed the routing from sign ups to navigate to sync. I realized that we need it because even on newly created accounts we need two entities: The Workspace and the User. This also means that now all authentication goes to the same screen, which adds some nice consistency and gives us some insurance in case some behavior changes for newly created accounts!

Feel free to point some better ways of handling things; I'm all ears

### Relationships
Closes #736 

### Pull Request checklist

### Pointers
Review commit by commit for great justice

### Review template

## Merge permissions
Only myself!


app/src/test/java/com/toggl/domain/reducers/LoadingReducerTests.kt
```diff
@@ -87,40 +70,86 @@ class LoadingReducerTests : CoroutineTest() {
...
+            private val user = User(
+                id = 0,
+                apiToken = ApiToken.from("12345678901234567890123456789012") as ApiToken.Valid,
+                defaultWorkspaceId = 1,
+                email = Email.from("validemail@toggl.com") as Email.Valid,
+                name = "name",
+                timezone = "utc/UTC"
+            )
+
             fun `sets the current route to the timer page`() {
```
daividssilverio, line 82: It looks like we forgot to add `@Test` to every single test in this class.

api/src/main/java/com/toggl/api/ApiTokenProvider.kt
```diff
@@ -4,8 +4,9 @@ import com.toggl.models.validation.ApiToken
 
 interface ApiTokenProvider {
     fun getApiToken(): ApiToken
+    fun setApiToken(apiToken: ApiToken)
```
daividssilverio, line 7: This is more of a `ApiTokenStorage` now.

daividssilverio (approved):
Really liked the simplification. Tested and working fine, I've tested the migration again and it still works fine. I didn't test all the ways to authenticate though.

I also made an adhoc, probably invalid test:
I've made so that the workspaces processing part of the sync manager always fails to see what would happen. To my surprise, the initial sync actually finishes alright and the app navigates to the `Timer` route, except that the app is in an "blank state" and its unresponsive. After removing the failure and running the app again everything just worked, which was nice.

I think we should have some way to enter an error state when this happens. I'm saying that just for the initial sync (some way to try again). I'm assuming we'll be adding specific error handling for the "user without workspaces" case at some point based on the TODO in the sync code.

heytherewill:
Error handling is part of #722. As part of that issue we should get the error in the sync view and retry rather than emitting a completion! But yeah, this is outside the scope of this PR

PR #881, opened 2020-09-18
🔨 Respect 24hours user prefs & fix Settings toggle dispatcher

### LINKED ISSUE #860: Dispatch the setting's toggle/switch actions
We just need to bind the actions.
(They are currently read-only)

### LINKED ISSUE #875: Consider 12/24 hour mode user preferences in the calendar
We are currently using a static 24 hours mode.

daividssilverio:
### Summary
This PR does two things:
- Makes the calendar hour indicators respect the 24 hour clock user preferences.
- Makes the settings toggle dispatch the tap action when it's being clicked

### Technical considerations
#860 was sort of working already, but you had to click on the row (not the switch) to trigger the toggle switch. So I've made a click on the toggle itself dispatch a tap action.
It looks like the switch implementation was broken before compose alpha-03, and the `onCheckedChange` was being invoked when it shouldn't. If you the code here without updating to compose alpha-03 (#880), the switches will enter into feedback loop.

🤷‍♂️
I've fixed #860 because it became easier to test the calendar then.

### Relationships
Closes #875 
Closes #860 
Depends on #880 

### Pull Request checklist

### Pointers
Do a functional test, ideally after #880 is merged. Or update the compose

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it.


calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarBackgroundDrawingDelegate.kt
```diff
@@ -13,13 +13,23 @@ class CalendarBackgroundDrawingDelegate(
...
     private val verticalLineLeftMargin: Float,
     private val timeSliceStartX: Float,
-    private val hoursX: Float
+    private val hoursX: Float,
+    use24HoursFormat: Boolean,
 ) {
-    private val format = DateTimeFormatter.ofPattern("HH:mm")
+    private val format: DateTimeFormatter
+        get() = if (use24HoursFormat) DateTimeFormatter.ofPattern("HH:mm")
+        else DateTimeFormatter.ofPattern("h a")
```
heytherewill, line 21: Is this the only place where we wanna distinguish between 24 and 12 hour format? If so, that's fine. Otherwise, extract the pattern for consistency across the board ™️

heytherewill (approved):

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarBackgroundDrawingDelegate.kt
daividssilverio: This exact pattern is only used here. We also use the 24hours preferences in the edit view and the calendar contextual menu, but with different patterns. Putting these patterns in together in some file wouldn't hurt I guess. We used to have a `TimeFormat` in the old implementation, we could have a constants file to keep things organized and help making changes when needed.

heytherewill: If we both see value in that, let's implement it :)

daividssilverio: 🤔 what do you think of the `TimeFormats` helper?

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarItemDrawingDelegate.kt
```diff
@@ -84,7 +83,7 @@ class CalendarItemDrawingDelegate(
         color = editingHoursLabelsTextColor
         textSize = editingHoursLabelsTextSize
     }
-    private val hourFormat = DateTimeFormatter.ofPattern("HH:mm")
```
heytherewill, line 87: Wait, why is this hardcoded to twenty four hours?

common-android/src/main/java/com/toggl/common/extensions/OffsetDateTimeExtensions.kt
```diff
@@ -10,7 +11,7 @@ import java.time.temporal.ChronoUnit
 import java.time.temporal.TemporalUnit
 import kotlin.math.absoluteValue
 
-private val timeFormatter = DateTimeFormatter.ofPattern("HH:mm")
+private val timeFormatter = TimeFormats.TwentyFourHoursFormat.hoursAndMinutes
```
heytherewill, line 14: Same here: Why is it constrained to `TwentyFourHoursFormat` instead of respecting the user's settings?

calendar/src/main/java/com/toggl/calendar/calendarday/ui/views/CalendarItemDrawingDelegate.kt
daividssilverio: Leftovers from when we didn't have the user preferences available. Fixing it.

common-android/src/main/java/com/toggl/common/extensions/OffsetDateTimeExtensions.kt
daividssilverio: Same reason I suppose. Fixing it as well!

PR #885, opened 2020-09-21
🐛 Fix pull sync deserialization issues

### LINKED ISSUE #882: Fix deserialization issues for Pull sync
## Description
Right now Pull sync is failing because when a user loses access to an entity, all data but the server deleted stamp is omitted. This means Moshi fails to deserialize our entities, since their fields are all non-nullable

## Definition of done
- [ ] Pulling doesn't fail when deleted entities are included
- [ ] The names of the pull/push entities are changed to better reflect their true intentions

heytherewill:
### Summary
This considers invalid access when receiving deleted entities from the API and handles it accordingly

### Technical considerations
I've renamed the entities to make their usage more straightforward and easier to reason about. 

### Relationships
Closes #882

### Pull Request checklist

### Pointers
Review commit by commit

### Review template

## Merge permissions
Anyone, anytime


simonrozsival (approved):
LGTM

PR #893, opened 2020-09-21
🛰️ Launch sync correctly

### LINKED ISSUE #883: Mark entities as needing sync
## Description
This is one of the final things we need in order to have the app properly syncing!

## Definition of done
- [ ] Whenever a new entity is added to the app, it should be marked as `SyncNeeded`
- [ ] Tests ensure this is happening

### LINKED ISSUE #884: Sync when appropriate
## Description
We need to sync 

## Dependencies
Depends on #883

## Definition of done
- [ ] The trigger to sync/freeze on the calendar/reports tabs is removed
- [ ] Sync triggers whenever a new entity is created
- [ ] Pull to sync is added to the app

heytherewill:
### Summary
This marks our entities as needing sync and triggers sync in all correct places

### Technical considerations
I'm exposing the current sync step in the state so anyone can listen to it! Other than that, there's nothing too fancy here

### Relationships
Closes #883 
Closes #884 

### Pull Request checklist

### Pointers
Review commit by commit

### Review template
- [ ] The code considers i18n

## Merge permissions
Only me


semanticer (commented):
Looks good, the only thing that I don't exactly love is the fact that we have to call `syncController.launch()` manually and remember where it should be called. I was originally thinking we could add a subscription that would observe database sync states and trigger sync anytime there is a SyncNeeded state detected. But considering we are only calling `syncController.launch()` on a handful of places I am probably just overthinking this 🤷

simonrozsival (approved):
Seems OK. I'm couldn't 100% check that those are all the places where we should trigger sync, but I guess we would catch that later if we missed something here.

PR #896, opened 2020-09-22
📁 Fix project creation flow

### LINKED ISSUE #866: Fix "close" button in the New Project view
It's not doing anything at the moment

### LINKED ISSUE #867: Validate project creation in the New Project view.
Currently it's possible to create projects with empty names.

Add tests to prevent us from doing this again.

heytherewill:
### Summary
This fixes the close button and the validation for the project card

### Technical considerations
I've followed the designs (roughly, polishing comes in the polish project PR) and the behavior of the xamarin app, which is still the same.

I hate not using Compose, but rewriting these views right now is not a viable option, so let's bite the bullet, make these views amazing as they are so we can make them pleasant to work with in the future!

### Relationships
Closes #866
Closes #867 

### Pull Request checklist

### Pointers
Review commit by commit

### Review template

## Merge permissions
Anyone, anytime


daividssilverio (approved):
LGTM.

PR #771, opened 2020-08-20
Will/pull sync step

heytherewill:
### Summary
This implements the basic pull sync steps and tests and adds the stub for the processor

### Technical considerations
This is looking surprisingly good! I was expecting it to look awful, but it's actually quite good 👏 

### Relationships
Closes #740 
Closes #741

### Pull Request checklist
- [X] The code does not hinder the app's performance
- [X] The code does not introduce regressions
- [X] The code in this PR is tested thoroughly enough that changing the behavior to an incorrect one breaks tests

### Pointers
Go at it commit by commit, please

### Review template
- [ ] The code is thoroughly tested in well written and resilient manner
- [ ] The code is performant and leads to an optimal user experience
- [ ] The code is testable and does not cut corners for the sake of delivering
- [ ] The code considers older platforms well and will work in older devices just like it will in new ones

## Merge permissions
Anyone, anytime

Issue #741, opened 2020-08-13
Implement basic pull request processing [C#ProcessPullResultQuery]


semanticer:
## Description
PullRequestProcessor should take out all the different entities from `PullResponse` and distribute them to entity-specific processing methods.

Take a look at [ProcessPullResultQuery](https://github.com/toggl/mobileapp/blob/8779617e49c7da8c9899d36061784dcbb13e4ebf/Toggl.Storage.Realm/Queries/ProcessPullResultQuery.cs)
and [WriteUsingThreeWayMerge method](https://github.com/toggl/mobileapp/blob/8779617e49c7da8c9899d36061784dcbb13e4ebf/Toggl.Storage.Realm/Queries/WritePulledData.cs#L19) 
## Dependencies
#740

## Definition of done
- [ ] Create PullRequestProcessor
- [ ] Pass reponse **user** object into empty implementation of `processUser()` method
- [ ] Pass reponse **preferences** object into empty implementation of `processPreferences()` method
- [ ] Pass reponse **tag** object into empty implementation of `processTags()` method
- [ ] Pass reponse **project** object into empty implementation of `processProject()` method
- [ ] Pass reponse **task** object into empty implementation of `processTask()` method
- [ ] Pass reponse **client** object into empty implementation of `processClient()` method
- [ ] Pass reponse **workspaceFeatures** object into empty implementation of `processWorkspaceFeatres()` method
- [ ] Pass reponse **workspace** object into empty implementation of `processWorkspace()` method
- [ ] Pass reponse **timeEntries** object into empty implementation of `processTimeEntries()` method

If you can come up with some clever abstractions over these methods by all means feel free to do it!

PR #899, opened 2020-09-22
Pull and reset

simonrozsival:
### Summary
This PR handles the case when the server rejects the `since` parameter we pass to the `/pull?since<since>` URL. That happens if the `since` is too old (older than 2 or 3 months, I'm not sure now). When that happens, we need to fetch all the currently accessible data (we use the `/pull` endpoint without the `since` parameter), and detect losing access to any now inaccessible workspace and remove the related entities. It's surprisingly easy to implement this.

I followed https://github.com/toggl/mobileapp/blob/develop/Toggl.Storage.Realm/Queries/ResetLocalStateQuery.cs and I improved it a bit. The pruning in the old app wasn't optimal (it could screw up relationships between entities) and so I managed to reuse the clean-up code in the `SyncDao`.

### Relationships
It seems we don't have an issue for this. This is a follow-up of #771 and it's related to #741.

### Pull Request checklist

## Reviewing
This code would deserve several unit tests. I will try to add some today, but I'm not sure if I'll manage to do that.

### Review template

- [ ] The code considers i18n

## Merge permissions
~Please don't merge this PR today just yet (22.9.), I'll try to add a few unit tests during the afternoon. After the unit tests are added or I didn't manage to do that before I leave on my short vacation, please merge it~ 😄 It's ready now 🦑


heytherewill (approved):

semanticer (approved):
LGTM

Issue #863, opened 2020-09-15
Add the "previously tracked time entries" header to the time entries suggestions popup.


daividssilverio:
Full behaviour not yet defined.

We might implement some collapsing behaviour for this kind of suggestion.

daividssilverio:
@heytherewill what should we do for now?
The simplest behaviour is to hide the suggestions for the current StartEdit session (maybe except by `@` and `#` starting queries).
Or we can add a toggle behaviour as well, clicking "x (for now)" could hide and show the time entries suggestions.

PR #906, opened 2020-09-23
🚧🎨 Update time entry & project suggestions designs

### LINKED ISSUE #871: Display the different headers when making project suggestions
Check the designs, but we have different headers when:
- There are no projects available
- The current search query doesn't match any existing project
- The current search query matches exactly the name of an existing project (Assuming how our model works, this should only appear if every workspace the user has access already contains a project exactly matching the search query).

### LINKED ISSUE #873: In the StartEdit view display different headers in the tags suggestion's popup when
- There are no available tags yet
- The current search query doesn't match any existing tag
- The current search query matches exactly the name of an existing tag in the current workspace (either from a selected project or the default workspace).

Does not depend directly on #871 but I'm assuming whatever is done there can be reused here.

### LINKED ISSUE #888: Final designs: suggestions
## Description
The various designs can be found in https://toggl.invisionapp.com/overview/-Android--Rebranding-ckem8stapzb0o01wrr0c27nd1/screens/21197403

There are some 'x' buttons we're missing, some horizontal lines and other cosmetic stuff. Bumping up the estimate because we had many issues with these popups in the past.

## Definition of done

- [ ] Suggestions popups look as in the linked designs (for time entries, projects, clients etc.)

daividssilverio:
### Summary
This PR adds some missing details in the suggestions designs and adds the missing headers when creating tags and projects.

### Technical considerations
I didn't make the headers fixed. Should they be?
I wish we could use compose here.

Fixed headers sound like trouble with our popup window. 

### Relationships
Closes #888
Closes #871 
Closes #873 

It should help dealing with #863 

### Pull Request checklist

### Pointers
Check the tests.
Do a functional test.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it ~after the draft is done. The UI still needs final touches.~


heytherewill (approved):
Really solid solution! Tested both on modern and older versions of Android and had no issues what-so-ever. I'm fine with non-fixed headers for now, we can reiterate on that once we Compose-fy all things :)

PR #907, opened 2020-09-23
Time entries log overhaul

### LINKED ISSUE #862: Display the missing information in the running time entry card
We need to:
- Display project/task/client information in the running time entry card
- Display the running time entry duration

### LINKED ISSUE #886: Final designs: timer screen
## Description
Designs start here:

https://toggl.invisionapp.com/console/%5BAndroid%5D-Rebranding-ckem8stapzb0o01wrr0c27nd1/ckeriu1sh02as01xkbrf3dcf5/play

Some points to keep in mind:
- stop button on running TE (the square is larger in designs)
- elapsed time, billable, project, tag badges shown on the running TE tab
- different colors for the TE cells
- different copy for suggestions
- margins (for sure we have too much whitespace between 'your activity' and the first date badge)
- colored dots when TE has a project
- error states on TEs (sync, etc)
- swipe behavior 

## Definition of done
- [ ] the timer screen looks as designed (dark and light theme), with a running TE and without

### LINKED ISSUE #892: Final designs: bottom nav
## Description
https://toggl.invisionapp.com/console/%5BAndroid%5D-Rebranding-ckem8stapzb0o01wrr0c27nd1/ckeriu1sh02aw01xk3ziubmky/play

We don't show labels and underline items when the are selected. The no labels part should be easy with         `app:labelVisibilityMode`. If the underlining proves too complicated, let's first reach out to Milad to see if we can just go with highlighting the whole icon.

## Definition of done

- [ ] bottom navigation looks as designed (no labels and items get underlined when selected)

heytherewill:
### Summary
This PR applies that very welcome coat of paint in the time entries log so that they are aligned with our current designs!

### Technical considerations
I needed to add a property yo check whether a time entry is part of a group of not. That was really easy to do thanks to our architecture!
The state handling doesn't look as nice, but that's because I'm biased towards Compose; with a compose state of mind, every view holder will look disgusting

Also, you'll notice that for the suggestion/time entry cards I'm not using the card background, but painting the ConstraintLayout instead. This is to prevent material design from changing the color of the cards when in dark mode while keeping the shadow in light mode (that's the default way of applying elevation in dark mode)

I need you guys to take a look at the solution to make the cards stack. I limited to max 3 cards because otherwise it would simply go on forever :joy:

### Relationships
Closes #862
Closes #886
Closes #892

### Pull Request checklist

### Pointers
The commits won't be that helpful since a lot of the changes are cosmetic. Instead, check the bits that have some domain logic and ensure those are correct. Then, check the ViewHolders to see if there are any crimes against humanity there. If all that is well, do a proper functional test while looking at the final designs. I recommend testing using a proper account with actual data.

The review template below is actually quite useful and I recommend using it. Ah: And don't forget dark mode :sogood:

### Review template
- [ ] The code considers i18n


pdrj (commented):
Screenshot 2020-09-24 at 16 20 10

I've only just started reviewing this and noticed that we don't have the billable/project icons shown on the running tab like here https://toggl.invisionapp.com/console/%5BAndroid%5D-Rebranding-ckem8stapzb0o01wrr0c27nd1/ckeriu1sh02aw01xk3ziubmky/play#project_console

Also, the running tab is pinkish instead of white. 

I'll continue with the feedback tomorrow. :)

pdrj (changes requested):
Visual comments:

1. The dot is not aligned with text Screenshot 2020-09-25 at 10 10 02
2. The bottom bar is expanding with long descriptions
Screenshot 2020-09-25 at 10 15 25
3. Suggestions header says 'Track again' in designs.
4. The 'Play icon' has sharp edges in the designs, we have sort of soft edges
5. 'Im working on' background is white in designs.
6. Description of TEs is in bold in designs
7. In dark mode, the settings icon is white in designs.
8. (From my previous comment) Background of running TE in light mode 
9. 8. (From my previous comment) Billable/project icons on running TE
10. Spacings are inconsistent in the designs, but do you think we should add just a little more room between TE description and project lines in log items?

pdrj (changes requested):
Functional test:

1. Can't edit singular TEs (can edit groups only, but not singular items in an expanded group)
2. Can't do swipe actions over groups (can swipe over singular TEs only). Can't do swipe actions over items of an expanded group either.
3. Leakcanary is reporting a leak of RunningTimeEntryFragment instance (via updateDurationTickJob) - the job is not cancelled I think.

timer/src/main/java/com/toggl/timer/log/ui/TimeEntryItemViewHolder.kt
```diff
@@ -1,23 +1,113 @@
...
 
 class TimeEntryItemViewHolder(
     itemView: View,
     private val onTappedListener: (Long) -> Unit,
     private val onContinueTappedListener: (Long) -> Unit
 ) : TimeEntryContentViewHolder<FlatTimeEntryViewModel>(itemView) {
 
+    var isInClosedGroup = false
+
+    private val oneDpElevation = itemView.context.resources.getDimension(R.dimen.base_elevation)
```
pdrj, line 19: We can DRY it up a bit by moving `oneDpElevation` to the parent.

timer/src/main/java/com/toggl/timer/running/ui/RunningTimeEntryFragment.kt
```diff
@@ -39,55 +50,81 @@ class RunningTimeEntryFragment : Fragment(R.layout.fragment_running_time_entry)
...
-            runningTimeEntryFlow
-                .map { it != null }
-                .distinctUntilChanged()
-                .onEach { setEditedTimeEntryState(it) }
-                .launchIn(this)
-        }
+        store.select(selector)
+            .distinctUntilChanged()
+            .onEach { updateRunningTimeEntryCard(it) }
+            .launchIn(lifecycleScope)
```
pdrj, line 68: Let's use `viewLifecycleOwner.lifecycleScope` to be safe.

timer/src/main/res/layout/fragment_running_time_entry.xml
```diff
@@ -4,53 +4,133 @@
...
+    <ImageView
+        android:id="@+id/arrow"
+        app:tint="@color/primary"
+        android:src="@drawable/arrow_right"
+        android:layout_marginStart="@dimen/grid_2"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent"
+        app:layout_constraintBottom_toBottomOf="parent"
+        android:layout_width="24dp"
+        android:layout_height="24dp" />
```
pdrj, line 18: Let's extract every hardcoded dimen (that makes sense to be reused) in this file.

timer/src/main/res/layout/fragment_running_time_entry.xml
```diff
@@ -4,53 +4,133 @@
...
+        android:layout_marginBottom="20dp"
+        android:layout_width="8dp"
+        android:layout_height="8dp" />
+
+    <com.google.android.material.textview.MaterialTextView
+        android:id="@+id/running_time_entry_project"
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
+        android:layout_gravity="center_vertical"
+        android:fontFamily="sans-serif-medium"
```
pdrj, line 89: Probably out of scope of this PR, but we're redefining these text styles all over timer module. Would be nice to have a style.

timer/src/main/res/layout/fragment_time_entries_log.xml
```diff
@@ -17,6 +17,7 @@
             android:layout_width="match_parent"
             android:layout_height="match_parent"
             android:clipToPadding="false"
+            android:paddingBottom="120dp"
```
pdrj, line 20: Where did the `120` come from? Should it be a dimen dependent on some other dimens?

timer/src/main/res/layout/time_entries_log_header.xml
```diff
@@ -2,28 +2,29 @@
 <FrameLayout
     xmlns:tools="http://schemas.android.com/tools"
     xmlns:android="http://schemas.android.com/apk/res/android"
-    android:layout_height="@dimen/time_entry_log_item_height"
+    android:layout_height="36dp"
```
pdrj, line 5: Again would be nice to not have hardcoded `dps`

timer/src/main/res/layout/time_entries_log_suggestion.xml
```diff
@@ -1,82 +1,96 @@
...
-        android:layout_width="24dp"
-        android:layout_height="24dp"
-        android:layout_marginEnd="12dp"
-        android:src="@drawable/ic_play"
-        app:layout_constraintBottom_toBottomOf="parent"
-        app:layout_constraintEnd_toEndOf="parent"
-        app:layout_constraintTop_toTopOf="parent" />
-</androidx.constraintlayout.widget.ConstraintLayout>
\ No newline at end of file
+    <androidx.constraintlayout.widget.ConstraintLayout
```
pdrj, line 13: Please run a reformat on this file so we start in a clean state.

timer/src/main/res/layout/time_entries_log_suggestion.xml
```diff
@@ -1,82 +1,96 @@
...
+            app:layout_constraintStart_toEndOf="@id/project_color_indicator"
+            tools:text="Short proj" />
+        <View
+            android:id="@+id/duration_gradient"
+            app:layout_constraintTop_toTopOf="parent"
+            app:layout_constraintBottom_toBottomOf="parent"
+            android:background="@drawable/gradient_transparent_to_surface"
+            app:layout_constraintEnd_toStartOf="@id/continue_icon"
+            android:layout_height="0dp"
+            android:layout_width="48dp" />
```
pdrj, line 84: Again, let's extract `dimens` that make sense.

timer/src/main/java/com/toggl/timer/log/domain/TimeEntriesLogSelector.kt
```diff
@@ -66,29 +85,34 @@ class TimeEntriesLogSelector(
...
                                 val groupId = timeEntryGroup.first().similarityHashCode()
-                                if (expandedGroupIds.contains(groupId)) {
-                                    yieldTimeEntryGroup(true, timeEntryGroup)
-                                    for (timeEntry in timeEntryGroup) {
-                                        yieldFlatTimeEntry(timeEntry)
-                                    }
-                                } else {
-                                    yieldTimeEntryGroup(false, timeEntryGroup)
+                                val groupIsExpanded = expandedGroupIds.contains(groupId)
+                                val groupSize = if (groupIsExpanded) timeEntryGroup.size else min(timeEntryGroup.size, 3)
```
pdrj, line 101: Let's name the constant, it is quite invisible hardcoded like that ;)

timer/src/main/java/com/toggl/timer/log/ui/SwipeActionCallback.kt
```diff
@@ -30,10 +29,14 @@ class SwipeActionCallback(
     override fun getSwipeDirs(
         recyclerView: RecyclerView,
         viewHolder: RecyclerView.ViewHolder
-    ): Int = when (viewHolder) {
-        is TimeEntryLogViewHolder -> ItemTouchHelper.LEFT.or(ItemTouchHelper.RIGHT)
-        else -> ItemTouchHelper.ACTION_STATE_IDLE
-    }
+    ): Int =
+        when (viewHolder) {
+            is TimeEntryGroupViewHolder -> ItemTouchHelper.LEFT.or(ItemTouchHelper.RIGHT)
```
pdrj, line 34: It reads nicer to me in full infix style (here and elsewhere in this file)
```suggestion
            is TimeEntryGroupViewHolder -> ItemTouchHelper.LEFT or ItemTouchHelper.RIGHT
```

timer/src/main/java/com/toggl/timer/log/ui/TimeEntryContentViewHolder.kt
```diff
@@ -27,13 +36,44 @@ abstract class TimeEntryContentViewHolder<in T : TimeEntryContentViewModel>(item
 
         val formattedProjectText = item.project.formatForDisplay()
         project.text = formattedProjectText
-        project.isVisible = formattedProjectText.isNotBlank()
+        if (formattedProjectText.isNotBlank()) {
```
pdrj, line 39: For projects, here and elsewhere I'd consider checking for null only. I can imagine someone wanting to only have a colored dot displayed, without any text and identify projects just by color 🤔

timer/src/main/java/com/toggl/timer/log/ui/TimeEntryItemViewHolder.kt
```diff
@@ -1,23 +1,113 @@
...
+
+        cardView.isClickable = true
+        cardView.isFocusable = true
+        holder.isClickable = true
+        holder.isFocusable = true
+    }
+
+    private fun configureEntryWithoutAGroup(item: FlatTimeEntryViewModel) {
+        configureVisibleEntry(item)
+        durationGradient.background = ContextCompat.getDrawable(
```
pdrj, line 55: Let's reuse all the drawables created on each bind (have just one instance of each).

timer/src/main/java/com/toggl/timer/suggestions/ui/SuggestionsAdapter.kt
```diff
@@ -53,7 +58,16 @@ class SuggestionsViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView)
 
         val formattedProjectText = item.project.formatForDisplay()
         project.text = formattedProjectText
-        project.isVisible = formattedProjectText.isNotBlank()
+        if (formattedProjectText.isNotBlank()) {
```
pdrj, line 61: Same comment as somewhere above about project name blank vs null.

timer/src/main/res/layout/fragment_running_time_entry.xml
```diff
@@ -4,53 +4,133 @@
...
+
+    <ImageView
+        android:id="@+id/arrow"
+        app:tint="@color/primary"
+        android:src="@drawable/arrow_right"
+        android:layout_marginStart="@dimen/grid_2"
+        app:layout_constraintStart_toStartOf="parent"
+        app:layout_constraintTop_toTopOf="parent"
+        app:layout_constraintBottom_toBottomOf="parent"
+        android:layout_width="24dp"
```
pdrj, line 17: Please reformat so attributes are in correct order.

timer/src/main/java/com/toggl/timer/running/ui/RunningTimeEntryFragment.kt
```diff
@@ -39,55 +50,81 @@ class RunningTimeEntryFragment : Fragment(R.layout.fragment_running_time_entry)
...
-                    requireContext(),
-                    R.color.start_time_entry_button_background
-                )
-                backgroundTintList = ColorStateList.valueOf(color)
-                setImageResource(R.drawable.ic_play_big)
-                setOnClickListener {
-                    store.dispatch(RunningTimeEntryAction.StartButtonTapped)
-                }
+    private fun updateDurationTickJob(timeEntryData: RunningTimeEntryData?) {
+        timeIndicatorTimer?.cancel()
```
pdrj, line 111: This should be also cancelled when the view gets destroyed.

pdrj (dismissed):
Lots of good stuff in here. Really digging the expand/collapse animations and haptic feedback 👍 

I've mostly cosmetic comments.

timer/src/main/java/com/toggl/timer/log/ui/TimeEntriesLogAdapter.kt
```diff
@@ -53,12 +53,17 @@ class TimeEntriesLogAdapter(
...
+            timeEntryInClosedGroupViewType, flatTimeEntryViewType -> TimeEntryItemViewHolder(itemView, onTappedListener, onContinueTappedListener)
             timeEntryGroupViewType -> TimeEntryGroupViewHolder(itemView, onGroupTappedListener, onContinueTappedListener, onExpandTappedListener)
             else -> throw IllegalStateException()
         }
 
     companion object {
+        const val dayHeaderViewType = 101
+        const val flatTimeEntryViewType = 103
+        const val timeEntryGroupViewType = 107
+        const val timeEntryInClosedGroupViewType = 109
```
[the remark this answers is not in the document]
heytherewill: The problem here is that we have two different view types that map to the same holder. We need to reuse the holder in order to have proper animations, but we need two different view types in order to identify which views are visible and which are not in the ViewDecorator. That's why I went with prime numbers instead

timer/src/main/res/layout/fragment_running_time_entry.xml
heytherewill: I'd rather focus our efforts on properly extracting Composable functions rather than extracting XML

timer/src/main/java/com/toggl/timer/log/ui/TimeEntryContentViewHolder.kt
heytherewill: Nice thinking!

timer/src/main/res/layout/time_entries_log_suggestion.xml
heytherewill: Turns out our editor config and our `spotlessApply` have many differences 🤔 We should consider improving spotless in another PR

timer/src/main/java/com/toggl/timer/log/ui/TimeEntryItemViewHolder.kt
heytherewill: How would you go about doing that while preventing them from leaking?

pdrj: Would having it stored as a field cause a leak? I don't think it holds the context reference, it just uses it to access resources. 🤔

timer/src/main/res/layout/time_entries_log_suggestion.xml
```diff
@@ -1,82 +1,103 @@
...
+            android:layout_height="0dp"
+            android:background="@drawable/gradient_transparent_to_surface"
+            app:layout_constraintBottom_toBottomOf="parent"
+            app:layout_constraintEnd_toStartOf="@id/continue_icon"
+            app:layout_constraintTop_toTopOf="parent" />
+
+        <ImageView
+            android:id="@+id/continue_icon"
+            android:layout_width="24dp"
+            android:layout_height="24dp"
```
pdrj, line 95: We can use the arrow size constant here too.

timer/src/main/java/com/toggl/timer/running/ui/RunningTimeEntryFragment.kt
pdrj: Why have you decided to keep `lifecycleScope`? Do you think it's safe in this case? 🤔

timer/src/main/java/com/toggl/timer/log/ui/OverlappingCardDecorator.kt
```diff
@@ -0,0 +1,43 @@
...
+
+    override fun getItemOffsets(
+        outRect: Rect,
+        view: View,
+        parent: RecyclerView,
+        state: RecyclerView.State
+    ) {
+        val holder = parent.getChildViewHolder(view)
+        if (holder !is TimeEntryItemViewHolder) {
+            outRect.set(0, 0, 0, 0)
```
pdrj, line 22: In the name of dry, it should be safe to just do it once at the top instead before each return.

pdrj (changes requested):
Just some minor comments and questions left.

Two things that need fixing that I've noticed:

- a bit of an edge case, but it'd be nice to have a bit more spacing Screenshot 2020-09-28 at 13 03 22
- swipe actions don't work over groups

timer/src/main/java/com/toggl/timer/running/ui/RunningTimeEntryFragment.kt
pdrj: Unresolving so you see this :)

pdrj (changes requested):
One more thing I just noticed is that we use black for the TE titles and stuff, while the designs use our purple. It should be OK to change this app-wide (it is, modify the text_on_{primary,surface}).

daividssilverio:
It looks like the running time entry card has the wrong shape, it has squared borders inside a rounded shape.
image

daividssilverio (changes requested):
Looking good! I left a few comments.
Also, the time entry log items in this PR are using this purple, and the designs are using only the `#161119`.
image

I like the purple, but I think we need to discuss this with the designers. Also, neither is working well in dark mode. Really hard to see the grouped time entries.
image

daividssilverio:
We are still missing the current page indicator in the `BottomNavigationView`:
image

I don't think the bottom navigation view supports that out of the box, so it's fine to create an issue for that if you'd prefer. I think we could use a pager or a pager strip or a custom view for that.

Also, the icon colours don't match the designs in both light/dark modes.

timer/src/main/java/com/toggl/timer/log/ui/OverlappingCardDecorator.kt
```diff
@@ -0,0 +1,43 @@
+package com.toggl.timer.log.ui
+
+import android.graphics.Rect
+import android.view.View
+import androidx.recyclerview.widget.ConcatAdapter
+import androidx.recyclerview.widget.RecyclerView
+import androidx.recyclerview.widget.RecyclerView.ItemDecoration
+
+class OverlappingCardDecorator(cardHeight: Int) : ItemDecoration() {
+    private val percentageOfVisibleCardAfterOffsetting = 100 / 5F
```
daividssilverio, line 10: This name is misleading. I've tried changing it the higher the value and this doesn't seem to work like a %.
For example, setting this value to 2 looks like this:
image
Setting it to 25 looks like this:
image

Another point, the visible offset in the designs are even smaller than the example above.
It looks like this:
image

daividssilverio:
Are you implementing the grouped time entries designs in this PR? They don't have the item separation in the designs.
https://toggl.invisionapp.com/console/%5BAndroid%5D-Rebranding-ckem8stapzb0o01wrr0c27nd1/ckeriu1sh02av01xk0y9b4m7z/inspect

heytherewill:
@pdrj We're actually no using black; We are using the correct shade of dark purple for text, as per designs 🤔

pdrj:
> @pdrj We're actually no using black; We are using the correct shade of dark purple for text, as per designs 🤔

My bad then, hard to tell them apart 😄

timer/src/main/java/com/toggl/timer/running/ui/RunningTimeEntryFragment.kt
```diff
@@ -34,60 +46,105 @@ class RunningTimeEntryFragment : Fragment(R.layout.fragment_running_time_entry)
...
+                duration?.post { duration?.setTextIfDifferent(formattedDuration) }
             }
         }
     }
+
+    private fun ProjectData.formatForDisplaying() =
+        buildSpannedString {
+            append(name)
+            append(" ")
+            append(clientName)
```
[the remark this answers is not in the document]
heytherewill: Let's create a followup issue, then

timer/src/main/java/com/toggl/timer/log/ui/SwipeActionCallback.kt
```diff
@@ -127,33 +129,35 @@ class SwipeActionCallback(
...
+            swipeBackground.color = backgroundColor.toColorStateList()
+            swipeBackground.setBounds(
+                backgroundBounds.left,
+                backgroundBounds.top,
+                backgroundBounds.right,
+                backgroundBounds.bottom
+            )
+            swipeBackground.draw(this)
+
+            clipRect(backgroundBounds)
```
daividssilverio, line 148: Is this doing anything?

timer/src/main/java/com/toggl/timer/log/ui/OverlappingCardDecorator.kt
```diff
@@ -0,0 +1,36 @@
...
+
+import android.graphics.Rect
+import android.view.View
+import androidx.recyclerview.widget.ConcatAdapter
+import androidx.recyclerview.widget.RecyclerView
+import androidx.recyclerview.widget.RecyclerView.ItemDecoration
+
+class OverlappingCardDecorator(cardHeight: Int) : ItemDecoration() {
+    private val percentageOfVisibleCardAfterOffsetting = 1F
+    private val visibleCardHeightAfterOffsetting = cardHeight / (100 / percentageOfVisibleCardAfterOffsetting)
```
daividssilverio, line 11: `percentageOfVisibleCardAfterOffsetting` now is not doing anything, right?

daividssilverio:
Not sure if this is a problem or a feature, but since we are not moving the grouped time entries when the group header is swiped, we have this effect.
image

I've also noticed that the swipe action's text paint could use a anti alias flag.

timer/src/main/java/com/toggl/timer/log/ui/TimeEntryItemViewHolder.kt
```diff
@@ -85,6 +85,7 @@ class TimeEntryItemViewHolder(
         unsyncable.isVisible = false
         continueIcon.isVisible = false
         duration.isVisible = false
+        itemView.isClickable = false
```
daividssilverio, line 88: actually, this has to happen after the `itemView.setOnClickListener(null)` call, it internally sets `isClickable` to true, whether or not the listener is null.

timer/src/main/res/layout/fragment_time_entries_log.xml
```diff
@@ -24,26 +24,27 @@
...
     <com.google.android.material.card.MaterialCardView
         android:id="@+id/running_time_entry_layout"
         android:layout_width="match_parent"
-        android:layout_height="80dp"
+        android:layout_height="wrap_content"
         android:layout_gravity="bottom"
         app:cardBackgroundColor="@color/toolbar"
-        app:shapeAppearanceOverlay="@style/ShapeAppearanceOverlay.Toggl.MediumComponent.Bound.Bottom"
-        app:elevation="@dimen/plane_02">
+        app:elevation="@dimen/plane_02"
```
daividssilverio, line 46: Hey, using `cardElevation` provides a nicer effect here. Also, We might consider putting this on a higher plane.

timer/src/main/res/layout/fragment_running_time_entry.xml
```diff
@@ -2,55 +2,187 @@
...
+        app:layout_constraintEnd_toStartOf="@id/duration"
+        app:layout_constraintStart_toEndOf="@id/project_color_indicator"
+        app:layout_constraintTop_toBottomOf="@id/running_time_entry_description"
+        tools:text="Time Management" />
+
+    <View
+        android:id="@+id/gradient"
+        android:layout_width="0dp"
+        android:layout_height="0dp"
+        android:background="@drawable/gradient_transparent_to_running_entry_card"
```
daividssilverio, line 115: The gradient isn't right here: 
image

daividssilverio (commented):
Looking good! Please create issues for the layout details and colours not handled here.
- The bg colour of the time entries
- The tab indicator in the bottom nav
- The layout of the grouped time entries
- ?? The grouped time entries swiping (not sure if we are doing, but it would be a nice detail to swipe the whole group) ¿¿

timer/src/main/java/com/toggl/timer/log/ui/SwipeActionCallback.kt
heytherewill: Nope

timer/src/main/java/com/toggl/timer/log/ui/OverlappingCardDecorator.kt
heytherewill: Not when the percentage is 1, but it makes it easier to understand and change the amount of visible card in the future so I think it's worth keeping it

timer/src/main/res/layout/fragment_time_entries_log.xml
heytherewill: The problem of card elevation is the fact that our designs didn't consider the automatic elevation added by the material design library to dark theme surfaces

heytherewill:
> The bg colour of the time entries

There's nothing wrong here, I'm using the exact same color as the one in designs, what do you mean? 🤔 

> The tab indicator in the bottom nav

I will give one final try here and create an issue otherwise 👌 

> The layout of the grouped time entries

I'm also not sure what you mean here D:

> The grouped time entries swiping (not sure if we are doing, but it would be a nice detail to swipe the whole group) ¿¿

We won't change them to swipe altogether until we are doing compose where this is much easier

pdrj (approved):
Looking good!

Minor points that I noticed, but not blocking:
- the drawable reusing I suggested in a previous conversation
- the swipe overlay over groups looks like it's too small, but we can live with that ;)

daividssilverio:
>> The bg colour of the time entries

>There's nothing wrong here, I'm using the exact same color as the one in designs, what do you mean? 🤔 

The grouped time entries have a different background colour on dark theme, look:
image

>> The tab indicator in the bottom nav

>~I will give one final try here and create an issue otherwise 👌~ Adressed :) 

👍 

>> The layout of the grouped time entries

>I'm also not sure what you mean here D:

They were supposed to look like this:
image

The time entries in the middle of the group don't have rounded borders and the spacing is minimal.
I'm fine the way it is right now tbh.

This is what we have right now:
image

>> The grouped time entries swiping (not sure if we are doing, but it would be a nice detail to swipe the whole group) ¿¿

>We won't change them to swipe altogether until we are doing compose where this is much easier

👍

daividssilverio (approved):
Looking good to me.
The app crashes when switching tabs sometimes btw. 
`swipe_refresh_layout` is null. Using `viewLifecycleOwner.lifecycleScope` on `TimeEntryLogFragment:L146` would fix it.

Please create issues regarding confusion on designs (or we can discuss this internally only with our designers, up to you).

PR #943, opened 2020-09-28
Hide implementation details of the TE duration deserialization

### LINKED ISSUE #814: Revisit the place where we check for the running time entry logic
## Description
As described in https://github.com/toggl/android/pull/805#discussion_r482053436, we shouldn't leak an implementation detail of v9 into the sync code. We need to figure out where this check should be placed instead.

## Definition of done
- [x] We decide where this check should be moved
- [x] We move the running time entry check elsewhere

simonrozsival:
### Summary
This PR moves code which interprets the `"duration"` property of the pulled TEs in the `api` module using a Moshi adapter instead of keeping the code in the `sync` module.

### Relationships
Closes #814 

### Pull Request checklist

### Pointers
Check the tests.

### Review template

## Merge permissions
🦑


api/src/main/java/com/toggl/api/network/adapters/NullableDurationAdapter.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import com.squareup.moshi.FromJson
+import com.squareup.moshi.ToJson
+import java.time.Duration
+
+class NullableDurationAdapter {
+    @ToJson
+    @Suppress("UNUSED_PARAMETER")
+    fun toJson(duration: Duration?): String =
+        throw NotImplementedError("serializing Duration? is not implemented")
```
heytherewill, line 11: Why not? 🤔

heytherewill (approved):
Great improvement :)

api/src/main/java/com/toggl/api/network/adapters/NullableDurationAdapter.kt
simonrozsival: Because if the duration is `null`, then the value should be `-{unix timestamp of start time}`, which is information we don't have. We would have to create a different type which combines the duration and start time and which could switch between the two values... 🤷 I guess we don't need that much refactoring at the moment.

heytherewill: Oh, fair enough, I thought returning `-1` would suffice, but yeah, we'd better just go kaboom instead

simonrozsival: Unfortunately `-1` wouldn't be enough, because the timestamp has to be within 40 days of the start time of the TE.

PR #945, opened 2020-09-28
Pull response processing: Throw if referenced entities are missing

### LINKED ISSUE #797: Figure out a way of handling entities not found in the database during Pull Sync
## Description
As Simon pointed out [here](https://github.com/toggl/android/pull/791#discussion_r477269249), when we can't find entities during a pull sync, it might mean some serious problem, so we should figure out how to handle this

## Dependencies
Let's do this after #750 so we can do it once for all entities

## Definition of done
- [ ] All the queries that have a `continue` in the pull processing methods do some kind of logging/throw so we can be aware of critical sync bugs.

simonrozsival:
### Summary
This PR throws instead of silently skipping the cases where one entity references an unknown entity (a workspace or in the case of tasks also a project). We should make sure this never occurs. If we see these exceptions in the logs, we know something's wrong and we need to fix it (either our app or the sync server).

### Relationships
Closes #797 

### Pull Request checklist

### Review template

## Merge permissions
🦑


heytherewill (approved):
Is the bug of trying to process a null user/preference fixed in this PR as well?

simonrozsival:
Nope, this was related to pull response and the issue was related to push responses.

PR #951, opened 2020-09-29
Fix push issues: batch 1

### LINKED ISSUE #949: Created TEs are missing the `created_with` property in the JSON body
What the title says. The API rejects an `create` actions for TEs which don't contain `"created_with": "<some string>"`. The string should be the same as the user agent header.

### LINKED ISSUE #950: Fix processing user and preferences update action result
We're always processing the `user` and `preferences` action result even though it's `null` which throws an exception.

simonrozsival:
### Summary
This PR fixes two bugs I found this morning. I know about more issues with updating TEs which I haven't finished debugging yet and which will be addressed later.

### Relationships
Closes #949, closes #950 

### Pull Request checklist

### Pointers
Try it yourself in the emulator!

### Review template

## Merge permissions
🦑


heytherewill (approved):

heytherewill (changes requested):
Actually, please add a test to prevent us from making the null user mistake again

sync/src/test/java/com/toggl/sync/processing/ResponseProcessorTest.kt
```diff
@@ -0,0 +1,31 @@
...
+            clients = emptyList(),
+            tasks = emptyList(),
+            tags = emptyList(),
+            user = null,
+            preferences = null,
+        )
+
+        processor.process(response)
+
+        // if we reach the end of the test, no exception was thrown
```
daividssilverio, line 26: There's this nice assertion that you could use here instead of leaving a comment
```
io.kotest.assertions.throwables.shouldNotThrowAny
```

sync/src/test/java/com/toggl/sync/processing/ResponseProcessorTest.kt
```diff
@@ -0,0 +1,31 @@
+package com.toggl.sync.processing
+
+import com.toggl.api.network.models.sync.PushResponse
+import com.toggl.sync.BaseDatabaseTest
+import io.mockk.mockk
```
daividssilverio, line 5:
```suggestion
import com.toggl.sync.BaseDatabaseTest
import io.kotest.assertions.throwables.shouldNotThrowAny
import io.mockk.mockk
```

sync/src/test/java/com/toggl/sync/processing/ResponseProcessorTest.kt
daividssilverio, line 26:
```suggestion
        shouldNotThrowAny {
            processor.process(response)
        }
```

heytherewill (approved):

daividssilverio (approved):

PR #953, opened 2020-09-29
💾 DataStore migration

### LINKED ISSUE #897: Replace SharedPreferences with DataStore
## Description

[DataStore](https://developer.android.com/topic/libraries/architecture/datastore) is a direct successor to SharedPreferences. The transition should be painless so let's do it now before we enter the production stage. 
 
## Dependencies
None

## Definition of done
- [ ] All SharedPreferences are removed and replaced with DataStore

heytherewill:
### Summary
This migrates our stuff to DataStore. This api is surprisingly nice

### Technical considerations
You see these `runBlocking`s and get disgusted? Well, it was literally as bad in SharedPrefs, only we weren't disgusted 😂 
It's a good exercise to embrace this async api more often in the future

### Relationships
Closes #897

### Pull Request checklist

### Pointers
Review commit by commit and do a functional test for the bits that are affected by the change

### Review template

## Merge permissions
Anyone, anytime


semanticer (approved):
I have nothing to add, it looks pretty good to me. 

The app is currently crashing left and right for me but none of these issues seem to be related to your changes.

PR #954, opened 2020-09-29
Access restriction reducer

### LINKED ISSUE #901: Add AccessRestrictionReducer
## Description
`AccessRestrictionReducer` should be implemented as a higher-order reducer. For now, it shouldn't do anything, just pass along all the actions it receives.

 ~`AccessRestrictionState` should be added too~

[Notion docs](https://www.notion.so/toggl/Access-Restriction-handling-92338df62076457da8120dd79e989eb8)

## Dependencies
None

## Definition of done
 ~`AccessRestrictionState` is added to global state~
- [ ] `AccessRestrictionReducer` is added as one of the higer order reducers

### LINKED ISSUE #903: Add AccessRestrictionUpdated action
## Description
`AccessRestrictionReducer` should react to `AccessRestrictionUpdated` by changing the `AccessRestrictionState` and navigating to the correct error handling screen. 

[Notion docs](https://www.notion.so/toggl/Access-Restriction-handling-92338df62076457da8120dd79e989eb8)

## Dependencies
#901 

## Definition of done
- [ ] `AccessRestrictionUpdated` action is handled by `AccessRestrictionReducer`
- [ ] Tests are added

semanticer:
### Summary
This is implementing Access restriction higher order reducer

### Technical considerations
🤷‍♂️

Let me know if there is a better place for `AccessRestrictions`

### Relationships
Closes #901 
Closes #903

### Pull Request checklist

### Pointers
It should be easy to review everything at once. The app should work normally as before. 

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


app/src/main/java/com/toggl/domain/reducers/AccessRestrictionReducer.kt
```diff
@@ -0,0 +1,44 @@
...
+    }
+}
+
+data class AccessRestrictions(
+    val clientOutdated: Boolean = false,
+    val apiOutdated: Boolean = false,
+    val unauthorizedAccess: Boolean = false,
+    val noWorkspaceStateReached: Boolean = false,
+    val noDefaultWorkspaceStateReached: Boolean = false
+)
```
heytherewill, line 44: Since a user can never have more than one of those, this class feels like it's leaking the implementation detail of the way we are persisting the errors. I'd turn it into a `sealed class` with a `NoRestrictions` object so that we never forget to handle any of the possible values

heytherewill (changes requested):
Looking good, but I think we can improve it a bit further

app/src/main/java/com/toggl/domain/reducers/AccessRestrictionReducer.kt
semanticer: You are 100% right

heytherewill (approved):

PR #957, opened 2020-09-29
🔨⏰ Display start/end times considering the local timezone

### LINKED ISSUE #876: Fix timezone issues in the calendar
There are some issues with the current implementation. Try some different timezones when creating time entries from the calendar, see what happens.

Basically it should work as wysiwyg. A time entry created now shouldn't appear in the calendar with a different timezone.

daividssilverio:
### Summary
This makes the currently displayed start and end times take into consideration the local timezone offset.

### Technical considerations
I actually don't know why simply using `toLocalDateTime` didn't work.

🤷‍♂️ If you are savvy with Java's time API, please enlighten us on what's the right way to go here.

### Relationships
Closes #876
Actually, the times in the calendar events were correct, but the editing hour indicators were wrong, and so were the start and end times displayed in the StartEdit view.

### Pull Request checklist

### Pointers
Please test functionally. Try creating time entries on different clients to check if they all match what you would expect to see.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 it


heytherewill (changes requested):
Should we not be using the timezone from the user's preferences and ask to update it if it doesn't match the phone's timezone? Or should we continue using the phone's timezone?

daividssilverio:
> Should we not be using the timezone from the user's preferences and ask to update it if it doesn't match the phone's timezone? Or should we continue using the phone's timezone?

I would just keep using the phone's timezone for now. The old implementation doesn't consider it as well. I don't remember a single complaint about that on reviews or support. Ideally we should do it at some point though.

heytherewill (approved):

PR #959, opened 2020-09-30
👨‍🎨 Finalize settings designs

### LINKED ISSUE #891: Final designs: settings
## Description
Designs here: https://toggl.invisionapp.com/overview/-Android--Rebranding-ckem8stapzb0o01wrr0c27nd1/screens/21535235

## Definition of done
- [ ] settings look as in invision, it should be mostly just coloring work (note that we're keeping our dialog approach to single-choice stuff, invision is not up to date on that part)

pdrj:
### Summary
Applies the final designs to the main settings screen

### Technical considerations
- The colors, especially in dark theme are not there yet. As I understand, @heytherewill is doing some color work app-wide and I didn't modify the values in the shared package
- Same as above, there are some hardcoded colors which I believe will go away when Will is done
- This covers the main screen only, I'll create a separate issue for the sub-screens, as otherwise this PR might get too large
- I'm using some hardcoded font sizes, I'm not sure if we want to have them extracted somewhere, or we want to overwrite the theme 🤔 (the designs use mostly a 15sp font, while the default material theme uses 14sp for body)
- The sync indicator is always in the 'syncing' state. This might be a problem with this code, or something external. We need a separate issue anyway, to also handle the 'Signing you out securely' state

### Relationships

Closes #891 

### Pull Request checklist

### Pointers
Best done in light mode for now. Validate the settings screen looks as designed.
Code-wise commit-by-commit should be ok.

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


heytherewill (approved):
This looks really good! You are correct to assume that the colors will come alongside with my PR; so this one just needs to care about the visuals!

The only weird thing I noticed is that the `Toggl Blog on Personal Efficiency` and `Help` sections have a lot of empty space on the bottom even thought they don't have a subtitle. We should talk to designers to remove those 🤔 

Also, the sync indicator is broken because of a series of bugs in sync; it should normalize Soon:tm:

Issue #904, opened 2020-09-22
Implement AccessRestrictionSubscription


semanticer:
## Description
`AccessRestrictionSubscription` should observe `AccessRestrictionRepository`, merge all flows into one data class which is then wrapped in `AccessRestrictionUpdated` action and emited immediately. 

[Notion docs](https://www.notion.so/toggl/Access-Restriction-handling-92338df62076457da8120dd79e989eb8)

## Dependencies
#898 #903

## Definition of done
- [ ] AccessRestrictionSubscription is added
- [ ] AccessRestrictionSubscription is tested

PR #962, opened 2020-09-30
Implement access restriction repository and subscription 

### LINKED ISSUE #898: Define & Implement AccessRestrictionRepository
## Description
We need `AccessRestrictionRepository` to persist all error states that need to be resolved before we let the user interact with the app again.

## Dependencies
Consider doing this after migration to DataSotre #897

## Definition of done
- [ ] `AccessRestrictionRepository` interface is defined ([Xamarin version](https://github.com/toggl/mobileapp/blob/3faa63ac478fcc66898183ed0e36e10a353104ab/Toggl.Storage/Settings/IAccessRestrictionStorage.cs)) where all those is*/has* getter methods should be returning Flow in Aurora
- [ ] Repository implements AccessRestrictionRepository

semanticer:
### Summary
This implements AccessRestrictionRepository coresponding subscription

### Technical considerations
Nothing super interesting here, probably the hardest part is to test the subcription properly, **I will improve this part tomorrow.** 

Let me know if you have any ideas

### Relationships

Closes #898 #904 

### Pull Request checklist

### Pointers
Should be faily straight forwar to review this

### Review template

- [ ] The code considers i18n

## Merge permissions
Anyone


app/src/main/java/com/toggl/domain/loading/LoadAccessRestrictionsSubscription.kt
```diff
@@ -0,0 +1,46 @@
...
+class LoadAccessRestrictionsSubscription(
+    private val repository: Repository,
+    private val dispatcherProvider: DispatcherProvider
+) : Subscription<AppState, AppAction> {
+
+    override fun subscribe(state: Flow<AppState>): Flow<AppAction> =
+        repository.apiTokenFlow()
+            .distinctUntilChanged()
+            .flatMapLatest { apiToken ->
+                withContext(dispatcherProvider.io) {
```
heytherewill, line 25: Wouldn't using `flowOn` make this far more readable?

repository/src/main/java/com/toggl/repository/Repository.kt
```diff
@@ -196,6 +204,9 @@ class Repository @Inject constructor(
     override suspend fun getApiToken(): ApiToken =
         dataStore.get(apiTokenKey, "").let(ApiToken.Companion::from)
 
+    override fun apiTokenFlow(): Flow<ApiToken> =
+        dataStore.data.map { pref -> ApiToken.from(pref[apiTokenKey] ?: "") }
```
heytherewill, line 208:
```suggestion
        dataStore.data.map { pref -> ApiToken.from(pref[apiTokenKey] ?: "") }.distinctUntilChanged()
```

repository/src/main/java/com/toggl/repository/Repository.kt
```diff
@@ -227,4 +238,39 @@ class Repository @Inject constructor(
...
+
+    override suspend fun setNoDefaultWorkspaceStateReached(hasNoDefaultWorkspace: Boolean) {
+        dataStore.set(noDefaultWorkspaceStateReachedKey, hasNoDefaultWorkspace)
+    }
+
+    override fun isClientOutdated(): Flow<Boolean> =
+        dataStore.data.map { preferences -> preferences[clientOutdatedKey] ?: false }
+
+    override fun isApiOutdated(): Flow<Boolean> =
+        dataStore.data.map { preferences -> preferences[apiOutdatedKey] ?: false }
```
heytherewill, line 266: For these two specifically, we are persisting the app version that got a teapot, this way when the user updates the app version will change and we can then make sure that they are no longer outdated. If you persist only a boolean, how are you going to be able to recover an outdated client?

heytherewill (changes requested):

PR #963, opened 2020-09-30
🔨 Confirm destructive actions in the StartEdit flow

### LINKED ISSUE #868: Add the destructive action's confirmation dialog in the StartEdit time entry view.
Check the old implementation for more details. But basically the only case where we don't show the confirmation dialog is for when the user tries to start a time entry, doesn't touch anything else and presses the cancel or back button.

daividssilverio:
### Summary
This PR adds the destructive actions confirmation handling in the StartEdit flow.

### Technical considerations
I've tried using the idea discussed [here](https://toggl.slack.com/archives/C02APMYTM/p1601450812156100), but it wasn't looking worth it. At least with the way I tried, there would be a lot of extra changes for little added benefit. I'll push the wip for comparison.

Because it was too much trouble the other way.
By all means, if you know a better way to handle it, tell me.

I've tried adding a `Changeable` class and replacing the `editableTimeEntry` in the StartEdit state with a `Changeable<EditableTimeEntry>`.

```kotlin
class Changeable<T> internal constructor(
    val currentValue: T,
    private val originalValue: T
) {
    constructor(value: T) : this(value, value)

    val hasChanged: Boolean = currentValue != originalValue

    fun update(newValue: T.() -> T): Changeable<T> {
        return Changeable(newValue(currentValue), originalValue)
    }

    operator fun getValue(thisRef: T, property: KProperty<*>): Changeable<T> = this
    operator fun invoke(): T = this.currentValue
}
```

And then all references to the `editableTimeEntry` in the `StartEditReducer` had to be fixed and the `copy` usage replaced with `update`. All of that just to check if changes were made in the `editableTimeEntry`... Is it really worth it?

### Relationships
Closes #868

### Pull Request checklist

### Pointers
Give it a try.

### Review template

- [ ] The code considers i18n

## Merge permissions
🦑 after the tests are done


common/src/main/java/com/toggl/common/Confirmation.kt
```diff
@@ -0,0 +1,6 @@
+package com.toggl.common
+
+sealed class Confirmation {
+    object NotAsking : Confirmation()
+    object Asking : Confirmation()
+}
```
heytherewill, line 6: As much as I like sealed classes, this is just a boolean

timer/src/main/java/com/toggl/timer/startedit/domain/StartEditState.kt
```diff
@@ -46,11 +50,13 @@ data class StartEditState(
                 timeEntries = timerState.timeEntries,
                 preferences = timerState.preferences,
                 backStack = timerState.backStack,
+                originalTimeEntry = originalTimeEntry,
                 editableTimeEntry = editableTimeEntry,
```
heytherewill, line 54: The only problem I see here is that this forces consumers to understand that both the `originalTimeEntry` and the `editableTimeEntry` need to be set. The other syntax with the `Changeable<T>` was more solid in this regard (even though I didn't like it overriding `invoke`.

If you still feel strongly about using this current solution, please convince me of why this is better or how we can mitigate people using it wrong and only setting one of the two properties

heytherewill: Now that I think about it, we can put both original and editable tes in the route. This way the user is forced to supply both when creating the time entry and then during the state mapping we can be clever about it., but I'm not sure how much work this solution is

